PR #6, opened 2017-05-12
Consume libecl as targets

jokva:
This is a showcase of what downstream can look like with https://github.com/Statoil/libecl/pull/54 applied. Currently it's just the absolute minimal set of changes to demo what's going on.

Compiled and tested with `cmake -DCMAKE_MODULE_PATH=libecl/cmake/Modules -DCMAKE_MODULE_PATH=libecl/python/cmake/Modules`


joakim-hove:
Looks good - did you actually intend to pass path to lib*res* - or should that have been living*ecl*?

jokva:
Should've been *ecl* indeed, was a bit quick there. The point was to show that include paths and dependencies-of-dependencies won't need to be specified downstream, they're all handled by cmake. When libecl is installed we can just point it to that installation and we're good.

PR #10, opened 2017-05-18
Extract loading site_config from enkf_main

markusdregi:
Enkf_main is now provided a site_config upon allocation. This is to separate how the configurations are retrieved from their applications. At the same time some refactoring is done to ease this work for the other config-objects.

### DIFF
--- a/libenkf/include/ert/enkf/site_config.h
+++ b/libenkf/include/ert/enkf/site_config.h
@@ -42,6 +42,7 @@ typedef struct site_config_struct site_config_type;
   const char *             site_config_get_location();
 
   const char *             site_config_get_config_file(const site_config_type*);
+  // TODO: remove?
   void                     site_config_set_config_file(site_config_type*, const char*);
   
   const char *             site_config_get_manual_url( const site_config_type * site_config );
@@ -103,6 +104,8 @@ typedef struct site_config_struct site_config_type;
   mode_t                   site_config_get_umask( const site_config_type * site_config );
 
   site_config_type       * site_config_alloc_empty();
+  site_config_type       * site_config_alloc_default();
+
   void                     site_config_add_config_items( config_parser_type * config , bool site_mode);
 #ifdef __cplusplus
 }
--- a/libenkf/include/ert/enkf/site_config.h
+++ b/libenkf/include/ert/enkf/site_config.h
@@ -42,7 +42,6 @@ typedef struct site_config_struct site_config_type;
   const char *             site_config_get_location();
 
   const char *             site_config_get_config_file(const site_config_type*);
-  // TODO: remove?
   void                     site_config_set_config_file(site_config_type*, const char*);
   
   const char *             site_config_get_manual_url( const site_config_type * site_config );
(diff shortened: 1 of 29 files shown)


joakim-hove:
Rebase?

markusdregi:
@jokva Done rebasing. I just fixed some broken Statoil tests on master first..

joakim-hove (approved):

libenkf/include/ert/enkf/site_config.h
```diff
@@ -101,6 +103,11 @@ typedef struct site_config_struct site_config_type;
   mode_t                   site_config_get_umask( const site_config_type * site_config );
 
   site_config_type       * site_config_alloc_empty();
+  site_config_type       * site_config_alloc_default();
+  site_config_type       * site_config_alloc_model_config(const char *);
```
joakim-hove: This allocates a `site_config` instance - i.e. it should be named something like `site_config_alloc( )`; the current name seems to indicate that this function will create a `model_config` structure from a `site_config` - that is not the case.

libenkf/include/ert/enkf/site_config.h
```diff
@@ -101,6 +103,11 @@ typedef struct site_config_struct site_config_type;
   mode_t                   site_config_get_umask( const site_config_type * site_config );
 
   site_config_type       * site_config_alloc_empty();
+  site_config_type       * site_config_alloc_default();
```
joakim-hove: Ideally I guess it should be possible to create a valid and usable `site_config` instance without any file-parsing; but I do not think that is yet the case? As long as a site configuration *file* is actually needed I think it is a better idea to let the `site_config_alloc`function take the configuration file as input, and remove `site_config_set_config_file( )` function.

libenkf/include/ert/enkf/site_config.h
```diff
@@ -101,6 +103,11 @@ typedef struct site_config_struct site_config_type;
...
+  site_config_type       * site_config_alloc_model_config(const char *);
+
+  config_content_type    * site_config_alloc_content(const site_config_type*, config_parser_type*);
```
joakim-hove: Two questions about this function:

1. Is it necessary to pass a `config_parser` instance? Possibly we could just create, use and discard the parser inside the function.

2. Is the function really required - at least public? I would guess that this could be integrated in the `site_config_alloc( const char * site_config_file)` constructor?

libenkf/src/model_config.c
```diff
@@ -634,3 +642,215 @@ void model_config_fprintf_config( const model_config_type * model_config , int e
   fprintf(stream , "\n\n");
 
 }
+
```
joakim-hove, line 645: Hmmm - there was already `model_config` structure, seems that has changed implementation and principles quite dramatically?

libenkf/include/ert/enkf/site_config.h
markusdregi: I agree! I already changed it in the PR extending this one..

libenkf/include/ert/enkf/site_config.h
markusdregi: I agree. I meant to remove both of these method.

libenkf/include/ert/enkf/site_config.h
markusdregi: 1. Unfortunately, yes (for now..). Currently, the config_content needs the config_parser provided at construction to be alive until the very end of parsing. Should this be changed so that the config_content owns it config_parser, YES! But this requires some effort (in a different part of the code) and should in my opinion it be done in a separate PR.

2. Yes, everyone parsing the site config file needs exactly this config_content set up. And then it should be done one place and I think that it is natural that it is site_config that knows how a site config file should be parsed.

libenkf/src/model_config.c
markusdregi: In a similar manner that site_config knows how the site configuration file should be loaded, the model_config now knows how the model configuration file should be loaded. This is basically just code moved from enkf_main.c to model_config.c together with the new method model_config_alloc_content.

joakim-hove:
A fix in Statoil/ert is required.

PR #17, opened 2017-05-29
Logging in job_queue

epa095:
**Task**
Improve (or add) logging to `job_queue`

**Approach**
- Adding a simpler log-method ```  res_log_add_message_str``` which only takes log-level and a string. 
- Replaced several uses of ``` res_log_add_message``` with the simpler ```  res_log_add_message_str```
- Added some log-lines to job_queue.c
- Added some log-lines to lsf_driver.c
- Replaced use of ```res_log_add_message``` with string-formating and free=true with ```res_log_add_fmt_message.``` which does the formating, allocation and freeing itself. 
- Replacing some fprints to sterr with logcalls in lsf_driver

**Pre un-WIP checklist**

### DIFF
--- a/libenkf/src/enkf_main.c
+++ b/libenkf/src/enkf_main.c
@@ -1538,7 +1538,7 @@ static void enkf_main_start_queue(enkf_main_type * enkf_main,
   enkf_main_submit_jobs( enkf_main , run_context, job_queue);
 
   job_queue_submit_complete( job_queue );
-  res_log_add_message( 1 , NULL , "All jobs submitted to internal queue - waiting for completion" ,  false);
+  res_log_add_message_str(LOG_INFO , "All jobs submitted to internal queue - waiting for completion");
 
   int max_runtime = analysis_config_get_max_runtime(enkf_main_get_analysis_config( enkf_main ));
   job_queue_set_max_job_duration(job_queue, max_runtime);
--- a/libenkf/src/ert_workflow_list.c
+++ b/libenkf/src/ert_workflow_list.c
@@ -257,7 +257,7 @@ void ert_workflow_list_add_jobs_in_directory( ert_workflow_list_type * workflow_
 	    if (full_path) {
 	      set_add_key( names , root_name );
 	      if (res_log_is_open())
-		res_log_add_message( 1 , NULL , util_alloc_sprintf("Adding workflow job:%s " , full_path ), true);
+            res_log_add_fmt_message(LOG_INFO , NULL , "Adding workflow job:%s " , full_path);
 
 	      ert_workflow_list_add_job( workflow_list , root_name , full_path );
 	    }
--- a/libjob_queue/src/lsf_driver.c
+++ b/libjob_queue/src/lsf_driver.c
@@ -26,10 +26,12 @@
 #include <ert/util/util.h>
 #include <ert/util/hash.h>
 #include <ert/util/stringlist.h>
+#include <ert/res_util/res_log.h>
 
 #include <ert/job_queue/queue_driver.h>
 #include <ert/job_queue/lsf_driver.h>
 #include <ert/job_queue/lsf_job_stat.h>
+#include <ert/util/log.h>
 
 #ifdef HAVE_LSF_LIBRARY
 #include <ert/job_queue/lsb.h>
@@ -310,6 +312,7 @@ static void lsf_driver_internal_error( const lsf_driver_type * driver ) {
   fprintf(stderr , "** a value for LSF_SERVER. Set this and try again.             **\n");
 #endif
   fprintf(stderr , "*****************************************************************\n\n");
+  res_log_add_message_str(LOG_ERROR,"In lsf_driver, attempt at submitting without setting a value for LSF_SERVER.");
   exit(1);
 }
 
@@ -513,6 +516,7 @@ static int lsf_driver_submit_shell_job(lsf_driver_type * driver ,
 
       if (driver->debug_output)
         printf("Submitting: %s %s %s \n",driver->rsh_cmd , argv[0] , argv[1]);
+      res_log_add_fmt_message(LOG_DEBUG,NULL, "Submitting: %s %s %s \n",driver->rsh_cmd , argv[0] , argv[1]);
 
       util_spawn_blocking(driver->rsh_cmd, 2, (const char **) argv, tmp_file, NULL);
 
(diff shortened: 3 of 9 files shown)


libres_util/include/ert/res_util/res_log.h
```diff
@@ -28,6 +28,16 @@ extern "C" {
...
 #include <ert/util/log.h>
 
+//Same as pythons default log levels.
+typedef enum {
+  CRITICAL=0, //OOM.
+  ERROR=1, //When something we really expected to work does not, e.g. IO failure.
+  WARNING=2, //Important, but not error. E.g. combination of settings which can be intended, but probably are not.
+  INFO=3, //Entering functions/parts of the code
+  DEBUG=4 //Inside the for-loop, when you need the nitty gritty details. Think TRACE.
+} message_level;
```
joakim-hove: Should be `message_level_enum` or `message_level_type` - the identifer `message_level` alone will look like variable.

libres_util/src/res_log.c
```diff
@@ -59,6 +58,15 @@ void res_log_init_log_default( bool verbose){
   res_log_init_log(DEFAULT_LOG_LEVEL, DEFAULT_LOG_FILE,verbose);
 }
 
+/**
+ * Simplified method for adding a message to the log. Does not duplicate the message, and does not
+ * attempt to free the message.
+ * @param message_level Importance of the message.
+ * @param message The message, will not be free'd
+ */
+void res_log_add_message_str(message_level message_level, char* message){
```
joakim-hove: The `message` argument should be `const char *`

libres_util/include/ert/res_util/res_log.h
```diff
@@ -28,6 +28,16 @@ extern "C" {
 
 #include <ert/util/log.h>
 
+//Same as pythons default log levels.
+typedef enum {
+  CRITICAL=0, //OOM.
```
joakim-hove: Maybe these should be in the `libecl/log` implementation; and prefixed with `LOG_`- i.e. `LOG_ERROR` and `LOG_DEBUG`?

joakim-hove:
This is good. The most important new log sites in `job_queue` is in the lsf_driver.c: https://github.com/Statoil/libres/blob/master/libjob_queue/src/lsf_driver.c#L770  and https://github.com/Statoil/libres/blob/master/libjob_queue/src/lsf_driver.c#L942

libres_util/include/ert/res_util/res_log.h
epa095: Yeah, that sounds like a good idea.

libres_util/src/res_log.c
```diff
@@ -16,14 +16,23 @@
    for more details. 
 */
 
+/*TODO:
```
epa095, line 19: Probably not the place for these maybe, but where else? At least it is close to the code, and that is kind of nice.

libjob_queue/src/lsf_driver.c
```diff
@@ -919,12 +923,16 @@ void * lsf_driver_submit_job(void * __driver ,
       lsf_submit_method_enum submit_method = driver->submit_method;
       pthread_mutex_lock( &driver->submit_lock );
 
-      if (driver->debug_output)
-        printf("LSF DRIVER submitting using method:%d \n",submit_method);
+      if (driver->debug_output){
+        printf("LSF DRIVER submitting using method:%d \n",submit_method); //TODO: remove?
+        res_log_add_fmt_message(LOG_INFO, NULL ,"LSF DRIVER submitting using method:%d \n",submit_method);
```
epa095: Could avoid the call above by using stdout instead of NULL here, but I dont know if I want to mix logging and normal user-output in that way.

joakim-hove: The whole system with debug output in the lsf_driver was there because we did not have access to a logging facility - that is now changed. 

I would actually suggest completely removing the `driver->debug_output` facility.

libres_util/src/res_log.c
joakim-hove: That is OK.

pgdr:
Please rebase on master.

libenkf/src/enkf_main.c
```diff
@@ -1538,7 +1538,7 @@ static void enkf_main_start_queue(enkf_main_type * enkf_main,
   enkf_main_submit_jobs( enkf_main , run_context, job_queue);
 
   job_queue_submit_complete( job_queue );
-  res_log_add_message( 1 , NULL , "All jobs submitted to internal queue - waiting for completion" ,  false);
+  res_log_add_message_str(LOG_ERROR , "All jobs submitted to internal queue - waiting for completion");
```
joakim-hove: Does not sound like an `ERROR` does it?

libenkf/src/ert_workflow_list.c
```diff
@@ -257,7 +257,7 @@ void ert_workflow_list_add_jobs_in_directory( ert_workflow_list_type * workflow_
 	    if (full_path) {
 	      set_add_key( names , root_name );
 	      if (res_log_is_open())
-		res_log_add_message( 1 , NULL , util_alloc_sprintf("Adding workflow job:%s " , full_path ), true);
+            res_log_add_fmt_message( 1 , NULL , "Adding workflow job:%s " , full_path);
```
joakim-hove: I guess this `1` should be one of the symbolic names?

libjob_queue/src/lsf_driver.c
joakim-hove: I think the `res_log_add_fmt_message( ... )` call should be unconditional - i.e. outside the `if (driver->debug_output)`test.

libres_util/src/res_log.c
```diff
@@ -16,14 +16,25 @@
...
+ * - Go over the current log-calls and make sure they use reasonable log-levels
+ * - Should we change log.h to use the normal semantics of log-levels, where higher means more important message?
```
joakim-hove: Yes - change to 'normal' - whatever that is!!

joakim-hove (approved):
Minor comments - overall ready to be merged.

libenkf/src/enkf_main.c
epa095: No, it certainly does not. A result of the simple/braindead numeric->semantic conversion I did.

libenkf/src/ert_workflow_list.c
epa095: mhm. I change it to LOG_INFO

libres_util/src/res_log.c
epa095: That must actually be done in libecl, I just used this list to remember the suggestion. I will make a task on it for libecl and do it. I will then also change the numeric values of the different levels to correspond to Pythons, so we have the same semantic as Python.

PR #34, opened 2017-06-13
Changed to use local path for test-data logging.

joakim-hove:
**Task**
Some tests left files lying around in /tmp/xxxx - this could lead to permission problems between different users.

**Approach**
Changed log path to a local path.

**Pre un-WIP checklist**


markusdregi (approved):
Good idea!

PR #37, opened 2017-06-15
Centralized logstash logging from python, part 1

epa095:
**Task**
Logging information from the *working nodes* to logstash.
This is an early test, just to get a feel for how it works

**Approach**
In this version we send explicit messages using the request library. In the future we might want to instead use the python logging framework with a HTTP or logstash backend. Then we outsource work, and can get logging to both logstash and local files. 

Currently I have only changed job_manager.py and not job_dispatcher.py, since the latter is in ertStatoil, and does not support to be deployed only to testers. But there is WIP to move job_dispatcher into libres, and then some of this code (particularly the LOG_URL) will come from there. And if we use the logging framework it makes sense to have it initialized in job_dispatcher, since it is the one calling job_manager. 

LOG_URL will clearly be changed to wherever we end up having logstash. 

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/libjob_queue/src/forward_model.c
+++ b/libjob_queue/src/forward_model.c
@@ -20,6 +20,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <ctype.h>
+#include <unistd.h>
 
 #include <ert/util/util.h>
 #include <ert/util/subst_list.h>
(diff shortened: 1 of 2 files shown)


joakim-hove (approved):

joakim-hove:
Merging this now - hopefully deploy is back on track soon!

PR #38, opened 2017-06-20
Remove Statoil specific site-config setting.

joakim-hove:
**Task**
The build system will "compile in" the default path to a site-config file. That would previously check for a Statoil specific file - that Statoil-ism has been removed. 

If this is merged we must pass `-DSITE_CONFIG_FILE=/path/to/statoil/site-config` when building for deploy.

**Approach**
Minor updates in root CMakeLists.txt

**Pre un-WIP checklist**


markusdregi:
I think this is a good idea and the approach appears to be sound! I'll have a final look when all the tests pass.

markusdregi (approved):
Looks good! Could you clean up your commits (squash the last two at least) to avoid failing commits on master?

PR #40, opened 2017-06-23
Will terminate job if illegal fil-server is used.

joakim-hove:
**Task**
This is very Statoil specific. If a forward model simulation is found to be started from a "illegal" file server the whole job can be terminated with a message to the user.

**Approach**
Send an e-mail to the user and `sys.exit(1)`

**Pre un-WIP checklist**

### DIFF
--- a/bin/job_dispatch.py
+++ b/bin/job_dispatch.py
@@ -17,6 +17,7 @@ import random
 import subprocess
 import shutil
 import datetime
+import smtplib
 from email.mime.text import MIMEText
 
 try:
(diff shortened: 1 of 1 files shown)


bin/job_dispatch.py
```diff
@@ -17,8 +17,11 @@
 import subprocess
 import shutil
 import datetime
+import smtplib
 from email.mime.text import MIMEText
 
+block_illegal_fileserver = False
```
markusdregi: Please move this assignment below the import "section"

joakim-hove: OK - done

bin/job_dispatch.py
```diff
@@ -17,19 +17,37 @@
...
 ERROR_URL     = "http://st-vlinbuild01.st.statoil.no:8000/api/error/"
 
 
+def illegal_fileserver_exit(msg_txt , user):
+    from_ = "no-reply@statoil.com"
+    to = "%s@statoil.com" % user
+    msg = MIMEText( msg_txt )
+    msg['to'] = to
+    msg['from'] = from_
+    msg['subject'] = "Illegal fil-server"
```
markusdregi, line 42: `fil-server` -> `file server`

markusdregi (approved):
When the minor typo in the message subject is fixed, this one is good to go.

I would like us to in the future restructure this so that the Statoil specifics are removed from github.

PR #42, opened 2017-06-27
Added run_id to run_context implementation.

joakim-hove:
**Task**
Added a run_id to the `ert_run_context`

**Approach**
Use time/uid/pid and some random numbers to create a random ID.

**Pre un-WIP checklist**

**Depends on**
* TNO/Seba#109
* Statoil/ert#44
* https://ci.tno.nl/gitlab/statoil/Seba/merge_requests/38

Observe that this is *tightly bound* to https://github.com/TNO/Seba/pull/109 - the commits marked with: `{libres:xxx}` should have a matching commit in the SEBA PR.

**Extended description**

When starting a bunch of simulations there is a set of metadata which describes this batch. The metadata include:

1. Cases:
   - Which case should be used for initialization.
   - Which case should be used to store the simulation results.
   - Which case should be used for updated parameters.
2. Which iteration number is this.
3. Which realisations are active.
4. Where should the realisation folders be located.
5. How should we initialize the realisations.
6. ....

All of this information is kept in an instance of `ert_run_context`, and the new `run_id` was added to this structure. The `ert_run_context` instance should have lifetime which can be illustrated with the following hypothetical `with`statement:

```python
with ErtRunContext(  ) as run_context:
    init_realisations( run_context )
    create_folders( run_context )
    submit:_simulations( run_context )
    update_realisations( run_context )
``` 

However it turned out the actual use of the `ert_run_context` type was not exactly like this; on the contrary `ert_run_context` instances were recreated from relevant underlying arguments several times one logical initialize-simulate-store-update cycle, with the consequence that the `run_id` became completely bogus. 

So the bulk of this is many changes to ascertain that the `ert_run_context` instance has the correct lifetime. Unfortunately the most fragile (i.e. not-really-tested) changes are in the gui code in Statoil/ert#44 and also the changes in Seba are 100% untested.


libenkf/src/ert_run_context.c
```diff
@@ -86,7 +90,15 @@ stringlist_type * ert_run_context_alloc_runpath_list(const bool_vector_type * ia
 }
 
 
-static ert_run_context_type * ert_run_context_alloc(bool_vector_type * iactive , run_mode_type run_mode , enkf_fs_type * init_fs , enkf_fs_type * result_fs , enkf_fs_type * update_target_fs , int iter) {
+char * ert_run_context_alloc_run_id( ) {
+  int year,month,day,hour,min,sec;
+  time_t now = time( NULL );
+  unsigned int random = util_dev_urandom_seed( );
+  util_set_datetime_values_utc( now , &sec, &min, &hour, &day, &month, &year);
+  return util_alloc_sprintf("%d:%d:%4d-%0d-%02d-%02d-%02d-%02d:%ud" , getpid() , getuid(), year , month , day , hour , min , sec, random);
```
epa095, line 106: I must admit that I get a bad feeling about embedding this information this clearly/nicely in the ID. Not that I have a problem with them being the _source_ of the ID, but when they are so nicely formatted it is easy for someone (e.g. me in the future) to think "I would like to have the uid, ohh it is right there in the ID, let me just extract it" and suddenly the ID is no longer just a id, it is actually a structure containing more information, and we are locked to keep it like that forever, or figure out who (a)buses the field and change them. I would feel better if we hashed or scrambled the data a bit first. 

If we are interested in the pid/uid/local-timestamp we should send them as explicit fields. I have no problems with that, there is no reason to be stingy with the meta-information we send in the jobs-document, as long as it is explicit.

epa095:
Can you add the new field (and maybe also the ert_pid) to the json structure in the test at "tests/res/job_queue/test_jobmanager.py" and check that it is read properly? Markus told me that I should have done this when I added the dummy "run_id" and "ert_pid", but I sinned then.

epa095:
I see that you are "piggybacking" on the name "run_id". Know that this is just a name I pulled out of my arse, I don't know if it is a name which makes sense in an ERT context. Is it? Do we already have a name for the lifespan this ID identifies, and is it "a run"? Or is it maybe one simulation (I don't know)?

We should have IDs with names for all of the "significant lifespans"(**edit**: This overview is wrong, see discussion below): 
- An ERT instance starts when the user starts the ERT program on the controller machine, and this lifespan has  ert_pid as ID. Within one ERT instance a user can press "simulate" several times, and each time she starts
- a run, which has run_id as ID, and each run start several
- realizations, which had as ID "realisation" which were extracted from the cwd field (but I stopped that, we should explicitly add it to the json instead), and each realization has several 
- iterations, which has as ID "iteration" (which was removed for same reason as above). 

Did I forget/misunderstand something, is there e.g a notion of "job" in between run and realization? Anyway, there is no problem changing the name "ert_pid" at this point, but the more data we have with that the harder it is to change, so this is  good time to come up with a reasonable name.

libenkf/src/ert_run_context.c
joakim-hove: > [....] I would like to have the uid, ohh it is right there in the ID, let me just extract it

I thought of that too - and I agree with you. Will change it.

joakim-hove:
> Can you add the new field (and maybe also the ert_pid) to the json structure in the test at "tests/res/job_queue/test_jobmanager.py" and check that it is read properly?

I will

joakim-hove:
> We should have IDs with names for all of the "significant lifespans":

Very good analysis and summary.:


> * An ERT instance starts when the user starts the ERT program on the controller machine, and this lifespan has ert_pid as ID. Within one ERT instance a user can press "simulate" several times, and each time she starts
> * a run, which has run_id as ID, and each run start several
>* realizations, which had as ID "realisation" which were extracted from the cwd field (but I stopped that, we should explicitly add it to the json instead), 
>* and each realization has several
iterations, which has as ID "iteration" (which was removed for same reason as above).

Your three first points are essentially correct, but *iteration* is wrong. When the user press "Simulate" some of actions only include *one run* [Ensemble Experiment]  - other include more than one run [Ensemble smoother], [Iterated Ensemble Smoother] and [MDA?] - each of these "runs" is called an *iteration* - so with the current approach the different *iterations/runs* in a multi iteration "button-pressed-by-user" will have separate RUN_IDs - I think that makes sense.

The realisation and iteration values should be explicitly added to the json - I agree; but those can be just numbers `0...`- I think?

joakim-hove:
>I see that you are "piggybacking" on the name "run_id". Know that this is just a name I pulled out of my arse

I like the name - I guess there are flowers growing out of your arse!

epa095:
So now I am confused about the difference between "iteration" and "run_id", are they the same (except one is a number and the other is a string)?
Also it makes sense to me that if a user starts a "Ensemble smoother" or any of the other things that start many "things", there is a common ID there which links them all, in addition to their separate IDs.

joakim-hove:
>So now I am confused about the difference between "iteration" and "run_id", are they the same (except one is a number and the other is a string)?

They represent essentially the same thing; a collection of simulations.

>Also it makes sense to me that if a user starts a "Ensemble smoother" or any of the other things that start many "things", there is a common ID there which links them all, in addition to their separate IDs.

That I certainly agree - however ...... the datastructures/abstractions representing the "run many things" object is very poor; mostly is centered around the handling of one collection of simualtions - currently managed by the `ert_run_context` implementation - and now in the process of being decorated with a spanking new `run_id`.

libenkf/src/ert_run_context.c
joakim-hove: I have not changed this yet; all in all the implementation of the `run_id` became a completely insignificant part of this PR. I still agree with @epa095 and will change the format of the `run_id`- but would like to merge this first.

libenkf/src/enkf_main.c
```diff
@@ -1806,13 +1792,14 @@ void enkf_main_run_iterated_ES(enkf_main_type * enkf_main,
 
 
 ert_run_context_type * enkf_main_alloc_ert_run_context_ENSEMBLE_EXPERIMENT(const enkf_main_type * enkf_main , enkf_fs_type * fs , bool_vector_type * iactive , int iter) {
-  return ert_run_context_alloc_ENSEMBLE_EXPERIMENT( fs , iactive , model_config_get_runpath_fmt(enkf_main_get_model_config(enkf_main)) , enkf_main_get_data_kw(enkf_main) , iter );
+  return ert_run_context_alloc_ENSEMBLE_EXPERIMENT( fs, fs , iactive , model_config_get_runpath_fmt(enkf_main_get_model_config(enkf_main)) , enkf_main_get_data_kw(enkf_main) , iter );
 }
 
-ert_init_context_type * enkf_main_alloc_ert_init_context(const enkf_main_type * enkf_main , enkf_fs_type * fs, const bool_vector_type * iactive , init_mode_type init_mode , int iter) {
+
+/*ert_init_context_type * enkf_main_alloc_ert_init_context(const enkf_main_type * enkf_main , enkf_fs_type * fs, const bool_vector_type * iactive , init_mode_type init_mode , int iter) {
```
markusdregi: Please remove this completely.

libenkf/src/ert_run_context.c
markusdregi: I've made a Jira task regarding the format of the `run_id` in an attempt to prevent that this discussion goes mute when this PR is merged..

libenkf/src/ert_run_context.c
```diff
@@ -146,12 +191,39 @@ ert_run_context_type * ert_run_context_alloc_SMOOTHER_RUN(enkf_fs_type * simulat
...
+ert_run_context_type * ert_run_context_alloc(run_mode_type run_mode,
+                                             init_mode_type init_mode,
+                                             enkf_fs_type * init_fs,
+                                             enkf_fs_type * simulate_fs ,
+                                             enkf_fs_type * target_update_fs ,
+                                             bool_vector_type * iactive ,
+                                             path_fmt_type * runpath_fmt ,
+                                             subst_list_type * subst_list ,
+                                             int iter) {
+  if (run_mode == SMOOTHER_UPDATE)
```
markusdregi: Please consider using a switch for handling various enum values. The compiler will then help us keeping track of when a specific value is not handled.

python/python/res/enkf/ert_run_context.py
```diff
@@ -69,3 +115,27 @@ def createRunpathList(cls, mask, runpath_fmt, subst_list, iter=0):
     def createRunpath(cls, iens , runpath_fmt, subst_list, iter=0):
         """ @rtype: str """
         return cls._alloc_runpath(iens, runpath_fmt, subst_list, iter)
+
+    
+    def get_id(self):
```
markusdregi, line 120: These all seems like good candidates for being properties..?

python/python/res/server/ertrpcclient.py
```diff
@@ -13,6 +13,7 @@
 
 
 def convertFault(fault):
+    print fault
```
markusdregi: Remove?

python/python/res/server/simulation_context.py
```diff
@@ -1,50 +1,53 @@
...
+        run_arg = self._run_context.iensGet( iens )
         queue = self._queue_manager.get_job_queue()
         self._run_args[iens] = run_arg
         self._thread_pool.submitJob(ArgPack(self._ert, run_arg, queue))
 
 
     def isRunning(self):
         return self._queue_manager.isRunning()
-
+
```
markusdregi, line 50: Trailing whitespace, please remove

python/python/res/server/simulation_context.py
```diff
@@ -93,3 +96,10 @@ def __repr__(self):
         fmt = '%s, #running = %d, #success = %d, #failed = %d, #waiting = %d'
         fmt =  fmt % (running, numRunn, numSucc, numFail, numWait)
         return 'SimulationContext(%s)' % fmt
+
+    def get_result_fs(self):
```
markusdregi, line 100: To me it seems like `result_fs`and `run_context` are good candidates for being properties..

python/tests/res/enkf/test_runpath_list.py
```diff
@@ -127,36 +128,44 @@ def test_assert_export(self):
...
-
-            # ensure field symlink is replaced by file
-            self.assertFalse( path.islink(linkpath) )
+#    def test_assert_symlink_deleted(self):
+#        with ErtTestContext("create_runpath2" , self.createTestPath("local/snake_oil_field/snake_oil.ert")) as tc:
+#            ert = tc.getErt( )
+#            runpath_list = ert.getRunpathList( )
+#
+#            ens_size = ert.getEnsembleSize( )
+#            runner = ert.getEnkfSimulationRunner( )
```
markusdregi: Please uncomment test

python/tests/res/server/test_rpc_service.py
```diff
@@ -40,7 +41,7 @@ def test_server_creation(self):
 
             server.stop()
 
-    def test_client_interaction(self):
+    def xtest_client_interaction(self):
```
markusdregi: Doesn't the naming `xtest..` result in this particular test never being ran?

python/tests/res/server/test_rpc_service.py
```diff
@@ -170,15 +174,19 @@ def clientInteraction():
...
             for iens in range(client_count):
                 thread = Thread(name="client_%d" % iens)
-                target_case_name = "target" if iens % 2 == 0 else ".target"
                 thread.run = createClientInteractor(target_case_name, iens)
                 threads.append(thread)
 
             for thread in threads:
                 thread.start()
-
+
```
markusdregi: Trailing whitespace

python/tests/res/server/test_rpc_storage.py
```diff
@@ -74,7 +94,7 @@ def test_rpc_storage(self):
                 self.assertEqual(data["Test:Gradient"][sim_id], gradients[sim_id])
                 self.assertEqual(data["Test:GradientDirection"][sim_id], gradient_directions[sim_id])
 
-
+
```
markusdregi: ...

python/tests/res/server/test_rpc_storage.py
```diff
@@ -84,25 +104,25 @@ def test_rpc_storage_with_simulation(self):
...
 
-    def runSimulation(self, client, realization_count, batch_id, batch_name):
-        client.startSimulationBatch("default", 2)
+    def runSimulation(self, client, realization_count, batch_id, case_name):
+        client.startSimulationBatch("default", case_name, 2)
         kw = {"SNAKE_OIL_PARAM": [0.50, 6, 1.750, 0.250, 0.990, 2, 1.770, 0.330, 0.550, 0.770]} # identical runs
 
         for iens in range(realization_count):
-            client.addSimulation(batch_name, geo_id=0, pert_id=0, sim_id=iens, keywords=kw)
+            client.addSimulation(0,0,iens,  kw)
```
markusdregi: Please fix the white space between the arguments

markusdregi (approved):
This is excellent and highly appreciated work! It ended up having much bigger ramifications then first anticipated, but these are truly important.

Good job! Ship it when the comments are addressed.

python/tests/res/server/test_rpc_service.py
joakim-hove: Ehhh - yes it does; hopefully changing it back does not give a surprise ...

joakim-hove:
Thank you for the review; I have addressed all the comment except the questions about properties. Regarding the use of properties I am generally positive; but I am reluctant to the possibility of introducing new regressions in this quite large and unwieldy patch.

markusdregi:
Very good! Please merge when the tests pass

python/python/res/server/ertrpcserver.py
```diff
@@ -137,23 +137,25 @@ def isInitializationCaseAvailable(self):
...
                 self._session.simulation_context = None
                 self._session.init_case_name = initialization_case_name
 
                 self.ert.addDataKW("<WPRO_RUN_COUNT>", str(self._session.batch_number))
                 self.ert.addDataKW("<ELCO_RUN_COUNT>", str(self._session.batch_number))
+                self._session.simulation_context = SimulationContext(self.ert, init_fs, target_fs, simulation_count, self._session.batch_number, verbose=self._verbose_queue)
                 self._session.batch_number += 1
-                self._session.simulation_context = SimulationContext(self.ert, simulation_count, verbose=self._verbose_queue)
-
+
```
pgdr: Please don't introduce new trailing white spaces.

PR #45, opened 2017-06-30
Res config test

markusdregi:
**Problem**
_The test coverage of enkf_main / res_config is not good enough!_

**Approach**
_Create a rather complete configuration with directory structure and test **all** loaded values_

**Pre un-WIP checklist**


python/tests/res/enkf/test_res_config.py
```diff
@@ -20,17 +20,50 @@
...
+        "NUM_REALIZATIONS" : 10,
+        "MAX_RUNTIME"      : 23400,
+        "MIN_REALIZATIONS" : "50%",
+        "MAX_SUBMIT"       : 13
+        }
+
+def expand_config_data():
+    for define_key in config_defines:
+        for data_key in config_data:
+            if type(config_data[data_key]) is str:
```
pgdr: `if isinstance(config_data[data_key], str): ...` and could you add a short description of what (and why) `expand_config_data` does?

test-data/local/snake_oil_structure/ert/model/user_config.ert
```diff
@@ -0,0 +1,50 @@
...
+-- ERT CONFIGURATION FILE
+--
+--===============================================================
+
+---------------------------------
+-- SECTION: Internal variables --
+---------------------------------
+DEFINE 	<USER> 		TEST_USER
+DEFINE 	<SCRATCH> 	scratch/ert
+DEFINE 	<CASE_DIR>  the_extensive_case
```
pgdr: Please search/replace tabs for spaces in this file

libenkf/src/queue_config.c
```diff
@@ -223,6 +223,11 @@ bool queue_config_init(queue_config_type * queue_config, const config_content_ty
   if (config_content_has_item(config_content, JOB_SCRIPT_KEY))
     queue_config_set_job_script(queue_config, config_content_get_value_as_abspath(config_content, JOB_SCRIPT_KEY));
 
+  if (config_content_has_item(config_content, MAX_SUBMIT_KEY)) {
+    queue_config->max_submit = config_content_get_value_as_int(config_content, MAX_SUBMIT_KEY);
+    queue_config->max_submit_set = true;
```
pgdr: why is `queue_config->max_submit_set` needed if the default is 2?

libenkf/src/gen_kw_config.c
```diff
@@ -240,6 +240,10 @@ const char * gen_kw_config_get_key(const gen_kw_config_type * config ) {
   return config->key;
 }
 
+const char * gen_kw_config_get_tag_fmt(const gen_kw_config_type * config) {
```
pgdr, line 243: should we have `gen_kw_config_get_tag_fmt` in the header file as well?

python/tests/res/enkf/test_res_config.py
markusdregi: Sure. For your own sanity, it injects all of the defines from the configuration file into the data in the dictionary. This is to allow one to copy the expected data from the configuration file without having to explicitly expand the data.

test-data/local/snake_oil_structure/ert/model/user_config.ert
markusdregi: Oh, will do! That is what you get for using an existing configuration file as a starting base..

libenkf/src/queue_config.c
markusdregi: Not entirely sure. This is legacy code from site_config, where max_submit_set (and max_submit) was duplicated and lived in both site_config and queue_config. I removed the one in site_config and simultaneously moved the loading of these variables into queue_config.

pgdr (approved):

pgdr:
Okay!  It looks very good!  Perhaps you can consider squashing Commits 3--6 (or whichever it is that make sense to squash).  Also, I'm not sure about the `snake_case_names` in the unittests since (unfortunately)  `unittest` uses `assertCamelCase`, but that's not really important.

Also, the `setUp` function is called before every test is run, and since you renamed it to `set_up_simple` that's no longer the case.  Perhaps that was the intention, though.

libenkf/src/gen_kw_config.c
markusdregi: We could. I do not really see any reason to expose it, but I wanted to create a Python binding to use in the __repr_-funtion.. I would actually prefer it to be static, but I'm not sure if this is possible with ctypes? I'll check it out

markusdregi:
@pgdr The reason I did not squash them was that I wanted to keep the commits purely fixing library code to be separate from the ones that extended the test. I would be happy to squash 3 and 5, but that is git magic above my current wizard level.

It seems like most of the test names are in snake case. And I think that since we have decided to migrate towards snake case for libres, this should also include the test code.

Yes, that was indeed the intention as it is used to set up the simple test case, while that is not what I want for the added test.

PR #58, opened 2017-07-21
Runpath to model config

markusdregi:
**Task**
_The keywords PRE_CLEAR_RUNPATH and DELETE_RUNPATH are loaded directly by enkf_main. This needs to be fixed in order to programatically instantiate an instance of enkf_main._

**Approach**
_Move the two keywords into model_config._

**Pre un-WIP checklist**

### DIFF
--- a/libenkf/include/ert/enkf/model_config.h
+++ b/libenkf/include/ert/enkf/model_config.h
@@ -68,6 +68,7 @@ extern "C" {
   void                   model_config_init(model_config_type * model_config , const config_content_type * , int ens_size , const ext_joblist_type * , int , const sched_file_type * , const ecl_sum_type * refcase);
   void                   model_config_free(model_config_type *);
   bool                   model_config_runpath_requires_iter( const model_config_type * model_config );
+  bool                   model_config_get_pre_clear_runpath(const model_config_type * model_config);
   path_fmt_type        * model_config_get_runpath_fmt(const model_config_type * );
   history_type         * model_config_get_history(const model_config_type * );
   forward_model_type   * model_config_get_forward_model( const model_config_type * );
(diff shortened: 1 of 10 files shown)


joakim-hove:
No - I'm sorry; instead of merging this we should just completely remove the functionality. The ability to configure whether the realisations should be removed or retained in the runpath directory is way to complex - and generally *not used* - people stick with the default behavior of keeping all realisations around.

markusdregi:
@joakim-hove Are we removing both the PRE_CLEAR_RUNPATH and DELETE_RUNPATH options then?

joakim-hove:
> Are we removing both the PRE_CLEAR_RUNPATH and DELETE_RUNPATH options then?

Yes - we just retain the default functionality, which is to *not* remove anything.

markusdregi:
> > Are we removing both the PRE_CLEAR_RUNPATH and DELETE_RUNPATH options then?
>
> Yes - we just retain the default functionality, which is to not remove anything.

What do you mean? All of this functionality was available in enkf_main prior to this PR.. We should either remove the loading of these keywords from enkf_main or follow the path of this PR.

joakim-hove:
> What do you mean? All of this functionality was available in enkf_main prior to this PR.. We should either remove the loading of these keywords from enkf_main or follow the path of this PR.

Sorry - that was not clear. This is about removing, or not removing, simulation folders. When I say:

> Yes - we just retain the default functionality, which is to not remove anything.

I mean that the final application should not remove any simulation folders - period. As for the changes in *code* that implies that the `PRE_CLEAR_RIUNPATH`, `DELETE_RUNPATH` and `KEEP_RUNPATH` keywords should just be removed.

markusdregi:
Aha 👍 Then we agree. I'll get to work then!

markusdregi:
The handling of the two keywords is done in https://github.com/Statoil/libres/pull/62

PR #59, opened 2017-07-21
Model config keeps NUM_REALIZATIONS

markusdregi:
**Task**
_Yet another PR on the path towards no config loading within enkf_main. Let model_config keep the number of realizations._

**Pre un-WIP checklist**

### DIFF
--- a/libenkf/src/enkf_main.c
+++ b/libenkf/src/enkf_main.c
@@ -2166,7 +2166,7 @@ static void enkf_main_init_obs(
   }
 }
 
-static void enkf_main_init_ensemble(enkf_main_type * enkf_main) {
+static void enkf_main_add_ensemble_members(enkf_main_type * enkf_main) {
   const model_config_type * model_config = enkf_main_get_model_config(enkf_main);
   int num_realizations = model_config_get_num_realizations(model_config);
   enkf_main_resize_ensemble(enkf_main, num_realizations);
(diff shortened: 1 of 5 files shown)


libenkf/src/enkf_main.c
```diff
@@ -2166,6 +2166,11 @@ static void enkf_main_init_obs(
   }
 }
 
+static void enkf_main_init_ensemble(enkf_main_type * enkf_main) {
```
joakim-hove: This is a *very limited* initialization - I fear the name creates confusion. What about: `enkf_main_add_ensemble_members( )`?

python/python/res/enkf/model_config.py
```diff
@@ -118,3 +119,7 @@ def getEnspath(self):
     def getRunpathFormat(self):
         """ @rtype: PathFormat """
         return self._get_runpath_fmt()
+
+    @property
```
joakim-hove, line 123: Discussion: are we all in on properties? I thought no!

joakim-hove (commented):
This can be merged - but:

1. The `enkf_main_init_ensemble( )` function should be renamed.

2. We should have a common agreement on the use properties.

libenkf/src/enkf_main.c
markusdregi: Sure, I'll change it

python/python/res/enkf/model_config.py
markusdregi: I thought yes. So lets have a discussion on this after stand up..

joakim-hove (approved):

PR #66, opened 2017-07-26
Enkf main no longer initializes from file

markusdregi:
**Task**
_Remove the last part of enkf_main's initialization that depends on the configuration file provided by the user._

**Approach**
_enkf_main_alloc no longer takes the filename of the configuration file as an argument._

**Pre un-WIP checklist**

**Depends on**
* Statoil/ert#47

### DIFF
--- a/python/python/res/enkf/enkf_main.py
+++ b/python/python/res/enkf/enkf_main.py
@@ -32,7 +32,7 @@ from res.enkf.key_manager import KeyManager
 
 class EnKFMain(BaseCClass):
     TYPE_NAME = "enkf_main"
-    _alloc = EnkfPrototype("void* enkf_main_alloc(char*, res_config, bool, bool)", bind = False)
+    _alloc = EnkfPrototype("void* enkf_main_alloc(res_config, bool, bool)", bind = False)
     _create_new_config = EnkfPrototype("void enkf_main_create_new_config(char* , char*, char* , int)", bind = False)
 
     _free = EnkfPrototype("void enkf_main_free(enkf_main)")
@@ -73,9 +73,16 @@ class EnKFMain(BaseCClass):
     _get_res_config = EnkfPrototype("res_config_ref enkf_main_get_res_config(enkf_main)")
 
 
-    def __init__(self, model_config, res_config=None, strict=True, verbose=True):
+    def __init__(self, model_config=None, res_config=None, strict=True, verbose=True):
         if model_config is not None and not isfile(model_config):
             raise IOError('No such configuration file "%s".' % model_config)
+
+        if model_config is not None and res_config is not None:
+            raise ValueError(
+                        "Expected either model_config or res_config to be None"
+                        )
+
+        # TODO: Deprecation warning it model_config is not None
 
         if res_config is None:
             res_config = ResConfig(model_config)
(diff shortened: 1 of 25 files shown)


python/python/res/enkf/enkf_main.py
```diff
@@ -73,27 +74,19 @@ class EnKFMain(BaseCClass):
     _get_res_config = EnkfPrototype("res_config_ref enkf_main_get_res_config(enkf_main)")
 
 
-    def __init__(self, model_config, res_config=None, strict=True, verbose=True):
-        if model_config is not None and not isfile(model_config):
-            raise IOError('No such configuration file "%s".' % model_config)
+    def __init__(self, config, strict=True, verbose=True):
```
joakim-hove, line 75: Some comment here: `config` *should* be a `ResConfig` instance, but for backward compatibility we will for a period also allow a string?!

python/python/res/enkf/enkf_main.py
```diff
@@ -106,10 +99,40 @@ def __init__(self, model_config, res_config=None, strict=True, verbose=True):
         self.__key_manager = KeyManager(self)
 
 
+    def _init_res_config(self, config):
```
joakim-hove, line 107: This can be a loose function outside of class scope.

python/python/res/server/ertrpcserver.py
```diff
@@ -39,24 +40,14 @@ def createFault(error, message):
     error_code = INVERSE_FAULT_CODES[error]
     return Fault(error_code, message)
 
-
-
 class ErtRPCServer(SimpleXMLRPCServer):
-    def __init__(self, config, host="localhost", port=0, log_requests=False, verbose_queue=False):
+    def __init__(self, enkf_main=None, res_config=None, host="localhost", port=0, log_requests=False, verbose_queue=False):
```
joakim-hove: Should maybe make it clearer from the api which of the arguments, i.e. `enkf_main` and `res_config` which is *preferred*?

python/tests/res/enkf/test_enkf.py
```diff
@@ -90,7 +90,7 @@ def test_invalid_res_config(self):
         with TestAreaContext("enkf_test") as work_area:
             with self.assertRaises(TypeError):
                 work_area.copy_directory(self.case_directory)
-                main = EnKFMain("simple_config/minimum_config", res_config=self)
+                main = EnKFMain(res_config=self)
```
joakim-hove: This looks *really* weird - `self` should be an instance of `ExtendedTestCase`??

joakim-hove (approved):
Some minor questions/comments - otherwise: GO

python/python/res/enkf/enkf_main.py
markusdregi: Exactly, it gives a deprecationwarning if the name of a configuration file is provided. I'll write a comment.

python/python/res/server/ertrpcserver.py
markusdregi: Sure thing. I've deprecated providing a configuration file and instead given the possibility of giving a res_config. I also kept the old possibility of providing an enkf_main (which seems to be the approach used in ERT at least..). Which of the options do we prefer?

python/tests/res/enkf/test_enkf.py
markusdregi: I just wanted to provoke a `TypeError` by not giving a `ResConfig` and had the test instance itself at hand.. I can change it if you like me to.

PR #70, opened 2017-07-31
Config parser refactoring

pgdr:
**Task**
Give `config_parser.py` some Pythonic love.

**Approach**
Add `__getitem__` and `__len__` and tests.   Fixes a broken (probably unused?) method `getSchemaItem` (it never returned anything).  Fixes whitespace.

**Pre un-WIP checklist**
- [X] Statoil tests pass locally
- [X] Have completed graphical integration test steps

### DIFF
--- a/python/python/res/config/config_parser.py
+++ b/python/python/res/config/config_parser.py
@@ -18,34 +18,41 @@ import sys
 import os.path
 import warnings
 
-from res.config import UnrecognizedEnum, ContentTypeEnum , ConfigContent, ConfigPrototype
 from cwrap import BaseCClass
+from res.config import UnrecognizedEnum
+from res.config import ContentTypeEnum, ConfigContent, ConfigPrototype
 
 
 class ConfigParser(BaseCClass):
     TYPE_NAME = "config_parser"
 
-    _alloc = ConfigPrototype("void* config_alloc( )", bind=False)
-    _add = ConfigPrototype("schema_item_ref config_add_schema_item( config_parser , char* , bool)")
-    _free = ConfigPrototype("void config_free( config_parser )")
-    _parse = ConfigPrototype("config_content_obj config_parse( config_parser , char* , char* , char* , char* , hash , config_unrecognized_enum , bool )")
-    _get_schema_item = ConfigPrototype("schema_item_ref config_get_schema_item( config_parser , char*)")
-    _has_schema_item = ConfigPrototype("bool config_has_schema_item( config_parser , char*)")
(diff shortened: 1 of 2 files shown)


python/tests/res/config/test_config.py
```diff
@@ -66,8 +66,8 @@ def test_item_types(self):
             schema_item.iset_type(2 , ContentTypeEnum.CONFIG_BOOL )
             schema_item.iset_type(3 , ContentTypeEnum.CONFIG_STRING )
             schema_item.iset_type(4 , ContentTypeEnum.CONFIG_PATH )
-            self.assertFalse( "TYPE_XX" in conf )
-            self.assertTrue( "TYPE_ITEM" in conf )
+            self.assertNotIn( "TYPE_XX", conf )
```
joakim-hove: Cool!

python/tests/res/config/test_config.py
```diff
@@ -91,7 +91,10 @@ def test_item_types(self):
             self.assertEqual( path_value , "file")
             self.assertEqual( type_item.igetString(4) , "file")
 
-
+            # test __getitem__
+            self.assertTrue(conf['TYPE_ITEM'])
```
joakim-hove: This is rather a test of the truthiness of the `ConfigSchemaItem`?

pgdr: Yes, you could say that, but also that  `__getitem__` is implemented, and that it returns a truthy type for the existing key `TYPE_ITEM`.  Should I remove the comment?

joakim-hove (approved):

python/python/res/config/config_parser.py
```diff
@@ -17,67 +17,86 @@
...
 
+    def add(self, keyword, required=False, value_type=None):
+        item = self._add(keyword, required).setParent(self)
+        if value_type:
+            item.iset_type(0, value_type)
         return item
 
 
-    def getSchemaItem(self , keyword):
+    def getSchemaItem(self, keyword):
```
markusdregi, line 59: I think we should we deprecate this one..

python/tests/res/config/test_config.py
joakim-hove: Fine - leave it!

python/python/res/config/config_parser.py
pgdr: done, I will merge when green

pgdr:
I have fixed downstream call sites post #66 and issued a rebuild.

See Statoil/ert#48.

PR #73, opened 2017-08-01
Workflow job PRE_SIMULATION_COPY

joakim-hove:
**Task**
The purpose of this task is to facilitate a workflow where input data to the forward model is copied in a one-off operation before the forward model starts.

**Approach**
Created a workflow job which copies a directory, and sets an environment variable 'DATA_ROOT' which is then picked up by the forward model. Complete solution requires updates to internal file-system manipulation scripts, which must be 'DATA_ROOT' aware.

**Pre un-WIP checklist**

See also this MR: https://git.statoil.no/ert/ert-statoil/merge_requests/553

### DIFF
--- a/libenkf/src/enkf_main_jobs.c
+++ b/libenkf/src/enkf_main_jobs.c
@@ -537,3 +537,18 @@ void * enkf_main_analysis_update_JOB( void * self , const stringlist_type * args
   return NULL;
 }
 
+
+void * enkf_main_pre_simulation_copy_JOB( void * self , const stringlist_type * args) {
+  const char * source_path  = stringlist_iget( args , 0 );
+  const char * target_path = stringlist_iget( args , 1 );
+
+  if (util_is_directory( source_path )) {
+    util_make_path( target_path );
+    util_copy_directory( source_path , target_path );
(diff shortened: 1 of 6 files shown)


libenkf/src/enkf_main_jobs.c
```diff
@@ -537,3 +537,15 @@ void * enkf_main_analysis_update_JOB( void * self , const stringlist_type * args
...
+
+void * enkf_main_pre_simulation_copy_JOB( void * self , const stringlist_type * args) {
+  const char * source_path  = stringlist_iget( args , 0 );
+  const char * target_path = stringlist_iget( args , 1 );
+
+  if (util_is_directory( source_path )) {
+    util_make_path( target_path );
+    util_copy_directory( source_path , target_path );
+    setenv("DATA_ROOT" , target_path, 1);
+  }
```
pgdr: No warnings if ` ! util_is_directory( source_path )` ?

joakim-hove: Added a warning - and log message.

joakim-hove: Warning and log message added.

PR #75, opened 2017-08-02
Res config from config_content

markusdregi:
**Task**
_It should be possible to allocate a res_config instance from a config_content. Currently, config_contents are created by the various sub-configs from the user configuration filename._

**Approach**
_Create a single config_content that all sub-configs can load from._

**Future work**
_We are now ready to start installing configurations from a yaml-config file into a config_content, before passing this to the initialization of res_config.__

**Pre un-WIP checklist**


pgdr (approved):
LGTM

PR #77, opened 2017-08-04
Fix fprintf() warning.

joakim-hove:
**Task**
fprintf( ) created a warning.


pgdr (approved):
Thanks, that's great!  I'll merge this as soon as it's green!

PR #78, opened 2017-08-04
Replaced ert.enkf with res.enkf

andreabrambilla:
**Task**
The enkf module has probably been moved from ert to res. Some scripts/files still refer to the old location

**Approach**
Replaced ert.enkf with res.enkf

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/docs/code/python/packages/enkf/index.rst
+++ b/docs/code/python/packages/enkf/index.rst
@@ -1,2 +1,2 @@
-The ert.enkf package
+The res.enkf package
 ====================
--- a/docs/course/ex1/sol1.py
+++ b/docs/course/ex1/sol1.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
 import sys
 import time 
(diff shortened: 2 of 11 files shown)


pgdr:
This is very good, Andrea!  But!

The tests called enkf legacy something, must retain the `ert` namespace.

Those tests are supposed to test that our `ert`-shims work.  They are temporary, though, so we hope to delete them later this year after we've released a new stable version of Ert.

There are some Python programs internally in Statoil that we don't control that use the `ert` module.  Until we have managed to port all of them to `ecl`, we must retain that compatibility.

pgdr (dismissed):
The `legacy` tests should test `ert` not `ecl`.

andreabrambilla:
@pgdr I've restored `legacy`to the previous state, please have a look when you have time

joakim-hove (approved):

PR #80, opened 2017-08-06
Check that smspec node is valid before use.

joakim-hove:
**Task**
As part of Statoil/libecl#146 it became possible to create "invalid" smspec nodes; it is now required to check the "valid" status of a node prior to using it.

**Approach**
Check that an smspec_node is valid before using it.

**Pre un-WIP checklist**


pgdr (approved):
I would prefer if a test was added that contains an invalid smspec node but due to the time sensitive nature of this task, you may merge it as it is.

PR #81, opened 2017-08-07
Remove compiler warnings

pgdr:
* initialize variable ecl_data_type * load_type
* make enkf_main_get_analysis_config return const

### DIFF
--- a/libenkf/include/ert/enkf/enkf_main.h
+++ b/libenkf/include/ert/enkf/enkf_main.h
@@ -139,7 +139,7 @@ extern "C" {
   void                          enkf_main_load_obs( enkf_main_type * enkf_main , const char * obs_config_file , bool clear_existing);
   enkf_obs_type               * enkf_main_get_obs(const enkf_main_type * );
   bool                          enkf_main_have_obs( const enkf_main_type * enkf_main );
-  analysis_config_type        * enkf_main_get_analysis_config(const enkf_main_type * );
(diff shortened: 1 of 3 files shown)


libenkf/include/ert/enkf/enkf_main.h
```diff
@@ -139,7 +139,7 @@ extern "C" {
   void                          enkf_main_load_obs( enkf_main_type * enkf_main , const char * obs_config_file , bool clear_existing);
   enkf_obs_type               * enkf_main_get_obs(const enkf_main_type * );
   bool                          enkf_main_have_obs( const enkf_main_type * enkf_main );
-  analysis_config_type        * enkf_main_get_analysis_config(const enkf_main_type * );
+  const analysis_config_type  * enkf_main_get_analysis_config(const enkf_main_type * );
```
joakim-hove, line 142: This `const` is tempting - but the `analysis_config` object will be mutated in the tui: https://github.com/Statoil/ert/blob/master/libenkf/applications/ert_tui/enkf_tui_analysis.c#L76 - but I guess the Good&trade; solution is to do `const` return as in this PR, and just remove the offending functionality/menu option from the TUI.

joakim-hove (approved):
Please add a micro fix to the TUI as well

libenkf/include/ert/enkf/enkf_main.h
pgdr: Done.

PR #82, opened 2017-08-08
Added "drop constant columns" option to the CSV export

andreabrambilla:
When using the CSV export (ERT plugins), a checkbox now allows to exclude columns whose value is the same for every entry

I couldn't test this appropriately since I don't have suitable data

**Pre un-WIP checklist**

### DIFF
--- a/share/workflows/jobs/internal-gui/config/CSV_EXPORT
+++ b/share/workflows/jobs/internal-gui/config/CSV_EXPORT
@@ -1,7 +1,7 @@
 INTERNAL    True
 SCRIPT      ../scripts/csv_export.py
 MIN_ARG     1
-MAX_ARG     4
+MAX_ARG     5
 ARG_TYPE    0  STRING
 ARG_TYPE    1  STRING
 ARG_TYPE    2  STRING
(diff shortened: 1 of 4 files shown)


joakim-hove:
> I couldn't test this appropriately since I don't have suitable data

I understand that - but at least a manual test can be performed by invoking the `bin/gui_test` in the Statoil/ert repository and actually running the plugin there.

andreabrambilla:
Of course, I've run the tests and I've launched ert on my machine (I thought that was taken for granted). And I'm building the PR on Jenkins right now. 

What I meant is that I can't properly try out the new feature in ert: If I try to export something, the new checkbox is there. With the snake_oil data I can at test the csv_export (but that data has no constant columns). For the other type of export, I don't have suitable data.

At least, I did run the code I've added in a separate script and it seems to do what's supposed to

joakim-hove:
> Of course, I've run the tests and I've launched ert on my machine 

Very good.

> (I thought that was taken for granted). 

Well .... :-)

joakim-hove (approved):

PR #83, opened 2017-08-08
Add "application":"ert" to logged json

epa095:
Add  "application":"ert" to logged json so we can filter on ert in the central log.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/python/python/res/job_queue/job_manager.py
+++ b/python/python/res/job_queue/job_manager.py
@@ -336,6 +336,8 @@ class JobManager(object):
 
         payload = {"user": self.user,
                    "cwd": os.getcwd(),
+                   "application": "ert",
+                   "subsystem": "ert_forward_model",
                    "file_server": self.isilon_node,
                    "node": self.node,
                    "start_time": self.start_time.isoformat(),


python/python/res/job_queue/job_manager.py
```diff
@@ -336,6 +336,7 @@ def postMessage(self, job=None, extra_fields={}, url=None):
 
         payload = {"user": self.user,
                    "cwd": os.getcwd(),
+                   "application": "ert",
```
joakim-hove, line 339: I would prefer e.g. `"application": "ert_forward_model"`  -this is not actually the main `ert` process logging, but rather one of the forward model components.

epa095: It is not the same process, but is it not the same application? Or "system" maybe? It is definitely the same something, and I think this 'something' is what we want to capture with 'application'. 
When I go looking at my Ert-logs is it not natural that I see the logs for both the main ert process and the forward models? 

But I also agree that there is definitely a difference between the main ert process and the forward models which should be explicit in the log-messages. One option is to add an ert-specific keyword like "ert_context" (or similar), and another option (which I am slightly partial to) is to add a more general keyword like "subprocess" / "subsystem" / "service" which in this case can be "ert_forward_model", and for other applications whatever is natural for them (including empty).

joakim-hove: It is certainly the same "system". The reason I wanted to differentiate is that the type of problems which are logged by the main ert process and the forward model jobs is *widely* different.

> When I go looking at my Ert-logs is it not natural that I see the logs for both the main ert process and the forward models?

That might be - but it is certainly important that I can easily differentiate between the two. As long as that differentiation is simple and possible I have no strong opinions.

pgdr (approved):

PR #84, opened 2017-08-09
Modify python load path in job_dispatch

joakim-hove:
**Task**
Using a non-standard version of ERT is based on setting PYTHONPATH on the workstation. Unfortunately this environment variable does not get all the way through SSH + LSF when running the forward model.

**Approach**
Update the Python load directly in the job_dispatch script.

**Pre un-WIP checklist**

### DIFF
--- a/bin/job_dispatch.py
+++ b/bin/job_dispatch.py
@@ -20,13 +20,40 @@ import datetime
 import smtplib
 from email.mime.text import MIMEText
 
+
+#-----------------------------------------------------------------
+# Here is some initial fencing which will add a an entry to the load
+# path, both the sys.path variable and the PYTHONPATH environment
+# variable. The logic below is based on the assumption:
+#
+# ${ROOT}/bin/job_dispatch.py            <-- This script
+# ${ROOT}/lib/pythonX.Y/site-packages    <-- Add this lib path
+#
+# The reason for going through these hoops is that if the user has
+# invoked some non-standard PYTHONPATH to start the main application
+# on the workstation host, that PYTHONPATH setting will not be passed
+# all the way through the SSH + LSF quagmire.
+#-----------------------------------------------------------------
+
+tmp_list = os.path.abspath( os.path.dirname( __file__ ) ).split( os.path.sep )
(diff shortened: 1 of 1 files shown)


pgdr:
I am reluctant, borderlining skeptical, to this approach.

1.  Is `.../python3.5/...` not going to be used from `3.6`?
2. Would it not be possible to send the `PYTHONPATH` environment variable over SSH using `-o SendEnv=PYTHONPATH`?

Rarely do I see a runtime `PYTHONPATH` modification that _just works_.

Do you see any other solutions for this problem?

joakim-hove:
>I am reluctant, borderlining skeptical, to this approach.

I fully agree - that was why I poked explicitly for review.

> Is .../python3.5/... not going to be used from 3.6?

That question I did not really understand?

> Would it not be possible to send the PYTHONPATH environment variable over SSH using -o SendEnv=PYTHONPATH?

That would requires server changes to `sshd`to allow recieving env variables; but yes - in principle that might be possible.

joakim-hove:
This was my best attempt - but we let it linger and discuss, hopefully something better can be found.

pgdr:
The question was:  Suppose I am running Python 3.6, and I use libraries stored somewhere in a `.../python3.5/...` location.  Then this will still ignore my local `PYTHONPATH` and add `.../python3.6/...`?

I must understand I don't grok the full problem, and I don't see any other solution (yet).  Perhaps we can talk about it tomorrow.

joakim-hove:
> Rarely do I see a runtime PYTHONPATH modification that just works.

Hehe - I agree; however I must admit I am quite keen to merge this. Right now we have a situation where the non-standard versions of ert are *unusable*. So either this - or something else, needs to be merged quite quickly.

pgdr:
> the non-standard versions of ert are _unusable_

Well, not if the user has the correct environment set up in the login shell?

Could we have a script `enable_latest` that writes `export PYTHONPATH=PATH_TO_LATEST:$PYTHONPATH` to the end of `.bashrc` and respectively for `.cshrc`?

Then all would work, right?

pgdr (approved):

PR #87, opened 2017-08-11
Queue options

andreabrambilla:
**Task**
Allow to unset keywords in ert config files

**Approach**
If QUEUE_OPTIONS is specified with only 2 parameters (instead of 3), NULL is passed to the underlying driver set() method

**Pre un-WIP checklist**

### DIFF
--- a/libjob_queue/src/queue_driver.c
+++ b/libjob_queue/src/queue_driver.c
@@ -124,6 +124,16 @@ static bool queue_driver_set_generic_option__(queue_driver_type * driver, const 
   return option_set;
 }
 
+static bool queue_driver_unset_generic_option__(queue_driver_type * driver, const char * option_key) {
+  bool option_unset = false;
+  if (strcmp(MAX_RUNNING, option_key) == 0) {
+    queue_driver_set_max_running(driver, 0);
+    option_unset = true;
+  }
+  return option_unset;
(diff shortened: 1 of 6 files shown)


libjob_queue/src/queue_driver.c
```diff
@@ -124,6 +124,16 @@ static bool queue_driver_set_generic_option__(queue_driver_type * driver, const
...
 }
 
+static bool queue_driver_unset_generic_option__(queue_driver_type * driver, const char * option_key) {
+  bool option_unset = false;
+  if (strcmp(MAX_RUNNING, option_key) == 0) {
+    queue_driver_set_max_running(driver, 0);
+    option_unset = true;
+  }
+  return option_unset;
+}
```
andreabrambilla, line 134: Hmm I wrote this before fixing the `util_sscanf_*` methods in libecl.
Now, what happens with numeric option is that, if someone tries to unset them, they are just left unchanged

Should I modify this so that the same behavior applies also to MAX_RUNNING?

joakim-hove: > what happens with numeric option is that, if someone tries to unset them, they are just left unchanged

That seems fine.

joakim-hove (approved):

PR #93, opened 2017-08-22
Remove some network-log calls

epa095:
The log-calls "runJob", "running", "startStatus", and "completeStatus"
were posted for every job, causing a lot more messages than needed. Now
we only keep post on init and OK/Exit.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/python/python/res/job_queue/job_manager.py
+++ b/python/python/res/job_queue/job_manager.py
@@ -225,7 +225,6 @@ class JobManager(object):
         with open(self.STATUS_file, "a") as f:
             now = time.localtime()
             f.write("%-32s: %02d:%02d:%02d .... " % (job["name"], now.tm_hour, now.tm_min, now.tm_sec))
-        self.postMessage(job=job, extra_fields={"status": "startStatus"})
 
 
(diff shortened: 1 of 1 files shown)


python/python/res/job_queue/job_manager.py
```diff
@@ -236,11 +235,9 @@ def completeStatus(self, exit_status, error_msg, job=None):
         with open(self.STATUS_file, "a") as f:
             if exit_status == 0:
                 status = ""
-                self.postMessage(job=job, extra_fields=extra_fields)
             else:
                 status = " EXIT: %d/%s" % (exit_status, error_msg)
                 extra_fields.update({"error_msg": error_msg})
-                self.postMessage(job=job, extra_fields=extra_fields)
```
joakim-hove, line 243: This is exit failure - that should still be posted?

epa095: We are posting in the "exit" function, which it seems to me is called every time a job fails. Then we also post the stderr

joakim-hove (approved):
Please run the Statoil test (libres-PR in Jenkins)  - then this can be merged.

PR #97, opened 2017-08-23
Network-log ecl version on init

epa095:
**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/python/python/res/job_queue/job_manager.py
+++ b/python/python/res/job_queue/job_manager.py
@@ -28,6 +28,8 @@ import pwd
 import requests
 import json
 import imp
+from ecl import EclVersion
+from res import ResVersion
 
 def redirect(file, fd, open_mode):
     new_fd = os.open(file, open_mode)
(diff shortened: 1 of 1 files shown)


epa095:
Internal tests passed on jenkins

python/python/res/job_queue/job_manager.py
```diff
@@ -122,12 +123,17 @@ def __init__(self, module_file="jobs.py", json_file="jobs.json", error_url=None,
...
         pw_entry = pwd.getpwuid(os.getuid())
         self.user = pw_entry.pw_name
+        v = EclVersion()
         logged_fields= {"status": "init",
                         "python_sys_path": sys.path,
                         "pythonpath": os.environ.get('PYTHONPATH', ''),
+                        "ecl_major_version": v.major,
+                        "ecl_minor_version": v.minor,
+                        "ecl_micro_version": v.micro,
+                        "ecl_version": v.versionString(),
```
joakim-hove: I would have thought that `"ecl_version" : v.versionString()` should suffice?

joakim-hove (dismissed):
The `EclVersion` is the least interesting `ResVersion` and `ErtVersion`are in the pipeline; maybe wait for those?

python/python/res/job_queue/job_manager.py
epa095: Yeah, I considered it, and I am still uncertain. Maybe we would like to have easy access to see who runs e.g. 2.1.$whatever, or how many runs 2.$whatever. As long as the v.versionString() format remains locked it is of course enough for that, but it feels ugly to lock that format. On the other hand, it is a fricking version string, why would we every want to change its format?

epa095:
Ref ResVersion and ErtVersion, are they in now, and/or will they be inn before feature freeze?

joakim-hove:
> Ref ResVersion and ErtVersion, are they in now, and/or will they be inn before feature freeze?

They are very much on their way in. `ResVersion` here: #91

epa095:
Changed it now to send up ecl and res version, and for each of them we only send up a single string like "2.2.git" instead of one field each for major,minor, and micro.

epa095:
Review plz  @joakim-hove

joakim-hove:
Sorry it has been slow. Joakim

PR #100, opened 2017-08-31
Jobs failing because of dead node are now restarted

andreabrambilla:
**Task**
Jobs failing because a node in the cluster dies (state = DO_KILL_NODE_FAILURE) should be restarted

**Approach**
Put thos jobs in status "EXIT" instead of "DO_KILL"
See comments at https://sdp.statoil.no/jira/browse/ERT-1239

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/libjob_queue/tests/job_queue_timeout_test.c
+++ b/libjob_queue/tests/job_queue_timeout_test.c
@@ -76,7 +76,7 @@ job_type * alloc_job(int ind, const char * cmd) {
   job->callback_usleep = 0;
   job->run_usleep      = 2 * 1000*1000; // 2 sec - should be large enough so that the queue has time to detect the timeout
   job->run_path        = util_alloc_sprintf("timeout_test_%d", ind);
-  job->cmd             = cmd;
+  job->cmd             = util_alloc_abs_path(cmd);
   job->argc            = 4;
 
   job->argv    = util_malloc(4 * sizeof *job->argv);
@@ -102,6 +102,17 @@ job_type ** alloc_jobs(int num_jobs, const char * cmd) {
   return jobs;
 }
 
+void free_jobs(job_type ** jobs, int num_jobs) {
+  for (int i = 0; i < num_jobs; i++) {
+    util_free(jobs[i]->run_path);
+    util_free(jobs[i]->cmd);
+    util_free(jobs[i]->argv[3]);
+    util_free(jobs[i]->argv);
+    util_free(jobs[i]);
+  }
+  util_free(jobs);
+}
+
 
 void submit_jobs(job_queue_type * queue, int num_jobs, job_type ** jobs) {
   for (int i = 0; i < num_jobs; i++) {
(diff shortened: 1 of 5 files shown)


joakim-hove:
This looks correct to me - just have to investigate why the test fails.

andreabrambilla:
I have stopped the build manually since it won't succeed. The test now seems fine, in the sense that the control reaches the end of ```main()```. However, the program hangs in one of the other threads. GDB says there is a ```SIGSEGV``` in ```malloc_consolidate()```:
```
#0  0x00007ffff5e535aa in malloc_consolidate () from /lib64/libc.so.6
#1  0x00007ffff5e544fe in _int_free () from /lib64/libc.so.6
#2  0x00007ffff7ded0c8 in _dl_deallocate_tls () from /lib64/ld-linux-x86-64.so.2
#3  0x00007ffff61a1e67 in __free_stacks () from /lib64/libpthread.so.0
#4  0x00007ffff61a1f7f in __deallocate_stack () from /lib64/libpthread.so.0
#5  0x00007ffff61a2f05 in start_thread () from /lib64/libpthread.so.0
#6  0x00007ffff5ed034d in clone () from /lib64/libc.so.6
```

Any suggestion is more than welcome ;)

EDIT: note that if I change the retry_callback to return false (ie, the job is not restarted) the issue doesn't occur

joakim-hove (commented):
Try stopping the queue with e.g. `job_queue_user_exit( )`

libjob_queue/tests/job_queue_timeout_test.c
```diff
@@ -133,11 +152,14 @@ int main(int argc, char ** argv) {
...
+    usleep(4 * sec); // 4.0 sec - should be large enough so that the queue has the time to restart and the exit the job
+
+    for(int ijob = 0; ijob < number_of_jobs; ++ijob) {
+       if(jobs[ijob]->retry_call_count < 1)
+         util_exit("Job's retry callback should have been called at least once, instead it was never called");
 
-    if (job_status != JOB_QUEUE_IS_KILLED) {
-      util_exit("Job should have been killed, had status %d != %d\n", job_status, JOB_QUEUE_IS_KILLED);
+     if(jobs[ijob]->exit_call_count < 1)
+          util_exit("Job's exit callback should have been called at least once, instead it was never called");
```
joakim-hove, line 172: Try just adding a call to `job_queue_manager_wait( )` and then `job_queue_free( )` here.

andreabrambilla:
It seems the issue was that the local driver was trying to write the status to the job, but the job was freed while its thread was still running.

valgrind says that the test is fine now, if it complete succesfully. Probably there is still some leak in case it fails, but I guess we don't care too much about that

joakim-hove:
> Probably there is still some leak in case it fails, but I guess we don't care too much about that

We can live with that.

joakim-hove (approved):

PR #103, opened 2017-09-01
Update log server to devnull

pgdr:
**Task**
The server got a new name.

Should we move the vlinbuild01 target as well?


joakim-hove:
Will you update this PR to just remove the old Django logging?

pgdr:
Done, I also made appropriate changes (backwards compatible) in `JobManager`.

PR #105, opened 2017-09-04
Will not automatically chdir( )

joakim-hove:
**Task**
This is part of everest task - to better use libres as a stateless library, Current code will call `chdir( )` to the location of the configuration file before parsing. With this PR that `chdir( )` must be done by the calling scope. The actual call to `chdir( )` is certainly still required - but this is of course a step in the direction of making that not required any more.

Observe that the last commit in the PR implies a change in behavior when parsing the observation files.

**Pre un-WIP checklist**

**Depends on**
* Statoil/ert#68


pgdr (approved):
LGTM

PR #106, opened 2017-09-05
Removed debug message.

joakim-hove:
**Task**
Removed debug message.


joakim-hove:
This is *trivial* - immediate self-merge.

PR #111, opened 2017-09-13
Handle config errors

markusdregi:
**Task**
_There might be errors when instantiating a res_config from a dictionary. These should be handled._

**Approach**
_Improve the python implementation of res_config accordingly._

**Pre un-WIP checklist**


pgdr (approved):
Good work!

PR #118, opened 2017-09-18
Job queue resubmit

joakim-hove:
**Task**
There have been numerous reports that the "resubmit with failing node" does not work. With this PR we just add more information, and notably forcing the `lsf_driver->debug_output` flag to `true` when an error situation is encountered. 

Intend to backport this to 2.2

**Pre un-WIP checklist**


markusdregi (approved):
LGTM

PR #119, opened 2017-09-18
Job queue enum cleanup

joakim-hove:
**Task**
This is a general cleanup of the various status codes in the queue layer. The implementation of `job_queu_status_get_count( )`has changed to accept composite states, like:
```C
int count = job_queue_status_get_count( count, JOB_STATUS_DONE + JOB_STATUS_EXIT);
```
to get the number of realisations in one of the states `JOB_STATUS_DONE` and `JOB_STATUS_EXIT`. 

As a step in the direction of cleanup the status `JOB_QUEUE_RUNNING_CALLBACK` has been split into three different statuses - depending on which callback is actually running.

**Approach**
Formatting cleanups + changed implementation of get_status_count. Added more tests.

**Pre un-WIP checklist**


libjob_queue/include/ert/job_queue/queue_driver.h
```diff
@@ -46,30 +46,91 @@ extern "C" {
...
+    JOB_QUEUE_SUBMITTED             =     4, /* Internal status: It has has been submitted - the next status update will (should) place it as pending or running. */
+    JOB_QUEUE_PENDING               =     8, /* A node which is pending - a status returned by the external system. I.e LSF */
+    JOB_QUEUE_RUNNING               =    16, /* The job is running */
+    JOB_QUEUE_DONE                  =    32, /* The job is done - but we have not yet checked if the target file is produced */
+    JOB_QUEUE_EXIT                  =    64, /* The job has exited - check attempts to determine if we retry or go to complete_fail   */
+    JOB_QUEUE_IS_KILLED             =   128, /* The job has been killed, following a  JOB_QUEUE_DO_KILL*/
+    JOB_QUEUE_DO_KILL               =   256, /* The the job should be killed, either due to user request, or automated measures - the job can NOT be restarted. */
+    JOB_QUEUE_SUCCESS               =   512,
+    JOB_QUEUE_RUNNING_DONE_CALLBACK =  1024,
+    JOB_QUEUE_RUNNING_EXIT_CALLBACK =  2048,
```
andreabrambilla, line 121: looks like the JOB_QUEUE_RUNNING_KILL_CALLBACK is missing here (or it should be removed everywhere)

libjob_queue/include/ert/job_queue/queue_driver.h
```diff
@@ -46,30 +46,91 @@ extern "C" {
...
+                     |                                        |                                                |
+                     |                                        |                                                |
+                     |                                        |                                                |
+      +--------------v------------------+    +----------------v----------------+              +----------------v----------------+
+      |                                 |    |                                 |              |                                 |
+      | JOB_QUEUE_IS_KILLED             |    | JOB_QUEUE_SUCCESS               |              |  JOB_QUEUE_FAILED               |
+      |                                 |    |                                 |              |                                 |
+      +---------------------------------+    +---------------------------------+              +---------------------------------+
+
+*/
```
andreabrambilla, line 100: this is great! :heart_eyes:

libjob_queue/src/job_queue.c
```diff
@@ -765,17 +754,12 @@ static void job_queue_handle_DO_KILL_NODE_FAILURE(job_queue_type * queue, job_qu
...
   job_queue_kill_job_node(queue, node);
-  job_queue_change_node_status(queue , node , JOB_QUEUE_RUNNING_CALLBACK );
-  {
-    arg_pack_type * arg_pack = arg_pack_alloc();
-    arg_pack_append_ptr( arg_pack , queue );
-    arg_pack_append_int( arg_pack , job_queue_node_get_queue_index(node));
-    thread_pool_add_job( queue->work_pool , job_queue_run_DO_KILL_callback , arg_pack );
-  }
+  job_queue_node_free_driver_data( node , queue->driver );
+  job_queue_change_node_status(queue , node , JOB_QUEUE_IS_KILLED );
```
andreabrambilla, line 757: I understand why the kill callback shouldn't be called on node failure, but I don't get why we shouldn't call it when the user excplicitly ask to kill the job (but I have a feeling you are considering to remove the kill callback completely...)

libjob_queue/tests/job_status_test.c
```diff
@@ -27,7 +27,7 @@
 
 void call_get_status( void * arg ) {
   job_queue_status_type * job_status = job_queue_status_safe_cast( arg );
-  job_queue_status_get_count( job_status , JOB_QUEUE_DONE + JOB_QUEUE_DO_KILL);
+  job_queue_status_get_count( job_status , 2*JOB_QUEUE_DO_KILL_NODE_FAILURE);  // This enum value is completly missing; should give util_abort.
```
andreabrambilla: what about `pow(2, JOB_QUEUE_MAX_STATE)` instead of `2*JOB_QUEUE_...`? So we don't have to change the test in case we add more states

libjob_queue/src/job_queue.c
joakim-hove: The state `JOB_QUEUE_RUNNING_KILL_CALLBACK` was removed during this PR. The point is that killing a job is done syncronously, so then it is not necessary to (temporarily) go into a callback state.

libjob_queue/tests/job_status_test.c
joakim-hove: Good idea - will update.

PR #121, opened 2017-09-19
Added res_log calls gen_data loader.

joakim-hove:
**Approach**
The `GEN_DATA` datatype is notoriosuly difficult to get right. With this PR we add more verbose logging to the `GEN_DATA` load operations.

**Approach**
Added `res_log()` calls.

**Pre un-WIP checklist**

### DIFF
--- a/libenkf/src/gen_data.c
+++ b/libenkf/src/gen_data.c
@@ -32,6 +32,8 @@
 #include <ert/ecl/ecl_file.h>
 #include <ert/ecl/ecl_util.h>
 
+#include <ert/res_util/res_log.h>
+
 #include <ert/enkf/enkf_serialize.h>
 #include <ert/enkf/enkf_types.h>
 #include <ert/enkf/enkf_macros.h>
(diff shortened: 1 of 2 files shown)


libenkf/src/gen_data.c
```diff
@@ -271,7 +273,9 @@ static bool gen_data_fload_active__(gen_data_type * gen_data, const char * filen
             util_abort("%s: error when loading active mask from:%s - file not long enough.\n",__func__ , active_file );
         }
         fclose( stream );
-      }
+        res_log_add_fmt_message( LOG_INFO , NULL , "GEN_DATA(%s): active information loaded from:%s.", gen_data_get_key( gen_data ), active_file);
+      } else
```
markusdregi, line 277: If you feel in the mood, some early exists would be appreciated. If not, let it be ;)

libres_util/src/res_log.c
```diff
@@ -113,7 +113,10 @@ void res_log_add_message(message_level_type message_level , FILE * dup_stream ,
  * required arguments to the string-formatting.
  */
 void res_log_add_fmt_message(message_level_type message_level , FILE * dup_stream , const char * fmt , ...) {
-    if (log_include_message(logh,message_level)) {
+  if(logh==NULL)
```
markusdregi, line 116: Am I right that this could indeed result in logging being done in two different files (default first, and then after the log is initialized the file specified in the config)? If so, maybe we could add a log line in the `res_log_init_log` function that warns that this behaviour can happen if `logh` is non-NULL upon initialization and just print the previous log-file in the new log_file.

I can imagine that this will be very easy to forget when debugging and will be extremely frustrating!

libenkf/src/gen_data.c
joakim-hove: Hmmm - was not in the mood right now :-(

libres_util/src/res_log.c
joakim-hove: Yes - this is correct; that is the way it has always been - and I agree we should inform about it. Not part of this PR though?!

markusdregi (approved):
I made a task regarding the logging stuff. I agree that that is work for another PR..

I have no further comments. Well done improving the logs ;) Ship it!

PR #123, opened 2017-09-20
Summary load anyway

joakim-hove:
**Task**
The mapping between report_steps and true simulation time is established when loading a summary case. It is legitimate to not have any summary keys, in that case the summary file was not loaded - and we were left without a functional mapping report_step <-> true simulation. If we later wanted to load a GEN_DATA instance that would silently fail.

**Approach**
1. Uconditionally load summary results, even if no summary keywords are requested.
2. Log an error if GEN_DATA is loaded without a properly set last_report_step value.

Will backport to 2.2

**Pre un-WIP checklist**

### DIFF
--- a/libenkf/src/enkf_state.c
+++ b/libenkf/src/enkf_state.c
@@ -653,12 +653,24 @@ static void enkf_state_internalize_GEN_DATA(enkf_state_type * enkf_state ,
                                             const model_config_type * model_config ,
                                             int last_report) {
 
-  member_config_type * my_config     = enkf_state->my_config;
-  const int  iens                    = member_config_get_iens( my_config );
   stringlist_type * keylist_GEN_DATA = ensemble_config_alloc_keylist_from_impl_type(enkf_state->ensemble_config,
                                                                                     GEN_DATA);
+
+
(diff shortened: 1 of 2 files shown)


libenkf/src/enkf_state.c
```diff
@@ -653,12 +653,24 @@ static void enkf_state_internalize_GEN_DATA(enkf_state_type * enkf_state ,
...
                                             int last_report) {
 
-  member_config_type * my_config     = enkf_state->my_config;
-  const int  iens                    = member_config_get_iens( my_config );
   stringlist_type * keylist_GEN_DATA = ensemble_config_alloc_keylist_from_impl_type(enkf_state->ensemble_config,
                                                                                     GEN_DATA);
+
+
+  if (stringlist_get_size( keylist_GEN_DATA) > 0) {
+    if (last_report <= 0) {
```
markusdregi, line 661: Should this be `last_report < 0`? I think it would increase readability and decrease "nestedness" if the two if statements were merged into one by an and.

markusdregi (approved):
One comment. Besides this it looks good. Merge as you please.

PR #125, opened 2017-09-22
Create gen data

joakim-hove:
**Task**
The results which should be loaded by Everest will typically come in the form of `GEN_DATA` instances. With this PR creation of GEN_DATA instances from memory has been simplified.

**Pre un-WIP checklist**

### DIFF
--- a/libenkf/src/enkf_config_node.c
+++ b/libenkf/src/enkf_config_node.c
@@ -410,6 +410,24 @@ enkf_config_node_type * enkf_config_node_alloc_GEN_PARAM( const char * node_key 
 }
 
 
+enkf_config_node_type * enkf_config_node_alloc_GEN_DATA_everest( const char * key ,
+                                                                 const char * result_file_fmt,
+                                                                 const int_vector_type * report_steps) {
+
+  if (!gen_data_config_valid_result_format( result_file_fmt ))
+    return NULL;
+
+  enkf_config_node_type * config_node = enkf_config_node_alloc_GEN_DATA_result( key, ASCII , result_file_fmt );
+  gen_data_config_type * gen_data_config = enkf_config_node_get_ref( config_node );
+
+  for (int i=0; i < int_vector_size( report_steps ); i++) {
+    int report_step = int_vector_iget( report_steps , i );
+    gen_data_config_add_report_step( gen_data_config , report_step);
+    enkf_config_node_set_internalize( config_node , report_step );
+  }
+
+  return config_node;
+}
 
 
 enkf_config_node_type * enkf_config_node_alloc_GEN_DATA_result( const char * key ,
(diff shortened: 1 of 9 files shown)


libenkf/src/enkf_config_node.c
```diff
@@ -410,6 +410,26 @@ enkf_config_node_type * enkf_config_node_alloc_GEN_PARAM( const char * node_key
 }
 
 
+enkf_config_node_type * enkf_config_node_alloc_GEN_DATA_everest( const char * key ,
+                                                                 const char * result_file_fmt,
+                                                                 const int_vector_type * report_steps) {
+
+  if (!gen_data_config_valid_result_format( result_file_fmt )) {
+    printf("RETURN NULL\n");
```
markusdregi: Debug statement? I suggest we turn it into a proper res_log call or remove it..

python/python/res/enkf/config/enkf_config_node.py
```diff
@@ -115,6 +116,25 @@ def create_ext_param(cls, key, key_list, output_file = None):
         keys = StringList( initial = key_list )
         return cls._alloc_ext_param_node(key, keys, output_file)
 
+
+    # This method only exposes the details relevant for Everest usage.
+    @classmethod
+    def create_gen_data(cls, key, file_fmt, report_steps = None):
+        active_steps = IntVector( )
+        if report_steps is None:
```
markusdregi: Is the default value of `active_steps` to be `IntVector([1])`? In that case I think you should reflect this by the setting the default value of `report_steps` to be `[1]`.

python/tests/res/enkf/data/test_enkf_config_node.py
```diff
@@ -0,0 +1,17 @@
...
+class EnkfConfigNodeTest(ExtendedTestCase):
+
+    def test_gen_data(self):
+
+        # Must have %d in filename argument
+        with self.assertRaises(ValueError):
+            config_node = EnkfConfigNode.create_gen_data( "KEY", "FILE" )
+
+        config_node = EnkfConfigNode.create_gen_data( "KEY", "FILE%d" )
+        self.assertIsInstance( config_node, EnkfConfigNode )
```
markusdregi, line 17: Could you please also verify the values in this test, both for the default `report_step` and for `default_step=range(10)` or something similar?

python/python/res/enkf/config/enkf_config_node.py
joakim-hove: > Is the default value of active_steps to be IntVector([1])?

Yes.

> In that case I think you should reflect this by the setting the default value of report_steps to be [1].

Well - I am under the impression that it is bad form to use mutable objects as default values in initialization calls?

libenkf/src/enkf_config_node.c
joakim-hove: > I suggest we turn it into a proper res_log call or remove it..

Removed

python/python/res/enkf/config/enkf_config_node.py
markusdregi: That is a good point! It would also be quite bad if the method actually mutate `report_steps`. But, it is better to be safe than sorry, so how about `report_step=(1)`. Tuples for the win!?

joakim-hove:
OK - all comments should be addressed.

python/python/res/enkf/config/enkf_config_node.py
joakim-hove: OK - tuple coming.

joakim-hove:
I need this now - merging.

PR #126, opened 2017-09-22
Changed fprintf(stderr,"...") to res_log call.

joakim-hove:
**Task**
Log problem with `res_log` instead of `fprintf()`


markusdregi (approved):
This is good! The more logging we do in the res_log the better..

PR #129, opened 2017-09-25
Expose config keys

markusdregi:
**Task**
_Since we can no instantiate res_config from a dict in Python it is necessary to expose the various config keys in Python as well._

**Approach**
_Make prototypes._

**Note**
_I also realized that one of the programmatic tests did not check for failed keys, so just added that in the second commit. Completely unrelated to the rest of the work in this PR. Sorry about that._

**Pre un-WIP checklist**


joakim-hove (approved):
Apart from the obvious test failure, this should be merged.

PR #130, opened 2017-09-27
Release notes 2.2

joakim-hove:
**Task**
Created release-notes for version 2.2

### DIFF
--- a/script/download-pr
+++ b/script/download-pr
@@ -4,36 +4,32 @@ import json
 import os
 import sys
 
-closed_pr = set( [1197 , 1201, 1235, 1244, 1245, 1252, 1261, 1262, 1366,
-                  1357, 1282, 1279, 1267, 1273, 1305, 1312, 1313, 1339 , 1346,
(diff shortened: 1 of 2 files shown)


script/download-pr
```diff
@@ -4,36 +4,32 @@ import json
...
+
+        sys.stdout.write("%d" % pr )
+        sys.stdout.flush()
+        prev_len = len("%s" % pr)
+
+        if response.status_code == 200:
+            pr_list.append( json.loads( response.content ) )
+        pr += 1
+
+        if pr > pr2:
```
markusdregi, line 28: I think `for pr in range(pr1, pr2+1):` instead of the `while True:` and `break` construction would be nicer..

markusdregi (approved):
One minor comment. Besides this it looks very good!

script/download-pr
joakim-hove: I agree; this script is Copy & Pasted to three repositories. Will leave it as is now.

PR #131, opened 2017-09-29
Write parameters.json

joakim-hove:
**Task**
All the parameters from the different GEN_KW parameters are generally assembled in a file `parameters.txt`. With this PR that file is duplicated with a `paramaters.json` for simpler downstream use.

**Approach**
Have added a small "container" `value_export` to temporarily hold key,value pairs and output them as txt or json. As a welcome side-effect more EnKF legacy has been removed in the `enkf_node_ecl_write( )` implementation.

**Pre un-WIP checklist**

**Depends on**

 - Statoil/ert#76


libenkf/tests/value_export.c
```diff
@@ -0,0 +1,118 @@
...
+  value_export_append(export, "KEY200", 200);
+  test_assert_int_equal( 2 , value_export_size( export ));
+  value_export_json( export );
+  test_assert_true( util_file_exists( "path/parameters.json" ));
+  value_export_free( export );
+  test_work_area_free( work_area );
+}
+
+
+void test_export_txt__() {
```
joakim-hove, line 60: This is not called!

markusdregi: So we should either call it or remove it..

markusdregi (approved):
LGTM!

PR #133, opened 2017-10-03
Will assume cwd is CONFIG_DIRECTORY not set.

joakim-hove:
**Task**
When building res_config programmatically from a dictionary cwd will be assumed if no path is given.


markusdregi:
It makes sense to default to `cwd`. Now, as we always have a `config_dir` I suggest to also always execute `internal_config.append((dir_key, config_dir))` (see line 132 of `res_config.py`). Without doing this all paths will be relative to `cwd`, but res_config never internalizes the directory and hence `res_config.config_path` will be `None` instead of `cwd`.

PR #135, opened 2017-10-06
Fixed location of .ert_runpath_list file

andreabrambilla:
Replaced <CWD> with the actual path

**Task**
Make sure that the .ert_runpath_list file is written to the same folder as the config file (unless otherwise specified in the config file itself)

**Approach**
Use the absolute path of the config file when needed

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


andreabrambilla:
@joakim-hove please have a look at the commit with the fix for RUNPATH_FILE. I've just realized replacement of REPLACEMENT_FILE seems to be broken also in the master branch. If the proposed solution is ok, I can cherry pick it on top of master

PR #140, opened 2017-10-18
Backup parameters.txt if it already exist (related to [ERT-1410])

andreabrambilla:
**Task**
If a parameters.txt file already exists, it should not interfere with the current ERT run

**Approach**
If the file exists, it is moved (backed up) first

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/libenkf/src/enkf_state.c
+++ b/libenkf/src/enkf_state.c
@@ -1241,15 +1241,23 @@ void enkf_state_ecl_write(enkf_state_type * enkf_state, const run_arg_type * run
      -----------------------------------------------------------------------------------------
   */
 
+  const shared_info_type * shared_info   = enkf_state->shared_info;
+  const model_config_type * model_config = shared_info->model_config;
+  const char * base_name                 = model_config_get_gen_kw_export_file(model_config);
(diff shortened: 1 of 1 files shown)


andreabrambilla:
Travis fails because the fix in not on the master branch, Jenkins says it's ok (https://ci.statoil.no/be/job/libres-PR/137/)

pgdr:
Shouldn't be merged in to master and ported into 2.2?

Or don't we need this patch in 2.3→?

andreabrambilla:
We probably need it in 2.3 as well, if this is fine I'll cherry pick on top of master as well

pgdr:
I would prefer this PR going towards master (then we get testing green as well) and then cherry-picked to version 2.2.

I don't have strong opinions, but I think that seems more natural.

andreabrambilla:
I have just had a look, in master we have to do things differently since `enkf_state_ecl_write()` doesn't access the files directly, but uses the `value_export_type` struct. I will prepare an analogous PR for master, but it would be anyway relatively different from this one

(anyway, Travis says no, but Jenkins says yes, is there any significant difference between those two builds?)

andreabrambilla:
PR for master is here https://github.com/Statoil/libres/pull/141

PR #143, opened 2017-10-23
Refactor enkf_obs_load into four new functions

pgdr:
The function `enkf_obs_load` was 100+ lines with several group together sections.

There should be no change in behavior due to this PR, this is only for readability and debugging purposes.

More specifically, we extract four functions:
* `handle_history_observation(enkf_obs, enkf_conf, last_report, std_cutoff);`
* `handle_summary_observation(enkf_obs, enkf_conf, last_report);`
* `handle_block_observation(enkf_obs, enkf_conf);`
* `handle_general_observation(enkf_obs, enkf_conf);`

### DIFF
--- a/libenkf/src/enkf_obs.c
+++ b/libenkf/src/enkf_obs.c
@@ -582,6 +582,159 @@ static void enkf_obs_update_keys( enkf_obs_type * enkf_obs ) {
 
 
 
+
+/*
+ *
+ * Here comes four helper functions for enkf_obs_load:
+ *
+ * * handle_history_observation(enkf_obs, enkf_conf, last_report, std_cutoff);
+ * * handle_summary_observation(enkf_obs, enkf_conf, last_report);
+ * * handle_block_observation(enkf_obs, enkf_conf);
+ * * handle_general_observation(enkf_obs, enkf_conf);
+ *
+ */
+
+
+/** Handle HISTORY_OBSERVATION instances. */
+static void handle_history_observation(enkf_obs_type * enkf_obs,
+                                       conf_instance_type * enkf_conf,
+                                       int last_report,
+                                       double std_cutoff) {
+  stringlist_type * hist_obs_keys = conf_instance_alloc_list_of_sub_instances_of_class_by_name(enkf_conf, "HISTORY_OBSERVATION");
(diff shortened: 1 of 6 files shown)


libenkf/src/enkf_obs.c
```diff
@@ -582,140 +582,194 @@ static void enkf_obs_update_keys( enkf_obs_type * enkf_obs ) {
...
+
+  if (!enkf_obs->valid)
+    return false;
+
+  int last_report                      = enkf_obs_get_last_restart( enkf_obs );
+  conf_class_type    * enkf_conf_class = enkf_obs_get_obs_conf_class();
+  conf_instance_type * enkf_conf       = conf_instance_alloc_from_file(enkf_conf_class, "enkf_conf", config_file);
+
+  if(!conf_instance_validate(enkf_conf))
+    util_abort("Can not proceed with this configuration.\n");
```
joakim-hove: These `util_abort( )` calls should all include a `%s` which resolvs to `__func__`:
```C
util_abort("%s: Can not proceed with this configuration.\n", __func__);
```

joakim-hove (approved):
Nice cleanup!

libenkf/src/enkf_obs.c
pgdr: Yes, I actually thought so ... I will update them whenever I find them.  Please stay put for another commit. :)

pgdr:
@joakim-hove I made one more change, that `enkf_obs_load` now aborts if the given config is invalid.  I updated the call site to check whether the `enkf_obs` object is valid before calling `load`.

I can remove that commit if you disagree with it, it just made more sense to me (here and now).

joakim-hove:
> that enkf_obs_load now aborts if the given config is invalid

I guess this is OK; what was the previous behavior?

pgdr:
> > that enkf_obs_load now aborts if the given config is invalid
>
> I guess this is OK; what was the previous behavior?

It returned false if `!config->valid`.  I moved that check up one level in the call stack.

I really can't defend it now either way, it just struck me as odd that the `load` function should check if `!config->valid` and if so, return false.  It seems like we have more information when doing this before the call.

PR #145, opened 2017-10-24
Fix CUSTOM_KW when current dir is not the config dir

andreabrambilla:
**Task**
When ert is executed from a folder different from where the config file is, ert will produce an error if CUSTOM_KW is used

**Approach**
Path to the result file specified in the custom_kw was determined wrongly. Now only the filename is stored, the actual path is determined when needed

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


pgdr:
Can you add a test for this specific behaviour (that fails prior to the test and works after the test)?

joakim-hove:
This looks correct to me.

python/tests/res/enkf/CMakeLists.txt
```diff
@@ -79,9 +79,9 @@ addPythonTest(tests.res.enkf.test_log_config.LogConfigTest)
 addPythonTest(tests.res.enkf.test_run_context.ErtRunContextTest)
 addPythonTest(tests.res.enkf.test_enkf_runpath.EnKFRunpathTest)
 addPythonTest(tests.res.enkf.test_programmatic_res_config.ProgrammaticResConfigTest)
+addPythonTest(tests.res.enkf.test_enkf.EnKFTest)
 
 if (STATOIL_TESTDATA_ROOT)
-  addPythonTest(tests.res.enkf.test_enkf.EnKFTest LABELS StatoilData)
```
andreabrambilla, line 84: This test is not using StatoilData, I see no reason why it should be dependent on STATOIL_DATA_ROOT

joakim-hove: Well - then it should probably be moved out of the `STATOIL_DATA_ROOT` section. We have gradually tried to reduce the Statoil data grip on the tests, so my assumption is that it used statoil data previously.

python/tests/res/enkf/test_enkf.py
```diff
@@ -231,4 +232,28 @@ def test_run_context(self):
...
             arg2 = run_context.iensGet( 2 )
             #self.assertEqual( arg0 , arg2 )
+    
+    
+    def test_run_context_from_external_folder(self):
+        with TestAreaContext('enkf_test') as work_area:
+            work_area.copy_directory(self.case_directory_custom_kw)
+            res_config_path = work_area.getFullPath('snake_oil/snake_oil.ert')
+            
+            os.chdir('/tmp')
```
joakim-hove: Make a new `TestAreaContext` inside instead of hardcoding `/tmp` - that generally comes back to bite us with permission problems.

joakim-hove (approved):

andreabrambilla:
Failed only because of runpathlist test as usual, I'm merging this

PR #146, opened 2017-10-24
WIP: Job dispatch cleanup

stefoss23:
**Task**
The job_dispatch.py (and job_manager.py class) script in libres has quite some crappy debug code from a time back when we were investigating cluster issues intensively. We should remove these no-longer required attributes:
Everything related to JobManager.fsinfo( ) is to be removed.

**Approach**
JobManager.fsinfo( ), reference to this function and consequences of the reference has been removed.

**Pre un-WIP checklist**

### DIFF
--- a/python/python/res/job_queue/job_manager.py
+++ b/python/python/res/job_queue/job_manager.py
@@ -488,31 +488,6 @@ class JobManager(object):
 
         return ('?', '?.?.?.?')
 
-    #@staticmethod
-    #def fsInfo(path = None):
-    #    if path is None:
-    #        path = os.getcwd()
-
-    #    if not os.path.isabs(path):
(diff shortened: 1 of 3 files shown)


python/python/res/job_queue/job_manager.py
```diff
@@ -500,30 +488,30 @@ def mountPoint(path):
...
         return ('?', '?.?.?.?')
 
-    @staticmethod
-    def fsInfo(path = None):
-        if path is None:
-            path = os.getcwd()
+    #@staticmethod
+    #def fsInfo(path = None):
+    #    if path is None:
+    #        path = os.getcwd()
```
joakim-hove: No - do not comment out code; delete properly!

joakim-hove (approved):

PR #134, opened 2017-10-05
Added basic everest-like test.

joakim-hove:
**Task**
Refcatoring to enable a dynamic number of realisations. See commit message for more extensive explanation.

**Approach**
A very large part of the job has been to downplay the importance of the enkf_state object - which represent a realisation which is supposed to have application lifetime. The enkf_state object still exists, but is *much* less important than previously.

**Pre un-WIP checklist**
- [x] Statoil tests pass locally
- [x] Have completed graphical integration test steps

Statoil/ert#81

PR #150, opened 2017-10-25
Tmp load nodes

joakim-hove:
**Task**
Work in progress on the way towards: #134

### DIFF
--- a/libenkf/src/enkf_state.c
+++ b/libenkf/src/enkf_state.c
@@ -657,23 +657,28 @@ static void enkf_state_internalize_GEN_DATA(enkf_state_type * enkf_state ,
       res_log_add_message( LOG_WARNING, NULL , "Trying to load GEN_DATA without properly set last_report - will only look for step 0 data.", false);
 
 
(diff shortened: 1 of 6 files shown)


libenkf/src/enkf_state.c
```diff
@@ -652,32 +641,33 @@ static void enkf_state_internalize_GEN_DATA(enkf_state_type * enkf_state ,
...
 
-      if (!enkf_node_internalize(node , report_step))
-        continue;
-      if (!enkf_node_has_func(node, forward_load_func))
+    /* 
+       This for loop should probably be changed to use the report
+       steps configured in the gen_data_config object, instead of
+       spinning through them all.
+    */
+    printf("%s: %s \n",__func__ , stringlist_iget( keylist_GEN_DATA , ikey));
```
andreabrambilla: I guess this printf is just for debugging

joakim-hove: Yes - that seems correct!

andreabrambilla (approved):
Looks good to me, even though I admit I don't fully understand the consequence of these changes

PR #153, opened 2017-10-30
Environment variables container

stefoss23:
**Task**
As part of ERT_1375 is it desirable to have a container to transfer a list of environment variables to the forward job. 

**Approach**
Such a container has been created using the hash_type object. 

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


libjob_queue/src/environment_varlist.c
```diff
@@ -0,0 +1,52 @@
...
+  hash_type * varlist;
+};
+
+env_varlist_type * env_varlist_alloc() {
+  env_varlist_type * list = util_malloc( sizeof * list );
+  list->varlist = hash_alloc();
+  return list;
+}
+
+void env_varlist_add(env_varlist_type * list, char * var_name, char * var_value) {
```
joakim-hove: The `var_name` and `var_value` arguments should `const char *`.

libjob_queue/src/environment_varlist.c
```diff
@@ -0,0 +1,52 @@
...
+
+int env_varlist_get_size(env_varlist_type * list) {
+  return hash_get_size(list->varlist);
+}
+
+char ** env_varlist_alloc_varlist(env_varlist_type * list) {
+  return hash_alloc_keylist(list->varlist);
+}
+
+char * env_varlist_get_value(env_varlist_type * list, char * var) {
```
joakim-hove: Both the return value and the argument should be `const char*`

libjob_queue/src/environment_varlist.c
```diff
@@ -0,0 +1,52 @@
...
+
+void env_varlist_add(env_varlist_type * list, char * var_name, char * var_value) {
+  hash_insert_string(list->varlist, var_name, var_value);
+}
+
+int env_varlist_get_size(env_varlist_type * list) {
+  return hash_get_size(list->varlist);
+}
+
+char ** env_varlist_alloc_varlist(env_varlist_type * list) {
```
joakim-hove: YAGNI!

libjob_queue/src/environment_varlist.c
joakim-hove: The `env_varlist_add` function should also actually set the environment variable by calling the `setenv( )` function.

joakim-hove: Should in addition add `env_varlist_add_pathvar()` function would get the current value and then reset with `setenv()`.

joakim-hove:
The main usage of this will be with the job_dispatch / job_manager.py functionality. Actual testing should be in terms of that. The ènv_varlist` container should in-itself be an untestable "black hole".

For testing it might be worth it adding a thin Python layer.

libjob_queue/src/environment_varlist.c
```diff
@@ -0,0 +1,60 @@
...
+env_varlist_type * env_varlist_alloc() {
+  env_varlist_type * list = util_malloc( sizeof * list );
+  list->varlist = hash_alloc();
+  return list;
+}
+
+
+void env_varlist_setenv(env_varlist_type * list, const char * key, const char * value) {
+  util_interp_setenv(key, value);
+  hash_insert_string(list->varlist, key, value);
```
joakim-hove: Must capture the return value from `util_interp_setenv()` - and then insert *that* value into the hash table.

libjob_queue/src/environment_varlist.c
```diff
@@ -0,0 +1,60 @@
...
+}
+
+void env_varlist_fprintf(env_varlist_type * list, FILE * stream) {
```
joakim-hove: This function should have `_json` in the name.

libjob_queue/src/environment_varlist.c
```diff
@@ -0,0 +1,60 @@
...
+  int size = hash_get_size(list->varlist);
+  fprintf(stream, "\"%s\" : {", ENV_VAR_KEY_STRING);
+  char ** keylist = hash_alloc_keylist(list->varlist);
```
joakim-hove: Use the `hash_alloc_stringlist()` function instead.

libjob_queue/src/forward_model.c
```diff
@@ -221,7 +221,8 @@ static void forward_model_json_fprintf(const forward_model_type * forward_model,
                                        const char * path,
                                        const char * data_root,
                                        const subst_list_type * global_args,
-                                       mode_t umask) {
+                                       mode_t umask,
+                                       env_varlist_type * varlist) {
```
joakim-hove: The `varlist` argument should be `const env_varlist_type *`

libjob_queue/src/forward_model.c
```diff
@@ -256,9 +258,10 @@ void forward_model_formatted_fprintf(const forward_model_type * forward_model ,
                                      const char * path,
                                      const char * data_root,
                                      const subst_list_type * global_args,
-                                     mode_t umask) {
+                                     mode_t umask,
+                                     env_varlist_type * list) {
   forward_model_python_fprintf( forward_model, path, global_args, umask);
```
joakim-hove: When you are in these parts of the code - just remove the `forward_model_python_fprintf` function completely.

python/python/res/job_queue/environment_varlist.py
```diff
@@ -0,0 +1,39 @@
...
+  _alloc                    = QueuePrototype("void* env_varlist_alloc()", bind=False)
+  _free                     = QueuePrototype("void env_varlist_free( env_varlist )")
+  _setenv                   = QueuePrototype("void env_varlist_setenv(env_varlist, char*, char*)")
+
+  def __init__(self):
+    c_ptr = self._alloc()
+    super(EnvironmentVarlist , self).__init__(c_ptr)
+
+
+  def SetEnv(self, var, value):
```
joakim-hove: This should be be `__setitem__`so that we can write:
```python
env_list["VAR"] = "VALUE"
```

python/tests/res/job_queue/test_forward_model_formatted_print.py
```diff
@@ -341,15 +342,44 @@ def test_no_jobs(self):
...
 
             self.verify_json_dump([], global_args, umask, run_id)
 
+    def test_env_varlist(self):
+        varlist_string = "global_environment"
+        first = "FIRST"
+        second = "SECOND"
+        first_value = "TheFirstValue"
+        second_value = "TheSecondValue"
+        varlist = EnvironmentVarlist()
```
joakim-hove: Please add a third variable:
```
third = "THIRD"
third_value = "$first:$second"
```
and then the test:

```python
self.assertEqual("%s:%s" % (first_value, second_value), env_config[third] )
```
this will test that the `util_setenv_interp()` function has been used correctly; which is not the case right now.

libjob_queue/src/environment_varlist.c
```diff
@@ -0,0 +1,60 @@
...
+};
+
+env_varlist_type * env_varlist_alloc() {
+  env_varlist_type * list = util_malloc( sizeof * list );
+  list->varlist = hash_alloc();
+  return list;
+}
+
+
+void env_varlist_setenv(env_varlist_type * list, const char * key, const char * value) {
```
joakim-hove: In addition a function updating a path variable should be added:
```
void env_varlist_update_path(env_varlist_type * list, const char * path_var, const char * new_path)  {
     hash_insert_string( list->varlist, path_var , util_update_path_var( path_var , new_path , false));
}
```

joakim-hove (commented):
Looks good - please make some changes.

libjob_queue/src/forward_model.c
stefoss23: That branches out to several other problems. This commit is just about the container. I suggest removing forward_model_python_fprintf as part of the complete ERT-1375 task.

joakim-hove (approved):

PR #154, opened 2017-10-31
Support programmatic GEN_DATA configuration

markusdregi:
**Task**
_Support programmatic GEN_DATA configuration._

**Approach**
_Extract and internalize._

**Pre un-WIP checklist**


joakim-hove:
This looks fine - and no reason not to merge it; but some comments:

1. The `GEN_DATA` datatype is way to flexible - and probably not all functionality should be exposed:
a) The format is always `ASCII`.
b) Output format, init_files and ecl_file are in practice never set - should not be exposed.
2. I am in a way hoping that configuring of `GEN_DATA` instances should be 100% handled programmatically - at a higher semantic level than envisioned here. My hope was that the everest application should itself infer what was needed of `GEN_DATA` instances. The inference step is completely missing, but the internalization is illustrated here: https://github.com/Statoil/libres/pull/134/files#diff-e28a4cb2a7ddf9b3f6e812b04506a16bR36

joakim-hove (approved):
If possible I would prefer if somewhat less functionality was exposed - as expressed in point 1) in review comment. Otherwise this can be merged.

markusdregi:
@joakim-hove 

1. Great! Then I'll remove the functionality to set output format, init files and ecl file in the configuration dictionary. In addition I'll default input format to `ASCII`.

2. I agree that _Everest_ should not expose `GEN_DATA`, but this is not _Everest_. Do you think that _libres_ should also not expose `GEN_DATA`? Because, then I agree that we should entirely change this. But if not, then I think that the programmatic handling of `GEN_DATA` is indeed a task for _Everest_ and not _libres_.

Thanks for the comments! I think that it is crucial that we evaluate what functionality we expose through the configuration dictionary. It is to some extent a second chance at implementing the _libres_ configuration.

joakim-hove:
> Then I'll remove the functionality to set output format, init files and ecl file in the configuration dictionary. In addition I'll default input format to ASCII.

Good - and then merge!

PR #155, opened 2017-11-03
Added subst_func.c which was orphaned in libecl

joakim-hove:
**Task**
The file `subst_func.c` was incorrectly left in libecl; moving it here.

Statoil/libecl#233


pgdr (approved):
Looks good!  Merge when green!

PR #157, opened 2017-11-03
Removed misplaced #include

joakim-hove:
**Task**
This `#include <>` should never have been there ....


joakim-hove:
It's green on Travis - does not deserve review.

PR #158, opened 2017-11-04
ENSPATH, RUNPATH, DATA_ROOT - accessed as paths

joakim-hove:
**Task**
The config settings ENSPATH, RUNPATH and DATA_ROOT are accessed as paths
from the config layer, implying that path translation is invoked if cwd
differs from the location of config file. The return value from
config_path_elm() will be normalized to remove superfluos '../' and './'
elements.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#232

### DIFF
--- a/python/tests/res/enkf/test_res_config.py
+++ b/python/tests/res/enkf/test_res_config.py
@@ -207,16 +207,13 @@ class ResConfigTest(ExtendedTestCase):
                     )
 
 
-    def assert_model_config(self, model_config, config_data, working_dir):
-        self.assertEqual(
-                config_data["RUNPATH"],
-                model_config.getRunpathAsString()
-                )
+    def assert_path(self, rel_path, config_path, model_path):
+        self.assertEqual( os.path.normpath( os.path.join( rel_path , config_path )), model_path)
 
-        self.assertEqual(
-                config_data["ENSPATH"],
-                model_config.getEnspath()
-                )
+
+    def assert_model_config(self, model_config, config_data, working_dir):
+        self.assert_path( working_dir, config_data["RUNPATH"], model_config.getRunpathAsString())
+        self.assert_path( working_dir, config_data["ENSPATH"], model_config.getEnspath() )
 
         self.assertEqual(
                 config_data["JOBNAME"],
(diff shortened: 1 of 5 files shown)


joakim-hove:
Ping

python/tests/res/enkf/test_res_config.py
```diff
@@ -316,10 +313,6 @@ def assert_ecl_config(self, ecl_config, config_data, working_dir):
                 ecl_config.getStartDate()
                 )
 
-        self.assertEqual(
-                config_data["ECLBASE"],
```
markusdregi, line 320: Why did the ECLBASE check go?

markusdregi (commented):
Please address my ECLBASE comment. Besides that, this looks very good!

markusdregi (approved):

python/tests/res/enkf/test_res_config.py
joakim-hove: The `ECLBASE` and `JOBNAME` functionality was very similar, and has been unififed in #232 - obviously the test updates fell in thw wrong PR.

markusdregi: I realized when I started looking at #232. Does this imply that we are also deprecating one of them in the interface and configuration file at some point? There is an `ECLBASE` in the config file loaded above and I would like it to be verified that it behaves as expected. If that means accessing `ecl base` or `jobname` is not important. Could we update this test either in this PR or #232?

joakim-hove: The intended behaviour is:

1. Both `JOBNAME` and `ECLBASE` can be used interchangeably in the config file. If both are present `JOBNAME`will be preferred.
2. If `ECLBASE` is present the `ecl_config_is_active( )` will return true.

I.e. the moral of the story is:

1. If you have eclipse simulations as part of your forward model - use `ECLBASE`.
2. If you do not have eclipse simulations as part of your forward model - use `JOBNAME`

I think this is tested as it should; but I guess I can look closer.

markusdregi: Ah, I understand now! The `JOBNAME` is still tested and then everything is good.

markusdregi:
Ship it!

PR #159, opened 2017-11-06
Set global environment variables from jobs json

stefoss23:
**Task**
As part of ERT-1375, environment variables need to be set from data written to the jobs.json file in each runpath. 

**Approach**
Environment variables are set in job_manager.py after loading jobs.json.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


python/python/res/job_queue/job_manager.py
```diff
@@ -171,8 +173,26 @@ def __init__(self, module_file="jobs.py", json_file="jobs.json", error_url=None,
...
+           data = self.global_environment
+           for key in data.keys():
+               os.environ[key] = data[key]
+
+    def update_path(self):
+        if (self.global_update_path):
+           data = self.global_update_path
+           for key in data.keys():
+               if (os.environ.get(key)):
+                  os.environ[key] = os.environ[key] + ':' + data[key]
```
joakim-hove: This should be turned around:
```python
os.environ[key] = data[key] + "." + os.environ[key]
```

python/python/res/job_queue/job_manager.py
```diff
@@ -136,6 +136,8 @@ def __init__(self, module_file="jobs.py", json_file="jobs.json", error_url=None,
         if log_url is None:
             self._log_url = error_url
         self._data_root = None
+        self.global_environment = None
+        self.global_update_path = None
```
joakim-hove, line 140: Here you can change the code slightly: Now the json file is *required*, and the module file should not be considered at all - i.e. the `_loadModule()`method can be completely removed.

python/python/res/job_queue/job_manager.py
```diff
@@ -171,8 +173,26 @@ def __init__(self, module_file="jobs.py", json_file="jobs.json", error_url=None,
         self.initStatusFile()
         if self._data_root:
             os.environ["DATA_ROOT"] = self._data_root
+        self.set_environment()
+        self.update_path()
         self.information = logged_fields
 
+        
+    def set_environment(self):
+         if (self.global_environment):
```
joakim-hove: Remove the paranthesis - i.e. `if self.global_environment:`

joakim-hove (approved):
This is very good - some minor comments and then this can be merged.

PR #167, opened 2017-11-13
Fixes in enkf_fs reference counting

joakim-hove:
**Task**
It seems the PR: #134 has revealed/introduced bugs with the reference counting of enkf_fs objects. When asking the `enkf_fs_manager` for a `enkf_fs` instance the reference count was not increased, depending GC ordering when things go out of scope this has seemingly worked ok - and in particular it has worked "flawlessly" on Travis (ubuntu). But on the RHEL machines it has turned up as a Heisenbug. #165 was a rather optimistic approach at fixing the same problem - think it is properly nailed now.

**Pre un-WIP checklist**


python/python/res/enkf/enkf_fs_manager.py
```diff
@@ -43,13 +43,18 @@ def atCapacity(self):
     def __contains__(self, full_case_name):
         return full_case_name in self._fs_list
 
+    def __get_fs(self, name):
+        fs = self._fs_map[name]
+        fs.incref( )
+        return fs
```
pgdr: Are you sure it is correct to increase the counter every time `getitem` is called?

Does this align with the calls to `free`?

joakim-hove: This is a quite small change in lines of code - quite significant in behaviour.

> Are you sure it is correct to increase the counter every time getitem is called?

I think so yes - but I am more than happy to be proven wrong. As I see it every `__getitem__()` call will return a new Python object, which might reference the same underlying fs instance:

```python
fs1 = fs_manager.getFileSystem("FS")
fs2 = fs_manager.getFileSystem("FS")
```
now `fs1` and `fs2` are two distinct Python objects with separate lifetime and GC behaviour, when they go out of scope get a call to the `decref( )` routine - which is essentially `free( )`.

pgdr: I don't know if I understand this completely, but I would like to see a test to verify what you're saying is correct.

Because if I run the following:
```python

        fs = self._fs_rotator[full_case_name]

        fs_copy = self._fs_rotator[full_case_name]
        fs_copy.free()

        return fs
```
then I later get `AssertionError: The EnkfFs instance has been umounted!` but perhaps I misunderstand.

I would like to see a test that verifies the count.  To me it doesn't make sense that the manager should call the fs' `incref`.  That sounds like fs' responsibility.

joakim-hove: First of all - thank you for looking properly into this, it is complicated code and I had low blood sugar and high frustration level when that code was written yesterday.

In the example:

```python
1: fs = self._fs_rotator[full_case_name]
2:
3: fs_copy = self._fs_rotator[full_case_name]
4: fs_copy.free()
5:
6: return fs
```

What happens here is:
1. The `self._fs_rotator[]` holds on to `enkf_fs` reference that has initial reference count 1.
2. The `fs` variable created at line 1 point to the same `enkf_fs`intance, the reference count is still 1.
3. The `fs_copy`variable created at line 3 points to the same underlying `enkf_fs` instance - still with referenc count 1.
4. Calling `fs_copy.free( )` will reduce the reference count to zero and discard the underlying `enkf_fs` instance.
5. Then subsequent use of `fs` will lead to `AssertionError`.

More tests coming.

joakim-hove:
OK - this has been rewritten quite considerably - I like it better now!

joakim-hove:
OK - more or less approved in Slack; merging.

PR #168, opened 2017-11-14
Robust simulator

joakim-hove:
**Task**
Create a *simulator* class which can be used from everest/seba to run simulations.

The last commit is the interesting part.

**Pre un-WIP checklist**


markusdregi:
Is the changes in the `config`-part related to the `RobustSimulator`-implementation?

Besides this connection that is not immediately clear to me, I think that the `RobustSimulator` looks very good!

joakim-hove:
> Is the changes in the config-part related to the RobustSimulator-implementation?

No; these changes are essentially no-ops. They were introduced as a means to facilitate simpler testing, but as we discussed that plan failed for other reasons.

python/python/res/config/config_content.py
```diff
@@ -232,3 +233,16 @@ def create_path_elm(self, path):
...
+
+
+    def as_dict(self):
+        d = {}
+        for key in self.keys():
+            d[key] = []
+            item = self[key]
+            for node in item:
+                d[key].append( [ x for x in node ])
+        return d
```
pgdr: I don't completely understand this, but is this equivalent to the following comprehension?
```python
return { key : [list(node) for node in self[key]] for key in self.keys() }
```

I'm not saying you should use mine, but isn't `[x for x in node]` equivalent to `node` or `list(node)`?

Then the `for node in item:` could at least be a comprehension over `self[key]`.

markusdregi:
> No; these changes are essentially no-ops. They were introduced as a means to facilitate simpler
> testing, but as we discussed that plan failed for other reasons.

Then it was as expected. I suggest we move those into a separate PR then, to get this PR really down to being about the simulator..

python/python/res/simulator/robust_simulator.py
```diff
@@ -0,0 +1,122 @@
...
+        self.control_keys = []
+        self.result_keys = []
+
+        ens_config = self.res_config.ensemble_config
+        for key in controls.keys():
+            ens_config.addNode( EnkfConfigNode.create_ext_param( key , controls[key] ))
+            self.control_keys.append(key)
+
+        for key in results:
+            ens_config.addNode( EnkfConfigNode.create_gen_data( key , "%s_%%d" % key))
```
pgdr: please use `"{}_%d".format(key)`

python/python/res/simulator/robust_simulator.py
```diff
@@ -0,0 +1,122 @@
...
+        three first are based on realisation 1, and the last is based on
+        realisation 2.
+        """
+        ens_config = self.res_config.ensemble_config
+        fsm = self.ert.getEnkfFsManager( )
+        fs = fsm.getFileSystem(case_name)
+
+        for sim_id, (geo_id, control_dict)  in enumerate(controls):
+            node_id = NodeId( 0, sim_id)
+            if not len(control_dict) == len(self.control_keys):
```
pgdr: `!=`

python/python/res/enkf/res_config.py
```diff
@@ -387,9 +386,12 @@ def _build_config_content(self, config):
...
         self._free()
 
+
     @classmethod
-    def init_config_parser(cls, config_parser):
-        cls._init_parser(config_parser)
+    def config_parser(cls):
+        parser = ConfigParser( )
+        cls._init_parser(parser)
+        return parser
```
andreabrambilla: I have a small curiosity, which is not necessarily related to this PR. Intuitively I'd say that an instance of `ResConfig` represents a configuration for libres, while an instance of `ConfigParser` represents something that parses a string and produces a config (I'd say a `ResConfig`)

Assuming my interpretation is reasonable, I find it a bit counter-intuitive that `ResConfig` is able to create an instance of `ConfigParser`

python/python/res/config/config_content.py
joakim-hove: Ehhh; I think your construction is the same - but it was quite brutal on the eyes.

python/python/res/simulator/robust_simulator.py
joakim-hove: ok - fixed

python/python/res/config/config_content.py
pgdr: But `d[key].append( [ x for x in node ])`?

python/python/res/simulator/robust_simulator.py
```diff
@@ -0,0 +1,122 @@
...
+        dictianry like this:
+
+            controls = {"cmode": ["Well","Group"], "order" : ["W1", "W2", "W3"]}
+
+        In this example the simulator will expect two arrays 'cmode' and
+        'order', consisting of two and three elements respectively. When
+        actually simualating these values will written to json files looking
+        like:
+
+             controls.json = {"Well" : 1.0, "Group" : 2.0}
```
andreabrambilla: I guess this should be `cmode.json`

python/python/res/enkf/res_config.py
joakim-hove: > Assuming my interpretation is reasonable, I find it a bit counter-intuitive that ResConfig is able to create an instance of ConfigParser

Well - I would say there is a minor twist to this; the `parser = ConfigParser()` statement will create a "naked" parser which does not know about any keywords, the parser returned by the `config_parser( )` method has been initialized to recognize all the keywords required to create a `ResConfig` instance - you could say that a schema has been configured.

python/python/res/simulator/robust_simulator.py
joakim-hove: yes - you are right.

joakim-hove:
> I suggest we move those into a separate PR then, to get this PR really down to being about the simulator..

Done; #169 has been created.

python/python/res/simulator/robust_simulator.py
```diff
@@ -0,0 +1,121 @@
...
+
+        self.res_config = res_config
+        self.ert = EnKFMain( self.res_config )
+        self.control_keys = []
+        self.result_keys = []
+
+        ens_config = self.res_config.ensemble_config
+        for key in controls.keys():
+            ens_config.addNode( EnkfConfigNode.create_ext_param( key , controls[key] ))
+            self.control_keys.append(key)
```
andreabrambilla: Just a minor comment, no need to change the code here. I am myself unsure about the following, so I take the chance to discuss it.

If you do ``` for k in d.keys: print(d[k])```, then I guess `k` has to be looked up in the dictionary at every iteration, which is potentially slow

I contrast, I would expect ```for k, v in d.iteritems(): print(v)``` to be faster since it doesn't have to lookup the value at every iteration (and I think it looks better).

Has anyone strong opinions about this? Is there any well-defined "best practice" about this?

joakim-hove: > Has anyone strong opinions about this? Is there any well-defined "best practice" about this?

I certainly don't know - actually I feel that this is one corner of Python where the "One natural & good way to do it" does not really apply.

python/python/res/simulator/robust_simulator.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+        mask = BoolVector( default_value = True, initial_size = len(controls) )
+        itr = 0
+        sim_context = RobustContext(self.result_keys, self.ert, fs, mask, itr)
+
+        for sim_id, (geo_id, control_dict)  in enumerate(controls):
+            sim_context.addSimulation(sim_id, geo_id)
+
+        return sim_context
```
andreabrambilla: I was wondering, is there any limitation on "how often" `start()`can be called? I mean, if someone writes
```
rs = RobustSimulator(config, controls, results) 
c1 = rs.start('test1', controls_1)
c2 = rs.start('test_2, controls_2)
```
then `c1` and `c2` will share the same EnKFMain instance. Can this ever be a problem?

joakim-hove: Quite a lot of effort has gone into this question; I think the `EnkfMain` instance can be safely shared in this manner - although I would not bet my firstborn on it quite yet. When it comes to more peripheral aspects - in particular the folders where simulations are run this will create problems. But those problems are quite well contained.

python/python/res/simulator/robust_simulator_context.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @property
+    def status(self):
+        """
+        Will return the state of the simulations.
+        """
+        return Status(running = self.getNumRunning(),
+                      waiting = self.getNumWaiting() + self.getNumPending(),
+                      complete = self.getNumSuccess(),
+                      failed = self.getNumFailed())
```
andreabrambilla: why not reporting waiting and pending separately?

python/python/res/enkf/res_config.py
andreabrambilla: Ok, my mistake was that I assumed the `ConfigParser` is specific for parsing res configuration, while it's actually a much more general parser. Thanks for the explanation

python/python/res/simulator/robust_simulator_context.py
joakim-hove: Waiting; implies waiting in the internal `libres` queue system, whereas Pending means that it is waiting in the LSF queue system, Could have been reported separately, but seen from the user it it is much the same thing. 

Maybe it would have been better to report them separately at this stage; then everest could later choose to collapse them - or not.

python/python/res/simulator/robust_simulator.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, res_config, controls, results):
+        """Will create simulator which can be used to run multiple simulations.
+
+        The @res_config argument should be a ResConfig object, representing the
+        fully configured state of libres.
+
+
+        The @controls argument configures which parameters the simulator should
+        get when actually simulating. The @controls argument should be a
+        dictianry like this:
```
markusdregi: typo

python/python/res/simulator/robust_simulator.py
```diff
@@ -0,0 +1,121 @@
+from ecl.util import BoolVector
+
+from res.enkf import ErtRunContext, EnKFMain, EnkfConfigNode, EnkfNode, NodeId
+from res.server import SimulationContext
+from .robust_simulator_context import RobustContext
+
+class RobustSimulator(object):
+
+    def __init__(self, res_config, controls, results):
```
markusdregi: Could we do some initial validation here of `controls` and `results`, ensuring that the first is indeed a dictionary mapping strings to lists of strings and the later a list of strings? It is at least unclear to me how this would go down if something else is provided and I think it is preferable to try to avoid the hard failures inside C when possible (especially for an API like this that is to be used mainly outside _libres_)..

python/python/res/simulator/robust_simulator.py
```diff
@@ -0,0 +1,121 @@
+from ecl.util import BoolVector
+
+from res.enkf import ErtRunContext, EnKFMain, EnkfConfigNode, EnkfNode, NodeId
+from res.server import SimulationContext
+from .robust_simulator_context import RobustContext
+
+class RobustSimulator(object):
```
markusdregi: Why is it called `RobustSimulator`? Would `BatchSimulator` be a better name or is this really tightly connected to doing robust optimization?

python/python/res/simulator/robust_simulator_context.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @property
+    def results(self):
```
markusdregi: I would prefer this to check if it is still running and then raise an error, instead of the join.

python/tests/res/simulator/test_robust_sim.py
```diff
@@ -0,0 +1,85 @@
...
+class RobustSimulatorTest(ExtendedTestCase):
+
+
+    def test_create_simulator(self):
+        config_file = self.createTestPath("local/robust_sim/robust_sim.ert")
+        with TestAreaContext("robust_sim", store_area = True) as ta:
+            sys.stderr.write("cwd:%s " % os.getcwd())
+            ta.copy_parent_content( config_file )
+            res_config = ResConfig( user_config_file = os.path.basename( config_file ))
+            rsim = RobustSimulator( res_config,
```
markusdregi: This is extremely cool!

python/python/res/simulator/robust_simulator.py
markusdregi: I agree with both of you, I too find the later to look nicer, but it does appear to deviate from the _one way_ mentality.

markusdregi (approved):
This is indeed a truly nice PR, easing the use of _libres_ dramatically in my opinion. Had some minor comments above. When those are addressed I|m ready to ship this

python/python/res/simulator/robust_simulator.py
andreabrambilla: It may be worth adding a note about this in the documentation string of this method, which will be a warning to whoever tries to use this method, and a reminder for us that this should be checked.

An alternative could be to merge `__init__` and `start()`, so that we are sure each run gets a shiny new `EnKFMain`.

I am ok with both options, up to you ;)

andreabrambilla (approved):
Looks good to me :+1:

One last comment: I actually agree with @pgdr about `results()` being a property. I'd use `@property`only for stuff that is "almost a member variable". If it takes more than 3 lines of code to compute it, I prefer having it as a method rather than a property. But once again, this is a minor thing, I won't loos my sleep over it ;)

python/python/res/simulator/robust_simulator_context.py
joakim-hove: I have changed this - reporting both seperately.

python/python/res/simulator/robust_simulator.py
joakim-hove: I have renamed

python/python/res/simulator/robust_simulator.py
joakim-hove: Comment added - would prefer the `EnkfMain`instance to stay around.

python/python/res/simulator/robust_simulator.py
joakim-hove: I have tried to test various incorrect input situations without actually resorting to `isinstance( )` - can of course extend that; but I do not see things actually entering the C layer - and crash hard - if the input is completely off. Might add more tests of course.

markusdregi: Yes, you have very good both validation and testing for incorrect input to the `start`-method. But not for the initializer as far as I can see..

joakim-hove: > But not for the initializer as far as I can see..

No - that might be - will look into it.

joakim-hove:
~~Some valgrind issues - please hold back merging.~~

PR #169, opened 2017-11-16
Minor config changes

joakim-hove:
**Task**
Minor config changes extracted from #168. I personally think these changes make sense, but they were minor steps on a larger journey which failed. So - not very interesting.

### DIFF
--- a/python/python/res/config/config_content.py
+++ b/python/python/res/config/config_content.py
@@ -159,6 +159,7 @@ class ConfigContent(BaseCClass):
     _create_path_elm = ConfigPrototype("config_path_elm_ref config_content_add_path_elm(config_content, char*)")
     _add_define = ConfigPrototype("void config_content_add_define(config_content, char*, char*)")
     _size = ConfigPrototype("int config_content_get_size(config_content)")
+    _keys = ConfigPrototype("stringlist_obj config_content_alloc_keys(config_content)")
 
     def __init__(self, filename):
         c_ptr = self._alloc(filename)
@@ -232,3 +233,16 @@ class ConfigContent(BaseCClass):
 
     def add_define(self, key, value):
         self._add_define(key, value)
+
+    def keys(self):
+        return self._keys()
+
+
+    def as_dict(self):
+        d = {}
+        for key in self.keys():
+            d[key] = []
+            item = self[key]
+            for node in item:
+                d[key].append( [ x for x in node ])
+        return d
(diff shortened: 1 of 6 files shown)


python/python/res/config/config_content.py
```diff
@@ -232,3 +233,16 @@ def create_path_elm(self, path):
...
+        return self._keys()
+
+
+    def as_dict(self):
+        d = {}
+        for key in self.keys():
+            d[key] = []
+            item = self[key]
+            for node in item:
+                d[key].append( [ x for x in node ])
```
andreabrambilla, line 247: as @pgdr suggested, this line can be rewritten as `d[key].append(list(node))`

markusdregi:
I would prefer if we hold this one back until we agree on the goal this is aiming towards. Otherwise we just add some code that is to be removed again in a couple of years when we realize that it is never used..

andreabrambilla (approved):
I checked `test_programmatic_res_config.py` only superficially, the rest looks fine to me :+1:

andreabrambilla:
@markusdregi: there is at least one of these changes that I think should be merged: being able to convert a `ConfigContent` to a dictionary frees the user from the burden of knowing what an `item` and a `node` are. It is a useful improvement of the API in my opinion

markusdregi:
> being able to convert a ConfigContent to a dictionary frees the user from the burden of knowing what an item and a node are. It is a useful improvement of the API in my opinion

Yes, but it is not clear to me that it should be a part of _libres_' Python interface to be able to either know of nor interact with the `ConfigContent`. And then, extending this interface feels odd.

Also, what Joakim was trying to achieve was to be able to go from a configuration file to a configuration dictionary. This has some advantages for testing etc, but also some hops/limitations as the dictionary does not have the exact same features as the config file. Some options are removed, while others are added/extended.

joakim-hove:
> being able to convert a ConfigContent to a dictionary

I also liked that  - but we will let it linger for now.

joakim-hove:
I already see av new use for the `config_content` to dict functionality -  I really want that!

andreabrambilla:
> Yes, but it is not clear to me that it should be a part of libres' Python interface to be able to either know of nor interact with the ConfigContent. And then, extending this interface feels odd.

@markusdregi I think I am missing some piece of information here. I know I am posting without being fully informed, so please bash me if the following comment is irrelevant or just utterly wrong. Is there an ongoing discussion about removing `ConfigContent` from the Python API? If not, we shouldn't reason about the future on `ConfigContent` here, we should just discuss whether or not the proposed change  is a valid improvement to the _current_ Python API.

Once again, I know I might be not fully aware of the current situation, sorry in advance if this comment is out of scope

markusdregi:
I agree with @andreabrambilla that it was wrong of me to start reasoning about the Python API of _libres_ in this PR. If you still want to merge this @joakim-hove, do not let my comments above stop you.

PR #172, opened 2017-11-22
Respect PATH in JOB_SCRIPT, now an EXECUTABLE

jokva:
Disclaimer: this is a ROUGH outline and not a portable, merge-ready implementation yet, it just outlines the idea.

Make job script an EXECUTABLE type key, and extend the notion of
EXECUTABLE to not only include absolute paths with an executable flag
set, but also system executables from the PATH variable.

Benefits:
* System provided JOB_SCRIPTS and other executables are more robust for
  relocations on the same system
* Jobs are more portable if they use this feature
* site-configs can use some system installed executable, rather than
  relying on some absolute path in a shared site-config
* Executable lookups follow system-wide logic, allowing users to easily
  select their own, full path to some executable, rather than being
  limited by $ERT_ROOT or other vars

Drawbacks:
* Some debugging scenarions might be harder when the wrong executable is
  accidently picked up (can be mitigated by falling back to absolute
  path)

**Task**

**Approach**

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/libconfig/src/config_content_node.c
+++ b/libconfig/src/config_content_node.c
@@ -186,6 +186,24 @@ const char * config_content_node_iget_as_relpath( config_content_node_type * nod
   }
 }
 
+const char * config_content_node_iget_as_executable( config_content_node_type * node, int index ) {
+  config_schema_item_assure_type(node->schema, index,
+        CONFIG_PATH + CONFIG_EXISTING_PATH + CONFIG_EXECUTABLE );
+  {
+    const char * config_value = config_content_node_iget(node , index);
+
+    char* path_value = NULL;
+    if( !util_file_exists( config_value ) )
+        path_value = util_alloc_PATH_executable( config_value );
+
+    if( !path_value )
+        path_value = config_path_elm_alloc_abspath( node->cwd , config_value );
+
+    config_content_node_push_string( node , path_value );
+    return path_value;
+  }
+}
+
 
 const stringlist_type * config_content_node_get_stringlist( const config_content_node_type * node ) {
   return node->stringlist;
--- a/libconfig/include/ert/config/config_content.h
+++ b/libconfig/include/ert/config/config_content.h
@@ -60,6 +60,7 @@ typedef struct config_content_struct config_content_type;
   const char * config_content_get_value_as_path( const config_content_type * config , const char * kw);
   const char * config_content_get_value_as_abspath( const config_content_type * config , const char * kw);
   const char * config_content_get_value_as_relpath( const config_content_type * config , const char * kw);
+  const char * config_content_get_value_as_executable( const config_content_type * config , const char * kw);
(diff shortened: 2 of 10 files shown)


libconfig/src/config_content_node.c
```diff
@@ -186,6 +188,50 @@ const char * config_content_node_iget_as_relpath( config_content_node_type * nod
...
+  config_schema_item_assure_type(node->schema , index ,
+        CONFIG_PATH + CONFIG_EXISTING_PATH + CONFIG_EXECUTABLE );
+  {
+    const char * config_value = config_content_node_iget(node , index);
+    const char * separator = strstr( config_value, "/" );
+
+    /* possible binary-name, consult PATH */
+    char* path_value = NULL;
+    const char* PATH = getenv( "PATH" );
+    if( !separator && PATH ) {
```
joakim-hove: There is some functionality quite close (I think) to this as`util_alloc_PATH_executable`

joakim-hove (commented):
As you mention debuggability might suffer from this - in particular the JOB_SCRIPT represents very important functionality where picking up the exact right version is quite important.

joakim-hove:
The `CONFIG_EXECUTABLE` should be used here as well: https://github.com/Statoil/libres/blob/master/libjob_queue/src/workflow_job.c#L102

libconfig/src/config_content_node.c
```diff
@@ -186,6 +188,50 @@ const char * config_content_node_iget_as_relpath( config_content_node_type * nod
   }
 }
 
+const char * config_content_node_iget_as_executable( config_content_node_type * node , int index) {
+  config_schema_item_assure_type(node->schema , index ,
+        CONFIG_PATH + CONFIG_EXISTING_PATH + CONFIG_EXECUTABLE );
+  {
+    const char * config_value = config_content_node_iget(node , index);
+    const char * separator = strstr( config_value, "/" );
```
joakim-hove: Think this should be equivalent:
```
if (!access_OK(config_value, X_OK))
   path_value = util_alloc_PATH_executable( config_value );

if (!path_value)
   path_value = config_path_elm_alloc_abspath( node->cwd , config_value );
```

libconfig/src/config_content_node.c
```diff
@@ -186,6 +186,24 @@ const char * config_content_node_iget_as_relpath( config_content_node_type * nod
   }
 }
 
+const char * config_content_node_iget_as_executable( config_content_node_type * node, int index ) {
+  config_schema_item_assure_type(node->schema, index,
+        CONFIG_PATH + CONFIG_EXISTING_PATH + CONFIG_EXECUTABLE );
+  {
```
joakim-hove, line 192: Even with local lexical scoping - #tears.in.my.eyes

joakim-hove (approved):

PR #173, opened 2017-11-23
Set GEO_ID immediately and resolve it consistently

markusdregi:
**Task**
_Using `<GEO_ID>` in job arguments or the runpath fails due to setting `run_arg->geo_id` too late after f234eea807d5681cd20337553cdc96ff5f0d693b._

**Approach**
This PR does two things:
 - Set the `geo_id` immediately when adding a simulation and hence recognize the stated relation between the given `iens` and `geo_id` as early as possible.
- Let `run_arg` have responsibility for a substitution lists that contains all keys relevant for that specific run. Use this one consistently when initializing the runpath.

**Pre un-WIP checklist**

### DIFF
--- a/libenkf/include/ert/enkf/run_arg.h
+++ b/libenkf/include/ert/enkf/run_arg.h
@@ -36,9 +36,29 @@ UTIL_SAFE_CAST_HEADER( run_arg );
 UTIL_IS_INSTANCE_HEADER( run_arg );
 
 
-  run_arg_type * run_arg_alloc_ENSEMBLE_EXPERIMENT(const char * run_id, enkf_fs_type * sim_fs, int iens , int iter , const char * runpath, const char * job_name);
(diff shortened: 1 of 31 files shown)


joakim-hove:
Have tried hard to manouver into a situation where the `ert_run_context` and `run_arg` variables can be *immutable*. The `geo_id` was already an afterthought which had to be bolted on, so in that sense this PR does not represent a change. I would still like to impose a slight feeling of unease when a setter is applied on the `run_arg` datatype, whereas in this PR applying the setter on the `run_arg` is made more elegant. All in all - I would like this PR to be less elegant - apart from that it is ready to be merged when all tests are green.

In the not-so-distant future it is my ambition to refactor so that `ert_run_context / run_arg` can be fully immutable, but we must go via this intermediate step.

joakim-hove (approved):

joakim-hove:
Please verify both the statoil internal tests and also the interactive gui test, when that has been done this is ready to be merged.

markusdregi:
I'm always running Statoil internal tests, so they are all passing. I'm finishing a test to ensure that the runpath list is indeed written and updating the code accordingly.

PR #176, opened 2017-11-28
Conservatively consult PATH

jokva:
The PATH consultation aborts whenever it constructs a path with a
non-existant directory, and should be consulted more conservatively. To
accomplish that, the alloc_relocate helper function returns absolute
paths, not relative, as they quickly contain to many back-references to
form a completely valid path.

The POSIX stat function considers ../../../usr/bin an invalid path,
where the shells resolve it to /usr/bin

**Task**

**Approach**

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/libconfig/src/config_content_node.c
+++ b/libconfig/src/config_content_node.c
@@ -203,7 +203,6 @@ const char * config_content_node_iget_as_executable( config_content_node_type * 
     }
 
     config_content_node_push_string( node , path_value );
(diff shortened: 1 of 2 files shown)


joakim-hove:
This looks fine - but the Statoil internal tests seem to fail.

markusdregi (approved):
Ship when all relevant tests are green.

PR #178, opened 2017-11-28
Set run id env

joakim-hove:
**Task**
Set environment variable `ERT_RUN_ID` for forward model jobs to pick up - if they are interested.


joakim-hove:
This is trivial - self merging.

PR #184, opened 2017-12-01
Added stop method to SimualtionContext

joakim-hove:
**Task**
Batch simulator had incorrect documentation string - fixed that.

Added `stop( )`method to `SimulationContext`

**Pre un-WIP checklist**

### DIFF
--- a/libjob_queue/include/ert/job_queue/job_queue_manager.h
+++ b/libjob_queue/include/ert/job_queue/job_queue_manager.h
@@ -31,6 +31,7 @@ typedef struct job_queue_manager_struct job_queue_manager_type;
   job_queue_manager_type * job_queue_manager_alloc( job_queue_type * job_queue );
   void job_queue_manager_free( job_queue_manager_type * manager );
   void job_queue_manager_start_queue( job_queue_manager_type * manager , int num_total_run , bool verbose);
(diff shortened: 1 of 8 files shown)


pgdr:
The title of this PR seems misleading.  It seems to include a new exposed symbol??

pgdr:
It seems like this feature doesn't actually require exposing this new function out in the header file.

joakim-hove:
Title us misleading yes.

Will look at the header, but as long as it is used from Python I would say it is deeply facto exported and should be in the header?

pgdr:
My point was that you can achieve the same thing _without_ exposing a new symbol.

Either use `queue.killAllJobs()` which seems to do the same, except has a harsher name, or use
`queue._start_user_exit()` and then `queue.isRunning()`.

joakim-hove:
> My point was that you can achieve the same thing without exposing a new symbol.

OK - I see your point; I personally like the new approach based on the `queue_manager` implementation better - the coupling to the actual queue implementation becomes looser. I have updated other call sites in libres to use the new symbol. The `JobQueue.killAllJobs( )` method is still present because it is used from the main ert application, but I would like to change that callsite as well - and then remove the `JobQueue.killAllJobs( )` method (that will have to be in a new PR).

PR #185, opened 2017-12-01
ext_job: writing arg specs to jobs.json

stefoss23:
**Task**
 ext_job: writing arg specs to jobs.json: The following needs to be written to jobs.json:
arg min
arg max
arg_types (int, str etc.)

**Approach**
Changes is ext_job_json_fprintf.

**Pre un-WIP checklist**

### DIFF
--- a/libjob_queue/src/ext_job.c
+++ b/libjob_queue/src/ext_job.c
@@ -226,7 +226,6 @@ ext_job_type * ext_job_alloc(const char * name , const char * license_root_path 
 
 
 
-
 ext_job_type * ext_job_alloc_copy(const ext_job_type * src_job) {
   ext_job_type * new_job  = ext_job_alloc__( src_job->name , src_job->license_root_path , true /* All copies are by default private jobs. */);
 
@@ -244,6 +243,10 @@ ext_job_type * ext_job_alloc_copy(const ext_job_type * src_job) {
 
   new_job->max_running_minutes   = src_job->max_running_minutes;
   new_job->max_running           = src_job->max_running;
+  new_job->min_arg               = src_job->min_arg;
+  new_job->max_arg               = src_job->max_arg;
+  new_job->arg_types             = int_vector_alloc_copy(src_job->arg_types);
+
   new_job->private_args          = subst_list_alloc_deep_copy( src_job->private_args );
 
   /* Copying over all the keys in the environment hash table */
@@ -301,6 +304,8 @@ void ext_job_free(ext_job_type * ext_job) {
   stringlist_free(ext_job->argv);
   stringlist_free(ext_job->deprecated_argv);
   subst_list_free( ext_job->private_args );
+
+  int_vector_free(ext_job->arg_types);
   free(ext_job);
 }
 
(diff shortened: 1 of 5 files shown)


python/python/res/job_queue/ext_job.py
```diff
@@ -143,17 +143,19 @@ def get_max_running_minutes(self):
...
         return self._min_arg()
-
-    def maximumArgumentCount(self):
+    
+    @property
+    def max_arg(self):
         return self._max_arg()
 
-    def argumentTypes(self):
+    def get_argument_types(self):
```
joakim-hove: Well;

if we are on the property bandwagon I suggest:
   
@property
def arg_types(self):
     ...

libjob_queue/src/ext_job.c
```diff
@@ -743,6 +748,40 @@ static void __fprintf_python_argList(FILE * stream,
...
+                                       const char * prefix,
+                                       const char * key,
+                                       const ext_job_type * ext_job,
+                                       const char * suffix,
+                                       const char * null_value) {
+  fprintf(stream, "%s", prefix);
+  if (ext_job->arg_types) {
+    fprintf(stream, "\"%s\" : [", key);
+    int i;
+    for (i = 0; i < ext_job->max_arg; i++) {
```
joakim-hove, line 761: This should loop over the size of arg_types - which can very well be less than max_arg.

libjob_queue/src/ext_job.c
```diff
@@ -743,6 +748,40 @@ static void __fprintf_python_argList(FILE * stream,
...
+      switch(type) {
+        case CONFIG_INT:    arg_type = JOB_INT_TYPE; break;
+        case CONFIG_FLOAT:  arg_type = JOB_FLOAT_TYPE; break;
+        case CONFIG_STRING: arg_type = JOB_STRING_TYPE; break;
+        case CONFIG_BOOL:   arg_type = JOB_BOOL_TYPE; break;
+        default: util_abort("ERROR in %s", __func__);
+      }
+
+      fprintf(stream, "\"%s\"", arg_type);      
+      if ((i + 1) < ext_job->max_arg)
```
joakim-hove, line 774: Again - consult length of arg_types.

joakim-hove (approved):
Very good - some minor comments; then she goes in!

libjob_queue/src/ext_job.c
stefoss23: I am not sure about that. arg_types should be set in the forward model config file and cannot be changed by the user (from the ert config file).

libjob_queue/src/ext_job.c
stefoss23: As above - unsure.

libjob_queue/src/ext_job.c
joakim-hove: The point is that you do not have to specify type for all the arguments, i.e. you can have the following situation:
```txt
MIN_ARG 1
MAX_ARG 3
ARG_TYPES 0 INT
```
In this case the function should have 1-3 arguments, the first argument should be an integer - *you do not care* about the two next arguments - meaning that they will be `CONFIG_STRING`.

I think the best solution is to use your implementation, but replace `int_vector_iget`with `int_vector_safe_iget()` which will return the default value - i.e. `CONFIG_STRING` if you ask for an out-of-range value.

PR #187, opened 2017-12-13
Added new user config keyword SIMULATION_JOB and functionality.

stefoss23:
**Task**
A new user config keyword is introduced: SIMULATION_JOB. 
This is similar to kw FORWARD_MODEL, except

ARGLIST in the forward model config file is ignored. The arguments are to be added in the user config file.
SYNTAX:
INSTALL_JOB                    THE_JOB      path_to_jof_file
SIMULATION_JOB            THE_JOB      ARG1 ARG2 etc.

The old and deprecated keyword FORWARD_MODEL will still be able to coexist with the new improved functionality.

**Approach**
Changes primarily in model_confic.c and forward_model.c allowing for per-line reading of both keywords. 

**Pre un-WIP checklist**


markusdregi:
I'll have a closer look at this later on. I just have a quick question: is it really so that the `FORWARD_MODEL` keywords is to be deprecated (now or in the future) or are the two to coexist possibly forever?

markusdregi:
And a second comment, wasn't the new job to be called `SIMULATION_JOB` (as written in the description above) and not `SIMULATION_MODEL` as presented in the title and code?

stefoss23:
1) The old FORWARD_MODEL keywords have to weeded out from every job config file, so this will take some time. 
2) My error, I started with SIMLATION_MODEL and forgot to change it to SIMULATION_JOB.

markusdregi:
> > The old FORWARD_MODEL keywords have to weeded out from every job config file, so this will take some time.

I did not know that we are deprecating the old one. But I think it is a good idea and I like that we are taking such active measures.

libenkf/src/model_config.c
```diff
@@ -450,12 +450,26 @@ void model_config_init(model_config_type * model_config ,
...
     char * config_string = config_content_alloc_joined_string( config , FORWARD_MODEL_KEY , " ");
+    fprintf(stdout, "XXXXXXX: %s: %s\n", __func__, config_string);
     forward_model_parse_init( model_config->forward_model , config_string );
     free(config_string);
+  }*/
+
+  
+  for (int i = 0; i < config_content_get_size(config); i++) {
+     config_content_node_type * node = config_content_iget_node( config , i);
+     if (strcmp(config_content_node_get_kw(node), SIMULATION_MODEL_KEY) == 0) {
```
joakim-hove: 1. Use the `util_string_equal()` function.
2. Please avoid the use of `{ .. }`

libjob_queue/include/ert/job_queue/ext_job.h
```diff
@@ -37,8 +37,9 @@ void                    ext_job_set_help_text( ext_job_type * job , const char *
 
 void                    ext_job_fprintf_config(const ext_job_type * ext_job , const char * fmt , FILE * stream);
 ext_job_type          * ext_job_alloc_copy(const ext_job_type * );
+void                    ext_job_set_not_deprecated(ext_job_type * ext_job); //DEPRECATED
```
joakim-hove: `not_deprecated` was a quite funny name?

libjob_queue/src/ext_job.c
```diff
@@ -148,6 +147,7 @@ struct ext_job_struct {
 
   bool              private_job;           /* Can the current user/delete this job? (private_job == true) means the user can edit it. */
   bool              __valid;               /* Temporary variable consulted during the bootstrap - when the ext_job is completely initialized this should NOT be consulted anymore. */
+  bool              deprecated;
```
joakim-hove: Hmmm - smells a bit that we need to internalize this?

joakim-hove (commented):
Will look more tomorrow  -but this looks very good.

libjob_queue/src/ext_job.c
stefoss23: When using ext_job_json_fprintf, we have to know whether to print from argv or deprecated_argv.
Alternative: 
1) We do a logic check of the population of deprecated_argv and argv.
2) We do not actually need deprecated_argv.

joakim-hove (approved):

joakim-hove:
Good work

PR #188, opened 2017-12-14
Default lsf

joakim-hove:
Will default the LSF driver to submit using the method `LSF_LOCAL_SHELL`

### DIFF
--- a/libjob_queue/src/lsb.c
+++ b/libjob_queue/src/lsb.c
@@ -147,6 +147,9 @@ void lsb_free( lsb_type * lsb) {
 
 
 bool lsb_ready( const lsb_type * lsb) {
+  if (!lsb)
+    return false;
+
   return lsb->ready;
 }
 
--- a/libjob_queue/src/lsb.c
+++ b/libjob_queue/src/lsb.c
@@ -1,25 +1,25 @@
 /*
-   Copyright (C) 2011  Statoil ASA, Norway. 
-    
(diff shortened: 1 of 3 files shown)


markusdregi (approved):
Could we have a condition on HAVE_LSF_DRIVER inside the test as well, verifying that it is indeed the correct driver that is set? Or do we expect this lsb_ready condition to be false every now and then?

Otherwise, it looks good to me!

joakim-hove:
> Could we have a condition on HAVE_LSF_DRIVER inside the test as well, verifying that it is indeed the correct driver that is set? 

Getting this LSF library functionality correct has been quite tricky, to actually use the internal lsf driver two requirements must be satisifed:

1. You need to have access to the lsf library and header files while building - that is essentally controlled by the `#ifdef HAVE_LSF_LIBRARY`

2. At runtime you need to be on host which is recognized as part of an official LSF cluster, otherwise the `lsb_init()` function will fail. Typically this second requirement fails - and this imposes hard restrictions on the test; if you are on computer so-and-so the result should be this, otherwise it should be that. 

The library submit method is in many ways the cleanest, no parsing of stdout files, assembling of command lines and so on, but the system is extremely picky that is actually hardly ever used. In hindsight the library submit method would probably not have been implemented at all.

markusdregi:
Thanks for the thourough feedback! Merge at will..

PR #189, opened 2017-12-17
Default lsf submit method should be LOCAL SHELL if no lsf lib.

joakim-hove:
**Task**
Will check that the default lsf submit method equals `LOCAL_SHELL` when the lsf library is *not* available.

**Pre un-WIP checklist**


markusdregi (approved):
LGTM

PR #192, opened 2017-12-20
Runpath list load

joakim-hove:
**Task**
The `runpath_list` implementation is mainly used to export a list of simulation directories to a file; with this PR it can also load from the same file.

**Approach**
Added quite basic load function.

**Pre un-WIP checklist**


joakim-hove:
@kjellkongsvik : with this merged you should be able to do something like:
```python
main = EnkfMain( res_config )
runpath_list = main.create_runpath_list( )
runpath_list.load( )

for node in runpath_list:
     ecl_sum = EclSum( os.path.join( node.runpath, node.basename ))
```

markusdregi (approved):
This is a very natural feature I would say. The implementation looks solid.

Merge at will.

PR #199, opened 2018-01-04
Give random seed in config file

markusdregi:
**Task**
_Functionality for providing the random seed directly in the configuration file._

**Approach**
- Support keyword `RANDOM_SEED` in configuration file and configuration dict.
- If both `RANDOM_SEED` and `LOAD_SEED` is provided, `RANDOM_SEED` takes precedence.
- The `RANDOM_SEED` is written to the log when `enkf_main` is instantiated, so that one can replicate the current results if wanted.

**Remarks**
Do we want to deprecate `LOAD_SEED` and `STORE_SEED` or should both possibilities live side by side? Both are feasible solutions. The later is the current status, but deprecation is only a log-message away..

Note that most of the diff is new tests. 

**Pre un-WIP checklist**

### DIFF
--- a/libenkf/src/rng_config.c
+++ b/libenkf/src/rng_config.c
@@ -125,7 +125,7 @@ rng_manager_type * rng_config_alloc_rng_manager( const rng_config_type * rng_con
   rng_manager_type * rng_manager;
 
   if (seed_load && util_file_exists( seed_load ))
-    rng_manager = rng_manager_alloc( seed_load );
+    rng_manager = rng_manager_alloc_load( seed_load );
   else
     rng_manager = rng_manager_alloc_random( );
 
--- a/libenkf/src/rng_config.c
+++ b/libenkf/src/rng_config.c
@@ -25,6 +25,8 @@
 #include <ert/config/config_parser.h>
 #include <ert/config/config_schema_item.h>
 
+#include <ert/res_util/res_log.h>
+
 #include <ert/enkf/rng_config.h>
 #include <ert/enkf/rng_manager.h>
 #include <ert/enkf/config_keys.h>
@@ -34,11 +36,21 @@
 
 struct rng_config_struct {
   rng_alg_type      type;
+  char            * random_seed;
   char            * seed_load_file;    /* NULL: Do not store the seed. */
   char            * seed_store_file;   /* NULL: Do not load a seed from file. */
 };
 
 
+static void rng_config_set_random_seed(rng_config_type * rng_config, const char * random_seed) {
+  util_safe_free(rng_config->random_seed);
+  // TODO: Handle odd sizes
+  rng_config->random_seed = util_alloc_string_copy(random_seed);
+}
+
+const char * rng_config_get_random_seed(const rng_config_type * rng_config) {
+  return rng_config->random_seed;
+}
 
 void rng_config_set_type( rng_config_type * rng_config , rng_alg_type type) {
   rng_config->type = type;
@@ -69,6 +81,7 @@ static rng_config_type * rng_config_alloc_default(void) {
   rng_config_type * rng_config = util_malloc( sizeof * rng_config);
 
   rng_config_set_type( rng_config , MZRAN );  /* Only type ... */
+  rng_config->random_seed = NULL;
   rng_config->seed_store_file = NULL;
   rng_config->seed_load_file = NULL;
 
@@ -101,6 +114,7 @@ rng_config_type * rng_config_alloc(const config_content_type * config_content) {
 void rng_config_free( rng_config_type * rng) {
   util_safe_free( rng->seed_load_file );
   util_safe_free( rng->seed_store_file );
+  util_safe_free( rng->random_seed );
   free( rng );
 }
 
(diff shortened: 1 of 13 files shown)


joakim-hove:
I will look into this later today.

joakim-hove:
> Do we want to deprecate LOAD_SEED and STORE_SEED or should both possibilities live side by side? Both are feasible solutions. The later is the current status, but deprecation is only a log-message away..

My personal wish is to take it even further: just remove the `LOAD_SEED` and `STORE_SEED` functionality altogether. This is rarely used expert functionality, the new functionality is equivalent - if not identical, the gain in reduced complexity is just too tempting. In addition to actually removing the load/store functionality from the code you can add a message to the config system [here](https://github.com/Statoil/libres/blob/master/libenkf/src/rng_config.c#L129):
```
config_install_message(parser, CONFIG_LOAD_SEED_KEY, "The LOAD_SEED functionality has been removed - use the alternative keyword `RANDOM_SEED'");
```

libenkf/src/rng_config.c
```diff
@@ -101,6 +130,7 @@ rng_config_type * rng_config_alloc(const config_content_type * config_content) {
 void rng_config_free( rng_config_type * rng) {
   util_safe_free( rng->seed_load_file );
   util_safe_free( rng->seed_store_file );
+  util_safe_free( rng->random_seed );
```
joakim-hove: The `util_safe_free( )` originates in a misunderstanding - it is safe to call `free( NULL )`.

libenkf/src/rng_config.c
```diff
@@ -110,11 +140,24 @@ rng_manager_type * rng_config_alloc_rng_manager( const rng_config_type * rng_con
...
+  else if (seed_load && util_file_exists( seed_load ))
+    rng_manager = rng_manager_alloc_load( seed_load );
   else
     rng_manager = rng_manager_alloc_random( );
 
+  res_log_add_fmt_message(LOG_CRITICAL, stdout, "To repeat this experiment, add the following random seed to your config file:");
+  if (rng_config->random_seed) {
+    res_log_add_fmt_message(LOG_CRITICAL, stdout, "RANDOM_SEED %s", rng_config->random_seed);
+  } else {
+    unsigned int random_seed [4];
```
joakim-hove: The lines 152-159 could be replaced with a call: `rng_config_log_seed( rng_config )`- which again unconditionally called the internal rng to get state; should not be necessary to distinguish between seeds explicitly provided by the user and default initialized seeds?? If this is indeed possible we can also remove the `rng_manager_get_state()` function completely.

libenkf/src/rng_config.c
```diff
@@ -34,11 +36,37 @@
...
   rng_alg_type      type;
+  char            * random_seed;
   char            * seed_load_file;    /* NULL: Do not store the seed. */
   char            * seed_store_file;   /* NULL: Do not load a seed from file. */
 };
 
 
+static void rng_config_set_random_seed(rng_config_type * rng_config, const char * random_seed) {
+  util_safe_free(rng_config->random_seed);
+  rng_config->random_seed = util_alloc_string_copy(random_seed);
```
joakim-hove: These two lines can be collapsed to: 
```
rng_config->random_seed = util_realloc_string_copy( rng_config->random_seed, random_seed );
```

libenkf/src/rng_manager.c
```diff
@@ -104,6 +124,10 @@ rng_type * rng_manager_iget(rng_manager_type * rng_manager, int index) {
 }
 
 
+void rng_manager_get_state(const rng_manager_type * rng_manager, char * state) {
```
joakim-hove: Maybe `rng_manager_export_state( )` is better - since the function actually is `void`! But see other comment in `rng_config.c` - unless I have misunderstood something else I would actually recommend the complete removal of this function.

joakim-hove:
The `rng_manager.c` implementation has a couple of magic '4' values sprinkled around; I am *not* suggesting we should go fully general and ask the rng for state size, but could at least introduce a slightly documented symbol:
```
#define RNG_STATE_SIZE 4
```

joakim-hove (changes requested):
This all over very good - but I would prefer if the `LOAD_SEED` and `STORE_SEED` functionality was completely removed.

markusdregi:
@joakim-hove The reason I kept both `RANDOM_SEED` and `LOAD_SEED` around is that wouldn't the immediate removal of a key in the configuration file impose a bump of the major version if staying true to semver?

If we want the `LOAD_SEED` and `STORE_SEED` functionality gone we can either just remove it completely as you suggest or we can deprecate it (which is technically very easy now) and remove it next time we do an appropriate change of the version. I'll let you decide and will update my code accordingly.

libenkf/src/rng_config.c
markusdregi: The thing is that the journey from a random seed in the configuration file to the internal state of the `rng_manager` is not injective. Due to this, if specifying `RANDOM_SEED 1234` in the configuration file, you might very well be asked to add `123412341234123412341234123412341234123` to the configuration file to repeat the experiment, which is kind of odd and unexpected.

joakim-hove: >Due to this, if specifying RANDOM_SEED 1234 in the configuration file, you might very well be asked to add 123412341234123412341234123412341234123 to the configuration file to repeat the experiment, which is kind of odd and unexpected.

Well - we can live with that - in my opinon.

joakim-hove:
> If we want the LOAD_SEED and STORE_SEED functionality gone we can either just remove it completely as you suggest or we can deprecate it (which is technically very easy now) and remove it next time we do an appropriate change of the version. I'll let you decide and will update my code accordingly.

OK - I think it should be removed in the long run; so then I suggest deprecation now.

joakim-hove (approved):

joakim-hove:
Resolve comments as you see fit - and this should be ready to merged.

libenkf/src/rng_config.c
markusdregi: Aha, I never really understood the reason for `util_safe_free` to exist. If you agree, we could deprecate it then?

markusdregi:
Addressed all comments. Will merge when Travis and Jenkins are green.

PR #204, opened 2018-01-10
Replace <CASE_NAME> in path with batch id

pgdr:
Make the BatchSimulator replace occurrences of `<CASE_NAME>` in paths with actual case names.

We slugify the paths in case they should contain slashes.

### DIFF
--- a/python/python/res/simulator/batch_simulator.py
+++ b/python/python/res/simulator/batch_simulator.py
@@ -3,6 +3,10 @@ from ecl.util import BoolVector
 from res.enkf import ResConfig, ErtRunContext, EnKFMain, EnkfConfigNode, EnkfNode, NodeId
 from .batch_simulator_context import BatchContext
 
+def _slug(entity):
+    entity = ' '.join(str(entity).split())
+    return ''.join([x if x.isalnum() else '_' for x in entity.strip()])
+
 class BatchSimulator(object):
 
     def __init__(self, res_config, controls, results):
@@ -95,6 +99,7 @@ class BatchSimulator(object):
         batch complete before you start a new batch.
         """
         ens_config = self.res_config.ensemble_config
+        self.ert.addDataKW("<CASE_NAME>", _slug(case_name))
         fsm = self.ert.getEnkfFsManager( )
         fs = fsm.getFileSystem(case_name)
 
--- a/python/tests/res/simulator/test_batch_sim.py
+++ b/python/tests/res/simulator/test_batch_sim.py
@@ -114,13 +114,21 @@ class BatchSimulatorTest(ResTest):
                                     "WELL_ON_OFF" : ["W1","W2", "W3"]},
                                    ["ORDER", "ON_OFF"])
 
+            case_name = 'MyCaseName_123'
             # Starting a simulation which should actually run through.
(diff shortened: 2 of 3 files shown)


joakim-hove:
The code you are touching now - i.e. the `<xxx>` replacements - is one of the most notoriously crappy parts of libres/ert - a test would be good!

python/python/res/simulator/batch_simulator.py
```diff
@@ -3,6 +3,13 @@
 from res.enkf import ResConfig, ErtRunContext, EnKFMain, EnkfConfigNode, EnkfNode, NodeId
 from .batch_simulator_context import BatchContext
 
+def _slug(entity):
+    entity = ' '.join(str(entity).strip().split())
+    for i in range(len(entity)):
```
markusdregi: Couldn't the for-loop be replaced by something along the lines of:
`return map(lambda x : x if x.isalnum() else '_', entity)`?
Possibly a bit more Pythonic?

python/python/res/simulator/batch_simulator.py
```diff
@@ -95,6 +102,7 @@ def start(self, case_name, controls):
         batch complete before you start a new batch.
         """
         ens_config = self.res_config.ensemble_config
+        self.ert.addDataKW("<CASE_NAME>", _slug(case_name))
```
markusdregi, line 102: I would like that we make `<CASE_NAME>` a part of the API of the `BatchSimulator` by documenting the behavior.

pgdr:
> The code you are touching now - i.e. the `<xxx>` replacements - is one of the most notoriously crappy parts of libres/ert - a test would be good!

I agree that tests are always good, but running `start` is not within the scope of a unit test, nor should we test the actual string replace functionality here.

I can do as Markus asks, documenting the behaviour, but other than that, testing the `start` functionality is not suitable for a unit test.  Thus, we have to fall back to a sort of integration test.  If we proceed along those lines, how would you like that to be set up?

joakim-hove:
> nor should we test the actual string replace functionality here.

Well - not exactly the actual low-level string replacement; that should certainly not be tested here. What has created problems numerous times in the past is related to ordering of the type *this* function must be called after *that* function which .... and the `BatchSimulator` is a new way of running, i.e. unfortunately again exposed to the possibility of these f.ups.

> If we proceed along those lines, how would you like that to be set up?

Let me look around a bit; if it is reasonably simple we can add a test - otherwise I am happy to merge as is.

joakim-hove:
> Let me look around a bit; if it is reasonably simple we can add a test - otherwise I am happy to merge as is.

OK - my suggestion is to modify [this test](https://github.com/Statoil/libres/blob/master/python/tests/res/simulator/test_batch_sim.py). If the config file which is referenced has a `RUNPATH` with `<CASE_NAME>` in you can just verify that runpath directory - with an expanded `<CASE_NAME>` actually exists when the simulation has stopped [here](https://github.com/Statoil/libres/blob/master/python/tests/res/simulator/test_batch_sim.py)

joakim-hove (commented):
All good!

joakim-hove (approved):

PR #206, opened 2018-01-14
Remove unused type LARS (Least Angle Regression)

pgdr:
Correct me if I'm wrong, but I believe LARS is not used in neither libres nor ert.

Removed 
* `libanalysis/include/ert/analysis/lars.h`
* `libanalysis/src/lars.c`
* entry `src/lars.c` from `libanalysis/CMakeLists.txt`


joakim-hove:
> [...] I believe LARS is not used in neither libres nor ert.

That is correct - it was written as support code for some experimental work on external analysis modules, that work is unfortunately not active any more. Merging.

joakim-hove (approved):

PR #212, opened 2018-01-19
Relabel log levels, higher level is more important

### LINKED ISSUE #211: Crash on log level CRITICAL
```
Activity will be logged to ..............: (null)

Error message: log_add_message: logh->stream == NULL - must call
log_reset_filename() first

See file: /tmp/ert_abort_dump.pgdr.20180118-160917.log for more details of the crash.
Setting the environment variable "ERT_SHOW_BACKTRACE" will show the backtrace on stderr.
Aborted (core dumped)
```

```
[pgdr@be-lx885510 everest/model]$ cat /tmp/ert_abort_dump.pgdr.20180118-160917.log

Abort called from: log_add_message
(...cache/libres-master/libres_util/src/log.c:160)

Error message: log_add_message: logh->stream == NULL - must call
log_reset_filename() first


----------------------------------------
 #00 util_abort__(..)                   in ???
 #01 log_add_message(..)                in ???
 #02 res_log_add_fmt_message(..)        in ???
 #03 rng_manager_log_state(..)          in ???
 #04 rng_config_alloc_rng_manager(..)   in ???
 #05 enkf_main_rng_init(..)             in ???
 #06 enkf_main_alloc(..)                in ???
 #07 ffi_call_unix64(..)                in /var...
 #08 ffi_call(..)                       in /var...
 #09 _ctypes_callproc(..)               in /var...
 #10 ???(..)                            in /var...
 #11 PyObject_Call(..)                  in /var...
 #12 PyEval_EvalFrameEx(..)             in /var...
```

pgdr:
**Task**
Fix #211

Update log levels according to more conventional design where a _higher number_ means _more important_ message.

Specifically, we set

| Level   | Value | Use  |
|---      |---    |---   |
| DEBUG   | 10    | Detailed information, typically of interest only when diagnosing problems.     |
| INFO    | 20    | Confirmation that things are working as expected.     |
| WARNING | 30    | An indication that something unexpected happened, or indicative of some problem in the near future (e.g. ‘disk space low’). The software is still working as expected.     |
| ERROR   | 40    |   Due to a more serious problem, the software has not been able to perform some function.   |
| CRITICAL| 50    |   A serious error, indicating that the program itself may be unable to continue running.   |

**Approach**
In the enum, `LOG_CRITICAL=0` and I suspect that somewhere in libres there is a
```c
if (log_level)
  log_open(logh);
```


libres_util/include/ert/res_util/log.h
```diff
@@ -29,31 +29,47 @@ extern "C" {
...
-  log_type   * log_open(const char *filename, int log_level);
-  void         log_add_message(log_type *logh, int message_level , FILE * dup_stream , char* message, bool free_message);
+  log_type   * log_open(const char *filename, message_level_type log_level);
+  void         log_add_message(log_type *logh, message_level_type message_level , FILE * dup_stream , char* message);
   void         log_add_message_str(log_type *logh, message_level_type message_level , const char* message);
-  void         log_add_fmt_message(log_type * logh , int message_level , FILE * dup_stream , const char * fmt , ...);
-  int          log_get_level( const log_type * logh);
-  void         log_set_level( log_type * logh , int new_level);
+  void         log_add_fmt_message(log_type * logh , message_level_type message_level , FILE * dup_stream , const char * fmt , ...);
+  message_level_type          log_get_level( const log_type * logh);
```
joakim-hove: Please maintain the indentation which is currently in the code.

libres_util/src/res_log.c
```diff
@@ -112,24 +110,31 @@ void res_log_add_message(message_level_type message_level , FILE * dup_stream ,
...
+  va_end(ap);
 }
 
 
-void res_log_close(){
-    if (log_is_open( logh ))
-      log_add_message( logh , false , NULL , "Exiting ert application normally - all is fine(?)" , false);
+void res_log_close() {
+    if (log_is_open(logh))
+      log_add_message(logh,
```
joakim-hove: I agree that many of the lines in the code are too long, and appreciate the splitting; in particular I was not aware of the possibility to split string literals like `"part1" "part2"` - that was very good to learn. However splitting the current `log_add_message( )` call to cover four vertical lines is a bit excessive in my opinion.

python/python/res/util/enums/message_level_enum.py
```diff
@@ -9,10 +9,39 @@ class MessageLevelEnum(BaseCEnum):
     LOG_WARNING  = None
     LOG_INFO     = None
     LOG_DEBUG    = None
+    LOG_NOTSET   = None
 
+    @staticmethod
+    def to_enum(val):
+        if val is None:
+            return MessageLevelEnum.LOG_WARNING
```
joakim-hove: Do we envisage actually using a numerical log level of e.g. 33?

python/python/res/util/res_log.py
```diff
@@ -1,19 +1,26 @@
...
-    _init = ResUtilPrototype("void res_log_init_log(int, char*, bool)", bind=False)
-    _write_log = ResUtilPrototype("void res_log_add_message_py(int, char*)", bind=False)
+    _init = ResUtilPrototype("void res_log_init_log(message_level_enum, char*, bool)", bind=False)
+    _write_log = ResUtilPrototype("void res_log_add_message_str(message_level_enum, char*)", bind=False)
     _get_filename = ResUtilPrototype("char* res_log_get_filename()", bind=False)
 
+    @staticmethod
+    def __enum(log_level):
+        if isinstance(log_level, int):
+            log_level = MessageLevelEnum.to_enum(log_level)
```
joakim-hove: What to do if the `isinstance(log_level, int)` test fails - exception?

python/python/res/util/res_log.py
```diff
@@ -1,19 +1,26 @@
...
     @classmethod
     def init(cls, log_level, log_filename, verbose):
-        cls._init(log_level, log_filename, verbose)
+        cls._init(cls.__enum(log_level), log_filename, verbose)
 
     @classmethod
     def log(cls, log_level, message):
         """ Higher log_level indicates higher importance"""
-        cls._write_log(log_level, message)
+        cls._write_log(cls.__enum(log_level), message)
```
joakim-hove: Wasn't it natural that the `log( )` method now expected an enum argument?

joakim-hove:
```txt
Where have all the braces gone
PR removed them everyone
where have all the braces gone 
....
```

If the braces must go, then please at least sure there is a blank line before the code block starte:
```C
if (early_exit_condition)
   return;

int var;
```

joakim-hove (commented):
It seems the Python api is now set up to use a mix of the new enum and integer log levels - or do I misunderstand? Apart from that this looks good.

libres_util/include/ert/res_util/log.h
pgdr: I can reindent the rest of the code if that is what we want.  Aligned code is a double-edged sword.

python/python/res/util/enums/message_level_enum.py
pgdr: Well, the classical loggers like log4j, log4c and Python's logging all accept integer levels for logging.  I don't know if we want that though.

python/python/res/util/res_log.py
pgdr: This was actually just under the assumption that the type is either int or a MessageLevelEnum.

python/python/res/util/res_log.py
pgdr: Yes, but now it supports both int and MessageLevelEnum.  I'm open for suggestions here!

libres_util/src/res_log.c
pgdr: > However splitting [...]  four vertical lines is a bit excessive in my opinion.

*If the arguments fit, I will acquit.*

However, I like to _either_ keep all arguments on one line, _or_ split all lines.  I am open for suggestions.

python/python/res/util/res_log.py
joakim-hove: I guess that is why it is slightly dangerous to start with `isinstance( )` checks in Python, it kind of breaks the duck-typing contract. Maybe the `MessageLevelEnum.to_enum( )` could be modified to accept a enum instance as input argument, then that method can be called unconditionally?

joakim-hove:
I had not seen #211 - does this fix that?

libres_util/src/res_log.c
joakim-hove: Well - this is not important; but I would prefer the current function call to be one *one single line*

> I like to either keep all arguments on one line, or split all lines.

My take on this is to try to group related arguments  - when that makes sense - on the same line. 

Anyway - no big deal either way.

joakim-hove:
Test failure smells of some log calls which fire before the logging has been properly initialzied?

joakim-hove:
I am happy with this - please merge when you feel ready.

libres_util/include/ert/res_util/log.h
markusdregi: I'm in favor of not reindenting, for the exact reason displayed here. Header files are too crucial to program flow that it is worth the cost of reindenting vs. the loss of git history in my opinion.

If you end up not reindenting, could you at least make the ~10 white spaces into a single one?

markusdregi (approved):
Reviewed in private communication. Merge when green.

pgdr:
All tests (Statoil incl.) now pass.

pgdr:
> I had not seen #211 - does this fix that?

Yes, @joakim-hove, this PR fixes #211.  The issue was with `LOG_CRITICAL = 0`.

libres_util/src/res_log.c
pgdr: fixed

PR #213, opened 2018-01-22
Batch revisted

markusdregi:
**Task**
_Change `BatchSimulator.start` to take named control variable values._

**Approach**
_Pass them as a dict instead of a list._

**Pre un-WIP checklist**

### DIFF
--- a/python/python/res/simulator/batch_simulator.py
+++ b/python/python/res/simulator/batch_simulator.py
@@ -20,15 +20,15 @@ class BatchSimulator(object):
         get when actually simulating. The @controls argument should be a
         dictionary like this:
 
-            controls = {"cmode": ["Well","Group"], "order" : ["W1", "W2", "W3"]}
+            controls = {"cmode": ["Well", "Group"], "order": ["W1", "W2", "W3"]}
 
         In this example the simulator will expect two arrays 'cmode' and
         'order', consisting of two and three elements respectively. When
         actually simualating these values will be written to json files looking
         like:
 
-             cmode.json = {"Well" : 1.0, "Group" : 2.0}
-             order.json = {"W1" : 1, "W2" : 1.0, "W3": 1.0}
+             cmode.json = {"Well": 1.0, "Group": 2.0}
+             order.json = {"W1": 1, "W2": 1.0, "W3": 1.0}
 
         When later invoking the start() method the simulator expects to get
         values for all parameters configured with the @controls argument,
@@ -70,7 +70,7 @@ class BatchSimulator(object):
         """Will start batch simulation, returning a handle to query status and results.
 
         The start method will submit simulations to the queue system and then
-        return a RobustContext handle which can be used to query for simulation
+        return a BatchContext handle which can be used to query for simulation
         status and results. The @case_name argument should just be string which
         will be used as name for the storage of these simulations in the
         system. The @controls argument is the set of control values, and the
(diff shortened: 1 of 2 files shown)


markusdregi:
@joakim-hove Please have a look to see that you agree regarding the API changes.

Please note that the formatting went kind of a wall. Will fix that in a commit no, so no need for formatting comments yet ;)

joakim-hove:
First of all: I assume this implies that seba is now ready to pass named parameters to the simulator?

python/python/res/simulator/batch_simulator.py
```diff
@@ -107,20 +120,25 @@ def start(self, case_name, controls):
...
-               node = EnkfNode(config_node)
-               ext_node = node.as_ext_param( )
-               ext_node.set_vector(values)
-               node.save(fs, node_id)
+            if set(control_dict.keys()) != set(self.controls_format.keys()):
+                err_msg = "Mismatch between initialized and provided control names."
+                raise KeyError(err_msg)
+
+            for control_name, control in control_dict.iteritems():
+                if set(control.keys()) != set(self.controls_format[control_name]):
```
joakim-hove: I am not too happy about this duplication of logic/configuration of keys, the `ext_param` implementation is the authorative source regarding these keys. The 
```Python
for var_name, value in control.iteritems():     
    ext_node[var_name] = value
```
will already raise a `KeyError` for invalid key. Could in addition have:
```Python
if len(ext_config) != len(control):
   raise ValueError("Not all keys set")
```
alternatively a convenience method could be added here: https://github.com/Statoil/libres/blob/master/python/python/res/enkf/config/ext_param_config.py

joakim-hove:
Well tested - apart from comment about duplicated validation logic this is ready to be merged (when Seba is!!)

markusdregi:
@joakim-hove: Yes, _Seba_ is somewhat ready for this. It's not perfect, but it is already extracting the necessary information from its tags. So we'll just use that for now and improve upon it with something more explicit in the future.

python/python/res/simulator/batch_simulator.py
markusdregi: I completely agree. I've fixed it now.

markusdregi:
Alright, I consider this one to be done. But I'll leave it unmerged for a little while to create relevant patches.

markusdregi:
The footprint of the last commit is relatively large and its solely a refactoring commit. In other words, the correctness of this PR is completely mute to whether we include the last commit or not. To summarize, it does the following:

- reformat code, in particular the dictionaries defining the batches. This is done as a join of my own personal taste and pylint recommendation, with the later taking precedence.
- Split one large test function into several smaller ones.
- Extract the logic that sets up the controls in `BatchSimulator.start` into a private method.

I leave it up to you to decide whether it should be included or not.

joakim-hove:
Details about formatting aside - is this ready for the green button? I guess there is some ordering requirement with other PR's?

markusdregi:
Yes, this one should go first, but will immediately break master _Seba_. The PR there is ready as well. In _Everest_ there should be no changes necessary. I'm running its tests with the two PR's as we speak to be sure.

I was thinking that we get both PR's reviewed and accepted first. Then, we merge them one by one as the tests are passing.

python/python/res/simulator/batch_simulator.py
```diff
@@ -50,27 +50,48 @@ def __init__(self, res_config, controls, results):
             raise ValueError("The first argument must be valid ResConfig instance")
 
         self.res_config = res_config
-        self.ert = EnKFMain( self.res_config )
-        self.control_keys = []
-        self.result_keys = []
+        self.ert = EnKFMain(self.res_config)
+        self.control_keys = tuple(controls.keys())
```
joakim-hove, line 54: Constification with `tuple` was a nice move

python/python/res/simulator/batch_simulator.py
```diff
@@ -50,27 +50,48 @@ def __init__(self, res_config, controls, results):
...
+        for sim_id, (geo_id, controls)  in enumerate(case):
+            assert isinstance(geo_id, int)
+
+            node_id = NodeId(0, sim_id)
+            if set(controls.keys()) != set(self.control_keys):
+                err_msg = "Mismatch between initialized and provided control names."
+                raise KeyError(err_msg)
+
+            for control_name, control in controls.iteritems():
+                ens_config = self.res_config.ensemble_config
```
joakim-hove, line 75: I would prefer if the `ens_config` variable  was moved out of both the loops.

joakim-hove (approved):

python/python/res/simulator/batch_simulator.py
markusdregi: It is just a shorter name for the `ensemble_config` in question. In my opinion it is odd to define it before needed. How about skipping the shorter name as it is only used once and go for
```python
node = EnkfNode(self.res_config.ensemble_config[control_name])
```
78 characters long and good to go?

joakim-hove:
No big deal. My point is just that if it is extracted outside, as I suggest, that will implicitly communicate that this variable is fully immutable.

PR #215, opened 2018-01-24
Close fh

pgdr:


andreabrambilla (approved):
:+1:

libenkf/src/runpath_list.c
```diff
@@ -257,6 +257,7 @@ bool runpath_list_load(runpath_list_type * list) {
       int iens, iter;
 
       int read_count = fscanf(stream, "%d %255s %255s %d", &iens, runpath, basename, &iter);
+      fclose(stream);
```
andreabrambilla: I think this should be after the while loop (well it would be nice that the file is closed in the same scope where it's created, just remove a couple of curly braces ;) )

pgdr: fixed

PR #222, opened 2018-01-25
Drop pythonpath manipulations from job_dispatch

joakim-hove:
**Task**
The PYTHONPATH manipulations in the `job_dispatch.py` script break komodo.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


pgdr (approved):
Good, that's the way it should be, yes.

Also, please remove all e-mail-related content at the same time.

PR #227, opened 2018-01-31
Remove const, unuse, uninit warnings

pgdr:
Some warnings have been purged, can be built with `-Werror=all` now

### DIFF
--- a/libconfig/src/config_content_node.c
+++ b/libconfig/src/config_content_node.c
@@ -92,9 +92,12 @@ static void config_content_node_push_string( config_content_node_type * node , c
   stringlist_append_owned_ref( node->string_storage , string );
 }
 
-const char * config_content_node_get_full_string( config_content_node_type * node , const char * sep ) {
+const char * config_content_node_get_full_string(const config_content_node_type * node,
+                                                 const char * sep) {
   char * full_string = stringlist_alloc_joined_string(node->stringlist , sep);
-  config_content_node_push_string( node , full_string );
+
+  // casting node to non-const, just to take ownership of full_string
+  config_content_node_push_string((config_content_node_type *)node, full_string );
   return full_string;
 }
 
--- a/libenkf/src/site_config.c
+++ b/libenkf/src/site_config.c
@@ -577,9 +577,9 @@ static bool site_config_init(site_config_type * site_config, const config_conten
 
   if (config_content_has_item(config, UMASK_KEY)) {
     const char * string_mask = config_content_get_value(config, UMASK_KEY);
-    mode_t umask_value;
+    int umask_value;
     if (util_sscanf_octal_int(string_mask, &umask_value))
-      site_config_set_umask(site_config, umask_value);
+      site_config_set_umask(site_config, (mode_t)umask_value);
     else
       util_abort("%s: failed to parse:\"%s\" as a valid octal literal \n", __func__, string_mask);
   }
--- a/libjob_queue/src/ext_job.c
+++ b/libjob_queue/src/ext_job.c
@@ -774,35 +774,35 @@ static void __fprintf_python_arg_types(FILE * stream,
                                        const char * suffix,
                                        const char * null_value) {
   fprintf(stream, "%s", prefix);
-  if (ext_job->arg_types) {
-    fprintf(stream, "\"%s\" : [", key);
-    int i;
-    for (i = 0; i < ext_job->max_arg; i++) {
-
-      char * arg_type;
-      int type = int_vector_safe_iget(ext_job->arg_types, i);
-      switch(type) {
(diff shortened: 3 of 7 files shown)


joakim-hove (approved):

libconfig/src/config_content_node.c
```diff
@@ -92,9 +92,12 @@ static void config_content_node_push_string( config_content_node_type * node , c
...
 }
 
-const char * config_content_node_get_full_string( config_content_node_type * node , const char * sep ) {
+const char * config_content_node_get_full_string(const config_content_node_type * node,
+                                                 const char * sep) {
   char * full_string = stringlist_alloc_joined_string(node->stringlist , sep);
-  config_content_node_push_string( node , full_string );
+
+  // casting node to non-const, just to take ownership of full_string
+  config_content_node_push_string((config_content_node_type *)node, full_string );
```
markusdregi, line 100: This is a really bad idea!

libenkf/src/site_config.c
```diff
@@ -577,9 +577,9 @@ static bool site_config_init(site_config_type * site_config, const config_conten
 
   if (config_content_has_item(config, UMASK_KEY)) {
     const char * string_mask = config_content_get_value(config, UMASK_KEY);
-    mode_t umask_value;
+    int umask_value;
     if (util_sscanf_octal_int(string_mask, &umask_value))
-      site_config_set_umask(site_config, umask_value);
+      site_config_set_umask(site_config, (mode_t)umask_value);
```
markusdregi, line 582: I would have preferred `util_sscanf_octal_int` to be refactored into `util_sscanf_octal_mode_t`. This is the only place this function is used within `libecl` and `libres`..

libjob_queue/src/ext_job.c
```diff
@@ -774,35 +774,35 @@ static void __fprintf_python_arg_types(FILE * stream,
...
-        case CONFIG_STRING:       arg_type = JOB_STRING_TYPE; break;
-        case CONFIG_BOOL:         arg_type = JOB_BOOL_TYPE; break;
-        case CONFIG_RUNTIME_FILE: arg_type = JOB_RUNTIME_FILE_TYPE; break;
-        case CONFIG_RUNTIME_INT:  arg_type = JOB_RUNTIME_INT_TYPE; break;
-        
-        default: util_abort("ERROR in %s", __func__);
-      }
+  if (!ext_job->arg_types) {
+    fprintf(stream , "\"%s\" : %s" , key, null_value);
+    goto postfix;
```
markusdregi, line 779: A `goto` statement?

markusdregi:
Oh, I see now that this was merged.

In general I'm skeptical to removing compiler warnings by casting, there is a reason for the warning being produced. Modifying a `const` pointer by casting it is just not good and the warning displayed this problem. Now it will remain forever mute and most likely never fixed properly..

libenkf/src/site_config.c
pgdr: I agree that this should have been taken out of `ecl` and made into a `..._mode_t`.

pgdr:
> modifying a const pointer by casting it is just not good and the warning displayed this problem

Markus, as long as the underlying object is non-const and you only deal with a "const pointer", it is safe to cast away the `const` modifier.  Since we're not in C++, we don't have the option of "safely" casting with `const_cast`.

markusdregi:
Safe yes, but there is in my opinion still something off with the path laid out for this pointer when one has to do such a cast. It is my opinion that the const should either be propagated down in the call path or removed also from the calling scope.

pgdr:
But the function `get_full_string` encapsulates this behavior completely.  For all practical purposes, the underlying struct does not change, except that it now has an extra `char*` it will free when it's free'd.

markusdregi:
First of all, are you sure that you can guarantee that in all future use cases this pointer is not actually pointing at something const?

Second, and to adress the above comment; if you need to cast away a const, then you are either locally doing something funky you should not do, the function should not have taken a const pointer to begin with or you are passing the pointer to a function that want a non-const pointer. In the last case, If you are certain that it is ok to call that function, that function again should have taken a const pointer. Apply argument inductively upwards or downwards in the call stack.

Sometimes you leave the libraries you have control over and need to make the best out of it. This was not the case this time (although we got into libecl, I think we should have propagated the change down if that was the correct thing to do).

pgdr:
Do you agree that there are cases where `const_cast` is necessary or even a good idea?

markusdregi:
In a world where code was written in a day and legacy was an unknown phenomenon; nope. But that is not always the case and due to this I accept the occasional necessity. But, then you are already in a scenario you preferably should have avoided..

I have a feeling this discussion is to be continued over a coffee or beer at some point ;)

PR #229, opened 2018-01-31
Added friendly message for change in interpretation of obs files.

joakim-hove:
**Task**
As part of the development cycle from 2.2 to 2.3 we have changed how paths are resolved in the observation file. They are now resolved relative to the main observation file, which is accordance with how the rest of ERT handles paths.

When the new version is deployed people must update their observation files. With this PR they should get a reasonably nice message informing them what to do.

**Pre un-WIP checklist**


joakim-hove (commented):
Very good Joakim :-)

PR #244, opened 2018-02-20
Do not crash on failed simulation — return dict from results

pgdr:
Before this PR, the BatchContext simulator would crash on `sim.results()` if some of the realizations failed.

Instead, log an `error` and return `None` for the specific realization.

The API `results` now returns a list of results or `None` if an error occurred, and the list order corresponds to the input order of `start`.

```python
[
  {"CMODE" : [1,2,3], "order" : [1,1,3]},
  None,
  {"CMODE" : [6,1,0], "order" : [0,0,8] }
]
```

### DIFF
--- a/python/python/res/simulator/batch_simulator_context.py
+++ b/python/python/res/simulator/batch_simulator_context.py
@@ -2,6 +2,8 @@ import time
 from collections import namedtuple
 from res.server import SimulationContext
 from res.enkf import NodeId, EnkfNode
+import logging
+import numpy as np
 
 Status = namedtuple("Status", "waiting pending running complete failed")
 
@@ -58,11 +60,15 @@ class BatchContext(SimulationContext):
         when the simulator was created the results will be returned as:
 
 
-          [  {"CMODE" : [1,2,3], "order" : [1,1,3]},
-             {"CMODE" : [1,4,1], "order" : [0,7,8]},
-             {"CMODE" : [6,1,0], "order" : [0,0,8] ]
+          [ {"CMODE" : [1,2,3], "order" : [1,1,3]},
+            {"CMODE" : [1,4,1], "order" : [0,7,8]},
+            None,
+            {"CMODE" : [6,1,0], "order" : [0,0,8]} ]
 
-        For a simulation which consist of a total of three "sub-simulations".
+        For a simulation which consist of a total of four simulations, where the
+        None value indicates that the simulator was unable to compute a request.
+        The order of the list corresponds to case_data provided in the start
+        call.
 
         """
         if self.running():
@@ -72,11 +78,15 @@ class BatchContext(SimulationContext):
         nodes = [ EnkfNode(self.res_config.ensemble_config[key]) for key in self.result_keys ]
         for sim_id in range(len(self)):
             node_id = NodeId( 0, sim_id)
+            if not self._queue_manager.didJobSucceed(sim_id):
+                logging.error('Simulation %d (node %s) failed.' % (sim_id, str(node_id)))
+                res.append(None)
+                continue
             d = {}
             for node in nodes:
                 node.load(self.get_sim_fs(), node_id)
-                gen_data = node.asGenData( )
-                d[node.name()] = gen_data.getData( )
+                data = node.asGenData().getData()
+                d[node.name()] = np.array(data)
             res.append(d)
 
         return res


pgdr:
I catch `Exception` here since that is what `enkf_node.load` raises.  I'd be happy to change `enkf_node.load` to raise something else instead if someone has a good idea what to raise instead.

joakim-hove:
The `SimulationContext._queue_manager`: https://github.com/Statoil/libres/blob/master/python/python/res/server/simulation_context.py#L19 is the main source to check simulation status.

This class has a method [getJobStatus()](https://github.com/Statoil/libres/blob/master/python/python/res/job_queue/job_queue_manager.py#L102) - could add  a forwarding from this to the `SimulationContext`.

joakim-hove:
I also thought we could add truth function to the context, which would evaulate to `True` if all simulations succeeded and `False`otherwise:
```Python
if context:
     # All simulations good
     res = context.get_results()
else:
    # Some realisations failed
    # Check status more carefully and ...?
```
I don't know?

pgdr:
But don't you agree that it would be better to have the following?
```python
    SimulationContext.success():
        return all_realizations_succeeded # bool

    SimulationContext.results():
        return all_results # list possibly containing None
```

Think as a user of the simulation context (as in Seba).

pgdr:
> Some realisations **failed**
> Check status more carefully and ...?

That sounds good, but what should the user of the API check?  Recall that the user doesn't know (nor care) whether we are using Eclipse or an email client for simulation.

joakim-hove:
Currently the `context.get_results()`returns a list of dictonaries. We could return a dictionary instead - like:
```Python
results = {"0" : {"NPV" : [0,1,1], "c1" : [0,1], "c2" : [9,7]},
           "3" : {"NPV" : [0,1,2], "c2" : [9,4], "c2" : [9,7]}}
```  
for batch which has only two succesfull simulations?

pgdr:
As long as Seba can extract which realization each value belongs to, it should be fine to return a dictionary, I guess.

joakim-hove:
> That sounds good, but what should the user of the API check? Recall that the user doesn't know (nor care) whether we are using Eclipse or an email client for simulation.

The API user I thought of here was Seba; which must be able to say whether it can be able to live with these failed realisations or not. Failure will be "atomic" for the whole forward model.

pgdr:
> Failure will be "atomic" for the whole forward model.

Yes, completely agree.

joakim-hove:
> But don't you agree that it would be better to have the following?

I certainly agree on the `suceess()` method; a bit more sceptical to returning `None` - I fear that will go undected for some iterations in Seba, and then just blow up in mysterious ways later. The argument in favor of `dict` was that then seba would not iterate more determinedly over the simulations, and would then probably be more prepared to handle missing data than a flat list with `None` in it?

agchitu:
In my view the caller of `results()`  Seba in this case should get a dictionary where there are results. It would be great if the failed simulations will be also part of that dictionary with a value saying "failed", maybe also the reason.  If this is not possible it is still fine.

pgdr:
Is the remaining question now which of the following three we return?
```python
# Alternative 1 (this PR)
[
  {"CMODE" : [1,2,3], "order" : [1,1,3]},
  None,
  {"CMODE" : [6,1,0], "order" : [0,0,8]}
]

# Alternative 2 (explicit)
{
  0: {"CMODE" : [1,2,3], "order" : [1,1,3]},
  1: None,
  2: {"CMODE" : [6,1,0], "order" : [0,0,8] }
 }

# Alternative 3 (implicit)
{
  0: {"CMODE" : [1,2,3], "order" : [1,1,3]},
  2: {"CMODE" : [6,1,0], "order" : [0,0,8] }
 }
```

agchitu:
> As long as Seba can extract which realization each value belongs to, it should be fine to return a dictionary, I guess.

Seba needs to know the simulation_id not the realization name (geo_id)

joakim-hove:
Iff we are to return `None` for failed realisastions they should at least befor the whole realisation, i.e.
```Python
results = [{"NPV" : [0,1,1], "c1" : [0,1], "c2" : [9,7]},
           None,
           None,
           {"NPV" : [0,1,2], "c2" : [9,4], "c2" : [9,7]}]
```
instead of:
```Python
results = [{"NPV" : [0,1,1], "c1" : [0,1], "c2" : [9,7]},
           {"NPV" : None, "c1" : None, "c2" : None,
           {"NPV" : None, "c1" : None, "c2" : None,
           {"NPV" : [0,1,2], "c2" : [9,4], "c2" : [9,7]}]
```

joakim-hove:
> Seba needs to know the simulation_id not the realization name (geo_id)

Yes - we know.

pgdr:
> I fear that will go undected for some iterations in Seba, and then just blow up in mysterious ways later

Let's have some faith in Seba.

> The argument in favor of dict was that then seba would not iterate more determinedly over the simulations, and would then probably be more prepared to handle missing data than a flat list with None in it?

I can agree that this makes failures less probable.

pgdr:
> Iff we are to return None for failed realisastions they should at least befor the whole realisation, i.e.

I agree.

agchitu:
```Python
results = [{"NPV" : [0,1,1], "c1" : [0,1], "c2" : [9,7]},
           {"failed" : True},
           {"failed" : True},
           {"NPV" : [0,1,2], "c2" : [9,4], "c2" : [9,7]}]
```
Is this possible?

joakim-hove:
Ehhh - I think this PR is:

```Python
# Alternative 1 (this PR)
[
  {"CMODE" : [1,2,3], "order" : [1,1,3]},
  {"CMODE" : None, "order" : None},
  {"CMODE" : [6,1,0], "order" : [0,0,8]}
]
```

joakim-hove:
> Is this possible?

Everything is possible; but I did not like it ....

agchitu:
"CMODE" : None, "order" : None, would that not be the same? Also if I have lost of results, would that not be cumbersome too have them all in a dictionary but set to None?

pgdr:
> Ehhh - I think this PR is...

yes, my bad ... this PR is bad.  I'll fix it to whatever we agree to.


I agree with Joakim that `'failure': True` is a bit un-pretty :)

pgdr:
Yes, Alin, we agree that it should be only `1: None` if we will have an entry for `1`.

agchitu:
Fine, just let me know would Seba need to expect

python/python/res/simulator/batch_simulator_context.py
```diff
@@ -58,25 +59,32 @@ def results(self):
...
 
         For a simulation which consist of a total of three "sub-simulations".
 
         """
         if self.running():
             raise RuntimeError("Simulations are still running - need to wait before gettting results")
 
-        res = []
+        res = {}
         nodes = [ EnkfNode(self.res_config.ensemble_config[key]) for key in self.result_keys ]
```
joakim-hove: This will probably work according to specification; I would still strongly prefer if we checked simulation status with the api, at the very least this will silently return stale data if you are rerunning a simulation, which worked the last time,

joakim-hove (dismissed):
Unless there is a good reason to the contrary I want this to use the api to check for simulation status; otherwise I am happy to merge this.

pgdr:
You mean instead of try/catch, I check the getJobStatus for the job_index (which is sim_id?) ?

joakim-hove:
> You mean instead of try/catch, I check the getJobStatus for the job_index (which is sim_id?) ?

Yes!

pgdr:
Will fix!

python/python/res/simulator/batch_simulator_context.py
```diff
@@ -58,25 +59,31 @@ def results(self):
         when the simulator was created the results will be returned as:
 
 
-          [  {"CMODE" : [1,2,3], "order" : [1,1,3]},
-             {"CMODE" : [1,4,1], "order" : [0,7,8]},
-             {"CMODE" : [6,1,0], "order" : [0,0,8] ]
+          { 0: {"CMODE" : [1,2,3], "order" : [1,1,3]},
```
markusdregi: It is not transparent what the keys of the result correspond to. They are indeed the indices in the list of control values passed to the BatchSimulator, but this needs to be documented! In addition, I find the inconsistency in simulation ids a bit odd, could we either make them implicit or explicit in both ends; that is both when passed to start and when returning the results?

Personally I would prefer to have them explicit on both sides, where each simulation passed to start could be a dict specifying geo_realization, simulation_id and input/controls (we should perhaps get rid the control notion in libres as a BatchSimulator is definitively useful outside the scope of OR/optimization.

markusdregi:
The layout of the results are documented also in the BatchSimulator. They need to be updated..

pgdr:
@joakim-hove today, the Simulator doesn't return `[1,2,3]` but a `DoubleVector`.  Could we possible either return a `list` or `np.array` instead?

joakim-hove:
> @joakim-hove today, the Simulator doesn't return [1,2,3] but a DoubleVector. Could we possible either return a list or np.array instead?

Hehe - I thought the zen of Python was to not care about explicit types? Seriously - we can of course change that - any preferences of python List or `np.array`?

pgdr:
If it is a numpy array, we can immediately run `min`, `mean`, `std`, etc.

I agree that we should be happy with duck typing, but the `__repr__` and perhaps also `__str__` of the `DoubleVector` makes working with them a bit more annoying than with lists or np.arrays.

Lists and np.arrays are also more suitable for pickling, if that becomes necessary.

I have no strong opinions, though, it's just simpler to work with those.

markusdregi (approved):
Looks good to me! :)

pgdr:
Jenkins build 251 succeeded.
```
Label Time Summary:
Python                = 345.43 sec
Python:SLOW_1         = 173.22 sec
Python:SLOW_2         = 228.51 sec
Python:StatoilData    =  70.53 sec
SLOW_1                = 143.77 sec
StatoilData           =  53.79 sec

Total Test time (real) = 300.15 sec
/var/jenkins/jenkins-be/workspace/libres-PR
Tested PR: 244 succesfully

Finished: SUCCESS
```

PR #246, opened 2018-02-23
Fix circular dependency on KeyManager

andreabrambilla:
**Task**
Fix reciprocal dependency between KeyManager and EnKFMain

**Approach**
KeyManager uses weakref. At least in the tests, it seems the KeyManager never accesses a dead instance of EnKF main. If that will ever happen, an exception is raised.

Also, access to the EnKFMain instance on KeyManager has been made private (again, tests seem happy with that)

**Pre un-WIP checklist**


pgdr (approved):
This looks good!

PR #247, opened 2018-02-24
Remove circular dependencies from EnKFMain

andreabrambilla:
`EnKFMain` owned instances of `EnkfFSManager`, `ESUpdate` and `EnkfSimulationRunnner`. And each of these instances referred back to `EnKFMain` as `BaseCClass.__parent`, ie:
```
    -----------------------------------
   |   real EnKFMain object in memory  |
    -----------------------------------
       ^                        ^
       |                        |
    (c_ptr)                  (c_ptr)
       |                        |
       |                        |
    EnKFMain ----(owns)------v  |
          ^                  v  |
          ^----(parent)-----EnkfFSManager
```

Now EnKFMain has been split in two classes, and the dependency structure is as follows:
```
    ------------------------------------
   |   real EnKFMain object in memory   |
    ------------------------------------
       ^                ^           ^
       |                |           |
    (c_ptr)          (c_ptr)        |
       |                |           |
  _RealEnKFMain         |           |
    ^  ^                |           |
    |  ^--(parent)-- EnkfFSManager  |
    |                   ^           |
  (parent)              |           |
    |                   |           |
  EnKFMain ----(owns)---            |
         ------------------(c_ptr)--
 ```
The code has been also modified so that the wrapped types `enkf_main`, `enkf_main_ref`, `enkf_main_obj` are mapped to the python class `EnKFMain` instead of `_RealEnKFMain`

**Pre un-WIP checklist**

Statoil/ert#99

### DIFF
--- a/python/python/res/enkf/enkf_fs_manager.py
+++ b/python/python/res/enkf/enkf_fs_manager.py
@@ -1,9 +1,10 @@
 import os.path
-from cwrap import BaseCClass
-from res.enkf import EnkfFs, StateMap, TimeMap, RealizationStateEnum, EnkfInitModeEnum, EnkfPrototype
-from ecl.util.util import StringList, BoolVector
-
 import re
+
+from cwrap import BaseCClass
+from ecl.util.util import StringList, BoolVector
+from res.enkf import EnkfFs, StateMap, TimeMap, RealizationStateEnum, EnkfInitModeEnum, EnkfPrototype
+
 
 def naturalSortKey(s, _nsre=re.compile('([0-9]+)')):
     return [int(text) if text.isdigit() else text.lower() for text in re.split(_nsre, s)]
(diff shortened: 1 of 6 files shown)


andreabrambilla:
The failure is (probably) because Travis runs also the tests for ERT, which is not yet patched. Working on it ... (https://github.com/Statoil/ert/pull/99)

joakim-hove:
> EnKFMain owned instances of EnkfFSManager, ESUpdate and EnkfSimulationRunnner. And each of these instances referred back to EnKFMain as BaseCClass.__parent, ie:

An ugly pattern from the dark ages - I am glad you are kicking it :-)

joakim-hove:
Question: If the EnkfMain object inherited from RealEnkfMain - could we then avoid the monkey-patching?

joakim-hove:
General comment: We will merge something like this quite fast, to solve the problems in Everest; but I must admit I feel the changes are invasive and I am slightly uncomfortable ... I guess we can rework it in the not-so-distant future.

Please include the beautiful ASCII art from the PR description in the  enkf_main.py file.

andreabrambilla:
Inheritance would reintroduce the circular dependency, so that's not an option, we need to have 2 distinct objects

These changes make me uncomfortable as well, but it is a necessary evil due to the mismatching implementation of EnKFMain between C (only 1 class) and Python (4 interconnected classes). A more elegant solution would be to modify either the C or the Python implementation and restore the 1-to-1 correspondence

I'll include the ASCII drawing ;)

pgdr:
> I feel the changes are invasive and I am slightly uncomfortable

I agree with Joakim that this is quite a scary change.  If we merge this, I guess it will be a part of 2.3,  and I therefore think we should tag a new tag immediately and start testing in it `unstable` (including long running LSF jobs with many realizations).

joakim-hove:
>Inheritance would reintroduce the circular dependency, so that's not an option, we need to have 2 distinct objects

You are the one who has thought this through, but still .... I agree we must havve tw distinct object instances, but I thought maybe the `EnkFMain` class could be derived from the `RealEnkfMain` - then methods could be in `RealEnkfMain` without need for monkey patching?

> A more elegant solution would be to modify either the C or the Python implementation and restore the 1-to-1 correspondence

I agree - but that is a non-trivial task. We will discuss this further tomorrow.

andreabrambilla:
Yes, we should spend some time testing this. Note that there is no change in the API, so this could be also part of a 2.2.x version if needed.

andreabrambilla:
@joakim-hove you are probably right about inheritance, it could work (I did not really think it through). 

However, this PR is a bit funky, so it will most likely need to be reworked. But it solves the issue, and leaves the API unchanged. We discussed it a bit here, and it's probably worth to merge it now, and think about how to improve it later on. Is that fine for you? We can have a chat on lync/appear.in if it can help

joakim-hove:
> We discussed it a bit here, and it's probably worth to merge it now, and think about how to improve it later on. Is that fine for you?

OK - that is fine by me; we merge it now; I will immediately create a `2.3.a3` tag so we get this in test-use.

joakim-hove (approved):

PR #250, opened 2018-02-27
Progress api

joakim-hove:
**Task**
With this PR you can ask the `SImulationContext` for the status/progress of the jobs in the forward model.

The actual api - as it will be used from Everest is in the final commit; feel free to suggest changes.

**Approach**
The `JobManager` class which is running the forward model will dump a file `status.json` - and the main application will load that when queried for status.

**Pre un-WIP checklist**

### DIFF
--- a/libjob_queue/include/ert/job_queue/ext_job.h
+++ b/libjob_queue/include/ert/job_queue/ext_job.h
@@ -1,19 +1,19 @@
 /*
-   Copyright (C) 2011  Statoil ASA, Norway. 
-    
-   The file 'ext_job.h' is part of ERT - Ensemble based Reservoir Tool. 
-    
-   ERT is free software: you can redistribute it and/or modify 
-   it under the terms of the GNU General Public License as published by 
-   the Free Software Foundation, either version 3 of the License, or 
-   (at your option) any later version. 
-    
-   ERT is distributed in the hope that it will be useful, but WITHOUT ANY 
-   WARRANTY; without even the implied warranty of MERCHANTABILITY or 
-   FITNESS FOR A PARTICULAR PURPOSE.   
-    
-   See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html> 
-   for more details. 
+   Copyright (C) 2011  Statoil ASA, Norway.
+
+   The file 'ext_job.h' is part of ERT - Ensemble based Reservoir Tool.
+
+   ERT is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   ERT is distributed in the hope that it will be useful, but WITHOUT ANY
+   WARRANTY; without even the implied warranty of MERCHANTABILITY or
+   FITNESS FOR A PARTICULAR PURPOSE.
+
+   See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
+   for more details.
 */
 
 #ifndef ERT_EXT_JOB_H
--- a/libjob_queue/include/ert/job_queue/ext_job.h
+++ b/libjob_queue/include/ert/job_queue/ext_job.h
@@ -89,7 +89,7 @@ void                    ext_job_add_environment(ext_job_type *ext_job , const ch
 void                    ext_job_clear_environment( ext_job_type * ext_job );
 hash_type             * ext_job_get_environment( ext_job_type * ext_job );
 int                     ext_job_set_private_args_from_string( ext_job_type * ext_job , const char * arg_string );
-const char            * ext_job_get_private_args_as_string( ext_job_type * ext_job ); 
+const char            * ext_job_get_private_args_as_string( ext_job_type * ext_job );
 const char *            ext_job_get_license_path(const ext_job_type*);
 //const char            * ext_job_get_arglist_as_string( ext_job_type * ext_job );
 //void                    ext_job_set_arglist_from_string( ext_job_type * ext_job , const char * argv_string );
--- a/python/python/res/job_queue/forward_model_status.py
+++ b/python/python/res/job_queue/forward_model_status.py
@@ -1,0 +1,141 @@
+#  Copyright (C) 2018  Statoil ASA, Norway.
+#
+#  The file 'forward_model_status.py' is part of ERT - Ensemble based Reservoir Tool.
+#
+#  ERT is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  ERT is distributed in the hope that it will be useful, but WITHOUT ANY
+#  WARRANTY; without even the implied warranty of MERCHANTABILITY or
+#  FITNESS FOR A PARTICULAR PURPOSE.
+#
+#  See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
+#  for more details.
+import os.path
+import json
+import datetime
+import time
+import sys
+
+def _serialize_date(dt):
+    if dt is None:
+        return None
+
+    return time.mktime(dt.timetuple())
+
+
+def _deserialize_date(serial_dt):
+    if serial_dt is None:
+        return None
+
+    time_struct = time.gmtime(serial_dt)
+    return datetime.datetime(*time_struct[0:6])
+
+
+class ForwardModelJobStatus(object):
+
+    def __init__(self, name, start_time = None, end_time = None, status = "Waiting", error=None):
+        self.start_time = start_time
+        self.end_time = end_time
+        self.name = name
+        self.status = status
+        self.error = error
+
+
+    @classmethod
+    def load(cls,data):
+        start_time = _deserialize_date(data["start_time"])
+        end_time = _deserialize_date(data["end_time"])
+        name = data["name"]
+        status = data["status"]
+        error = data["error"]
+
+        return cls(name,
+                   start_time=start_time,
+                   end_time=end_time,
+                   status=status,
+                   error=error)
+
+
+    def __str__(self):
+        return "name:{} start_time:{}  end_time:{}  status:{}  error:{} ".format(self.name, self.start_time, self.end_time, self.status, self.error)
+
+
+    def dump_data(self):
+        return {"name" : self.name,
+                "status" : self.status,
+                "error" : self.error,
+                "start_time" : _serialize_date(self.start_time),
+                "end_time" : _serialize_date(self.end_time)}
+
+class ForwardModelStatus(object):
+    STATUS_FILE = "status.json"
+
+
+    def __init__(self, run_id, start_time = None, end_time = None):
+        self.run_id = run_id
+        self.start_time = start_time
+        self.end_time = end_time
+        self._jobs = []
+
+    @classmethod
+    def try_load(cls, status_file):
+        fp = open(status_file)
+        data = json.load(fp)
+
+        start_time = _deserialize_date(data["start_time"])
+        end_time = _deserialize_date(data["end_time"])
+        status = cls(data["run_id"],
+                     start_time=start_time,
+                     end_time=end_time)
+
+        for job in data["jobs"]:
+            status.add_job(ForwardModelJobStatus.load(job))
+
+        return status
+
+
+    @classmethod
+    def load(cls, path):
+        num_retry = 10
+        sleep_time = 0.10
+        attempt = 0
+        status_file = os.path.join(path, cls.STATUS_FILE)
+        while attempt < num_retry:
+            try:
+                status = cls.try_load(status_file)
+                return status
+            except:
+                time.sleep(0.10)
+                attempt += 1
+
+        return None
+
+
+    @property
+    def jobs(self):
+        return self._jobs
+
+
+    def add_job(self, job):
+        self._jobs.append(job)
+
+
+    def dump(self, filename = None):
+        if filename is None:
+            status_file = self.STATUS_FILE
+        else:
+            status_file = filename
+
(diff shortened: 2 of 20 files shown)


python/python/res/job_queue/forward_model_status.py
```diff
@@ -0,0 +1,143 @@
...
+        return time.mktime(dt.timetuple())
+
+
+    @staticmethod
+    def deserialize_date(serial_dt):
+        if serial_dt is None:
+            return None
+
+        time_struct = time.gmtime(serial_dt)
+        return datetime.datetime(*time_struct[0:6])
```
andreabrambilla: `serialize_date` and `deserialize_date` doesn't seem to have much to do with this class. What about moving them outside this class?

python/python/res/job_queue/forward_model_status.py
```diff
@@ -0,0 +1,143 @@
...
+
+        return None
+
+
+    def __len__(self):
+        return len(self.jobs)
+
+
+    def __getitem__(self, index):
+        return self.jobs[index]
```
andreabrambilla: This is probably a matter of personal preference. I am not a big fun of using arbitrary classes as arrays, unless their main responsibility is to behave like arrays. And it doesn't seem to be the case here. Since we have a `add_job` method, I'd consider replacing `__len__` and `__getitem__` with one of these two options
```Python
@property
def jobs(self):
  return self._jobs
```
or
```python
def jobs_num(self): 
  return len(self._jobs)
def job(self, i):
  return self._jobs[i]
```

python/python/res/job_queue/job_manager.py
```diff
@@ -221,6 +227,12 @@ def _loadJson(self, json_file_name):
         self._ensureCompatibleJobList()
         self._buildJobMap()
 
+        for job in self.job_list:
+            self.job_status.add_job( ForwardModelJobStatus(job.get("name")))
+
+        # "Monkey-patching" the job object by attaching a status object.
+        for i in range(len(self.job_list)):
+            self.job_list[i]["status"] = self.job_status[i]
```
andreabrambilla: I feel this monkey patching is a bit unnecessary. The `job_list` here seems relatively static, and it seems that every entry in `self.job_list` is in a one-to-one correspondence with entries of `self.job_status`. So, the status of the `i`-th job in `self.job_list` can be always retrieved with `self.job_status[i]`

python/python/res/job_queue/job_manager.py
```diff
@@ -470,21 +482,48 @@ def addLogLine(self, job):
     def runJob(self, job):
         assert_file_executable(job.get('executable'))
         self.addLogLine(job)
+
+        status = job["status"]
+        status.start_time = dt.now()
+        status.status = "Running"
+        self.job_status.dump()
```
andreabrambilla, line 490: This feels a bit weird to me, but maybe it's because I don't know these classes sufficiently well. 
If I understood correctly, `runJob` receives a job and executes it. `runJob` belongs to (an instance of) `JobManager`, which has a `self.job_list` member.
As far as I can see here, there is no explicit relation between the input `job` and the member `self.job_list`.
So, when we write `self.job_status.dump()` we are kind of relying on the fact that somehow, the `job` we have in input has been added to the `self.job_status` object.
I'd prefer something like this: 
```Python 
def runJob(self, job):
         ...
         i = self.job_list.index(job)
         status = self.job_status[i]
         ... #do stuff with the status
         self.job_status.dump() # we know for sure that the status we modified is part of job_status
```

python/python/res/job_queue/forward_model_status.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+    def add_job(self, job):
+        self.jobs.append(job)
```
andreabrambilla: I think this class should make sure that what is added is indeed an instance of `ForwardModelJobStatus`.
One option could be to have an explicit assertion on the input type. 
Alternatively: 
```Python
def add_job(self, name, start_time = None, end_time = None, status = "Waiting", error=None):
    self.jobs.append( ForwardModelJobStatus(name, start_time, end_time, status, error))
```

python/python/res/server/simulation_context.py
```diff
@@ -121,3 +122,46 @@ def get_run_context(self):
...
+        if not iens in self._run_args:
+            return (None,None)
+
+        run_arg = self._run_args[iens]
+        details = ForwardModelStatus.load(run_arg.runpath, num_retry = num_retry)
+
+        queue_index = run_arg.getQueueIndex()
+        global_status = self._queue_manager.getJobStatus(queue_index)
+
+        return (global_status, details)
```
andreabrambilla: These two things doesn't seem to depend on each other, maybe we could write 2 separate functions, eg `queue_status` and `jobs_status`

andreabrambilla:
I had a few comments on the implementation details, overall the PR looks fine. However, my knowledge of libres is still somehow limited, I think it's better if someone else reviews this as well

python/python/res/server/simulation_context.py
```diff
@@ -122,3 +122,46 @@ def get_run_context(self):
...
+           jobs: [ (job1.name, job1.start_time, job1.end_time, job1.status, job1.error_msg),
+                   (job2.name, job2.start_time, job2.end_time, job2.status, job2.error_msg),
+                   ....
+                   (jobN.name, jobN.start_time, jobN.end_time, jobN.status, jobN.error_msg) ]
+
+           The element in the jobs list can be access with details[].
+
+        """
+        if not iens in self._run_args:
+            return (None,None)
```
pgdr: Shouldn't this raise a `KeyError` or perhaps `ValueError`?

test-data/local/batch_sim/jobs/square_params.py
```diff
@@ -12,6 +14,7 @@ def copy_file(input_file, output_file):
             sq = data[key] * data[key]
             f.write("%g\n" % sq)
 
+    time.sleep(random.randint(0,4))
```
pgdr, line 17: >  Updated test job with some randomness. 

Perhaps you can add a comment as to why we sleep a random int?  Why not just `sleep(1)` if that's necessary?

python/python/res/job_queue/forward_model_status.py
```diff
@@ -0,0 +1,143 @@
...
+                     end_time=end_time)
+
+        for job in data["jobs"]:
+            status.add_job(ForwardModelJobStatus.load(job))
+
+        return status
+
+
+    @classmethod
+    def load(cls, path, num_retry=10):
```
pgdr: Do you think we need `num_retry` to be configurable immediately?  Perhaps we can add that later if necessary?

libjob_queue/include/ert/job_queue/ext_job.h
```diff
@@ -34,7 +34,7 @@ typedef struct ext_job_struct ext_job_type;
 
 const char *            ext_job_get_help_text( const ext_job_type * job );
 void                    ext_job_set_help_text( ext_job_type * job , const char * help_text);
-
+  void                    ext_job_fprintf_initial_status(const ext_job_type * job, FILE * stream);
```
pgdr: inconsistent indentation

pgdr:
1. Isn't the relevant information already existent in the files in the fs today?  That is, adding the json file is _strictly_ not necessary?  Because when I suggested previously that we could get this information by reading the files in the fs, I seem to recall that this information was already internalized by the queue system.  My memory might be off here, though
2. I think this task should be solvable using Python only, is there a reason why we implement json related stuff in the C library instead of doing everything via Python?
3. Perhaps we should depend on a JSON library in C if we want to do more reading and writing of JSON instead of implementing an maintaining it ourselves?

joakim-hove:
Thank you for the extensive review.

joakim-hove:
> Isn't the relevant information already existent in the files in the fs today? That is, adding the json file is strictly not necessary? 

That is correct - the information read from the new `status.json` file is already present in the text file `STATUS` - but the latter is meant for human consumption; and I would *strongly* prefer to base the api on more controlled file.

> I seem to recall that this information was already internalized by the queue system

That is not correct - the queue system only knows the global status; i.e. waiting, running, failed and so on of the jobs, not any details of the elements in the job.

joakim-hove:
> 2. I think this task should be solvable using Python only, is there a reason why we implement json related stuff in the C library instead of doing everything via Python?

The C code is used to dump an *initial* version of the `status.json`; that again is needed for the api to be able to report correctly for jobs which have not yet started. The moment a job is picked up by the `job_dispatch` script and starts running it is a Python only solution. But I agree - we could just say that before a job has started you can not get any details - then all the C code changes can be reverted.

> 3. Perhaps we should depend on a JSON library in C if we want to do more reading and writing of JSON instead of implementing an maintaining it ourselves?

I think the best solution is to just say that status queries are not possible before the job is running; then the whole point collapses.

python/python/res/job_queue/job_manager.py
joakim-hove: > This feels a bit weird to me

It is weird - this code is more than ten years old, and in dire need for refactoring and cleanup. The last year we have made changes which should make such improvements easier; it will come - but not in this PR.

python/python/res/job_queue/forward_model_status.py
joakim-hove: I'll implement the property.

python/python/res/job_queue/forward_model_status.py
joakim-hove: Ehhh - here I disagree; at least this serialization is by no means generic - and in my opinion it makes sense to namespace it under the class, to signal that this is a one off serialize / deserialize pair; could attach it to the `ForwardModelJobStatus` class instead though.

python/python/res/job_queue/job_manager.py
joakim-hove: OK - that might be; I'll look into it.

python/python/res/job_queue/forward_model_status.py
joakim-hove: > I think this class should make sure that what is added is indeed an instance of ForwardModelJobStatus

I can add an `assert( isinstance( ) )` - but personally I think The Python Way&trade; is to just go along - if the duck quacks and so on???

python/python/res/server/simulation_context.py
joakim-hove: Sure - whatever feels most natural from everest. Not obvious what would be good names though.

test-data/local/batch_sim/jobs/square_params.py
joakim-hove: I'll add a comment; the point was to get jobs with different status.

python/python/res/job_queue/forward_model_status.py
joakim-hove: I don't know - I can remove the option.

python/python/res/server/simulation_context.py
joakim-hove: Yes - I agree.

joakim-hove:
OK - I have fixed the obvious comments; please advise on:

1. Should we just say that before simulation has started it is not possible to get any detailed information? Thereby rendering all C changes obsolete.
2. Current api is essentially:
```python
global_status, details = ctx.job_status(id)
```
Not obvious to me what is the best; one method like above or e.g.:
```python
job_status = ctx.job_status(45)
job_details = ctx.job_details(45)
```
or ???

python/python/res/job_queue/forward_model_status.py
andreabrambilla: Well, is there any reason they should be accessed outside of this file? If not, I'd move them outside the class and make them private to the module. But it's not that important, it is more urgent to have this PR merged ;)

andreabrambilla:
Well done!

About point 1. I don't have strong opinions. 

About point 2. I prefer the 2 separate calls. One function should do be concerned only about one task

There is another point I'd like to bring up again. I have reread my comment about the `ForwardModelStatus.add_job` method and realized I missed the important part.
If we leave the interface as is, whoever wants to know the status of the forward model (one task) has to deal with 2 different classes (the Status and the JobStatus) and has to know how they should interact to accomplish in order to complete the task. As a user, I'd rather have only one entry point only to use one feature.
As a matter of fact, there is really no reason for the `ForwardModelJobStatus` to be visible outside the module. Or, even better, it doesn't even need to be a class, it can be replaced with a dictionary or a namedtuple, and all the "logic" (which is really not much) can be put in `ForwardModelStatus`.

markusdregi:
Well done Joakim!

> 1. Should we just say that before simulation has started it is not possible to get any detailed information? Thereby rendering all C changes obsolete.

Is there any detailed information anyway before the simulation has started? I guess we can handle that scenario quite gracefully, even though the file is not present..(?) I would prefer a pure Python approach, but that is just me.

joakim-hove:
> Is there any detailed information anyway before the simulation has started?

Not much - it is the list of job names.

joakim-hove:
> As a matter of fact, there is really no reason for the ForwardModelJobStatus to be visible outside the module. Or, even better, it doesn't even need to be a class, it can be replaced with a dictionary or a namedtuple, and all the "logic" (which is really not much) can be put in ForwardModelStatus.

Of course all of this is correct; but I guess it is a matter of personal taste and I feel the small `ForwardModelJobStatus` represents much stronger typing and thereby information to the user than a dict or named tuple. If you invoke the `__str__` method of one such instance you will see very clearly what it contains.

> whoever wants to know the status of the forward model (one task) has to deal with 2 different classes (the Status and the JobStatus)

markusdregi:
>> Is there any detailed information anyway before the simulation has started?

>Not much - it is the list of job names.

My opinion is that it is perfectly fine to have no detailed status information for each job before that specific forward model is indeed running.

joakim-hove:
OK - I have addressed the majority of the comments; thank you for the review. Merging.

PR #251, opened 2018-02-27
Added start_callback callable to the BatchSimulator

joakim-hove:
**Task**
Add (optional) callable argument to the batch simulator which can be used to pass simulation context elsewhere.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/python/python/res/simulator/batch_simulator.py
+++ b/python/python/res/simulator/batch_simulator.py
@@ -87,7 +87,7 @@ class BatchSimulator(object):
                 node.save(file_system, node_id)
 
 
-    def start(self, case_name, case_data):
+    def start(self, case_name, case_data, callback=None):
         """Will start batch simulation, returning a handle to query status and results.
 
         The start method will submit simulations to the queue system and then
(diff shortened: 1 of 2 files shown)


python/python/res/simulator/batch_simulator.py
```diff
@@ -87,7 +87,7 @@ def _setup_case(self, case, file_system):
                 node.save(file_system, node_id)
 
 
-    def start(self, case_name, case_data):
+    def start(self, case_name, case_data, start_callback=None):
```
pgdr: This looks very good, but is there a reason you named if `start_callback` instead of `callback`?   Are we going to support more callbacks to this function?

pgdr (approved):
This seems ready to merge, but I would prefer the `start_callback` being called `callback` unless we have a good reason for it.

In addition, we could write in the documentation that we strongly recommend users of the api to design the callback function to take

```python
def callback(*args, **kwargs)
```

so that we later can change the call signature if we want; possibly like

```python
callback(context=sim_context, event='start')
```

That way, users can in the future filter on events etc.

joakim-hove:
> but I would prefer the start_callback being called callback unless we have a good reason for it.

No good reason - I will rename to just `callback`.

joakim-hove:
> `def callback(*args, **kwargs)`

That was a very good idea.

joakim-hove:
Updated; I chose to keep the name `start_callback` in the `TestMonitor` class - as that could - in principle -  get more callback methods in the future - happy to change that as well if requested.

pgdr (approved):
Very good, please merge at will! :)

joakim-hove:
Cool - thank you!

PR #252, opened 2018-02-28
Fix batch simulator job check

markusdregi:
**Task**
_`BatchSimulator` mistakenly reported that simulations succeed/failed._

**Approach**
_The implementation was passing the `sim_id` to the queue to get the information, while the queue expects the `queue_idx` instead. Used the function from `SimulationContext` wrapping this logic._

**Pre un-WIP checklist**


markusdregi:
Jenkins initiated..

joakim-hove (approved):

PR #253, opened 2018-02-28
Fewer prints

pgdr:
My deepest apologies for big diffs, I've got a bad case of OCD.

Will revert and remove commits if nobody wants to review this mess.

### DIFF
--- a/libenkf/src/gen_data_config.c
+++ b/libenkf/src/gen_data_config.c
@@ -26,6 +26,7 @@
 #include <ert/util/bool_vector.h>
 #include <ert/util/string_util.h>
 #include <ert/util/type_macros.h>
+#include <ert/res_util/res_log.h>
 
 #include <ert/config/config_parser.h>
 #include <ert/ecl/ecl_util.h>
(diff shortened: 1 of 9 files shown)


libenkf/src/gen_data_config.c
```diff
@@ -397,6 +397,49 @@ void gen_data_config_assert_size(gen_data_config_type * config , int data_size,
...
+    config->mask_modified = true;
+  }
+
+  if (!config->mask_modified)
+    return; // nothing to do
+
+  // The global mask has been modified after the last load;
+  // i.e. we update the on-disk representation.
+  char * filename = util_alloc_sprintf("%s_active", config->key);
+  FILE * stream = enkf_fs_open_case_tstep_file(
```
joakim-hove: The mother of all irrelevant comments; I would like if this was formatted with:

1. The first argument, `forward_load_context_get_...` one line up.
2. Everything indented to the column of the opening `(`

That way I think it is easier to see that this is indeed one (too) long function call, and not regular lines of code.

libenkf/src/gen_data_config.c
```diff
@@ -407,77 +450,40 @@ void gen_data_config_assert_size(gen_data_config_type * config , int data_size,
...
-bool gen_data_config_has_active_mask( const gen_data_config_type * config , enkf_fs_type * fs , int report_step) {
-  bool   has_mask;
-  {
-    char * filename = util_alloc_sprintf("%s_active" , config->key );
-    FILE * stream   = enkf_fs_open_excase_tstep_file( fs , filename , report_step);
+bool gen_data_config_has_active_mask(const gen_data_config_type * config,
+                                     enkf_fs_type * fs,
+                                     int report_step) {
+  char * filename = util_alloc_sprintf("%s_active" , config->key );
+  FILE * stream = enkf_fs_open_excase_tstep_file(fs, filename, report_step);
```
joakim-hove: In situations where several resources are aquired and then released again I think it is a good pattern to release in opposite order - i.e. in this case that would be:
```C
char * filename = util_alloc...
FILE * stream = enkf_fs_ ..
...
...
fclose(stream);
free(filename)
```

joakim-hove:
Some minor comments - address them if you agree.

Since it is marked as WIP I have not approved it; but modulo the test failure I would be happy to merge it in the current state. 

I go offline now - so I leave it to you; or someone else to push final green button.

pgdr:
Rebased, but Travis seems red to me, I don't know if it's me, since I can't reproduce locally.

pgdr:
Killed in #255

PR #261, opened 2018-03-09
Early exit on exitfile

pgdr:
If `EXIT` was written between first check and `job_queue->max_ok_wait_time`, we would not discover, and thus wait the full `job_queue->max_ok_wait_time`.

Check if `EXIT` is written, and if it is, return false immediately.


joakim-hove (approved):

joakim-hove:
Looks good - merge when green

PR #263, opened 2018-03-16
Callback argument moved from start method to __init__

joakim-hove:
**Task**
Moved the callback argument from  `BatchSimulator.start()` to `BatchSimulator.__init__()`. Actually having it on the `start()` method feels better, but due to the usage in the everest <-> seba <-> libres triangle moving it to the `__init__()` seems simpler. 

**Pre un-WIP checklist**

### DIFF
--- a/python/python/res/simulator/batch_simulator.py
+++ b/python/python/res/simulator/batch_simulator.py
@@ -9,7 +9,7 @@ def _slug(entity):
 
 class BatchSimulator(object):
 
-    def __init__(self, res_config, controls, results):
+    def __init__(self, res_config, controls, results, callback = None):
         """Will create simulator which can be used to run multiple simulations.
 
         The @res_config argument should be a ResConfig object, representing the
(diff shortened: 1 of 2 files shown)


pgdr:
We had a discussion in the previous PR and agreed to name it `callback`, why is it back to `start_callback`?

joakim-hove:
Because now it is no longer bound to the `start` method, can still go back to `callback `.

pgdr:
> Because now it is no longer bound to the start method, can still go back to callback.

I don't really understand what you mean.


What is the reason behind this PR, why don't we just leave it like it was in #251 ?  Did I miss this in a planning session?

joakim-hove:
> I don't really understand what you mean.

Well - my point is that when the callback is passed in to the `__init__()` method it is not clear when the callback is invoked, and naming it `start_callback` would clarify that.

> What is the reason behind this PR

Well Elena and myself discussed it, and concluded this would be a more convenient way to do it. But if you are implementing the monitoring changes - please feel free to close or merge this as you see fit.

andreabrambilla (approved):
Looks good :+1:
And I think it makes a lot of sense. Whoever uses the simulator (ie, whoever calls `start`, ie, the optimizer) should not be concerned about setting up the simulator. And, whoever calls `start` has direct access to the context object, so there is no need for the callback in that case.

PR #266, opened 2018-03-20
Add safe check of queue index

pgdr:
If you check `SimulationContext.progress` too early, libres will crash with
```
Error message: run_arg_get_queue_index: sorry internal error - asking for the queue_index in a not-initialized run_arg object.
```

This is due to job not even submitted before the queue index is looked up.

Check that job is submitted before calling `get_queue_index`.


joakim-hove:
Could add a 
```python
with self.assertRaises(ValueError):
       run_arg.getQueueIndex()
```
here: https://github.com/Statoil/libres/blob/master/python/tests/res/enkf/test_run_context.py#L28

Otherwise - merge when Travis is satisfied.

joakim-hove (approved):

pgdr:
Thanks, added the test.  Running tests now, then I'll merge when green.

joakim-hove:
Now - this is *really* approved!

PR #270, opened 2018-03-22
Del callback

pgdr:
Attempt to solve an issue with callbacks from Everest

### DIFF
--- a/python/python/res/simulator/batch_simulator.py
+++ b/python/python/res/simulator/batch_simulator.py
@@ -1,3 +1,5 @@
+import weakref
+
 from ecl.util.util import BoolVector
 
 from res.enkf import ResConfig, EnKFMain, EnkfConfigNode, EnkfNode, NodeId
@@ -65,7 +67,9 @@ class BatchSimulator(object):
         self.ert = EnKFMain(self.res_config)
         self.control_keys = tuple(controls.keys())
         self.result_keys = tuple(results)
-        self.callback = callback
+
+        # must be weakref to avoid circular references
+        self._callback = weakref.ref(callback) if callback else None
 
         ens_config = self.res_config.ensemble_config
         for control_name, variable_names in controls.items():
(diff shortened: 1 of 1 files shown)


joakim-hove (commented):
Good work, I agree this is probably the culprit, however this fix will not fly. After moving the callback to the class definition it will be reused each time the `start` method is called, and this PR will render the simulator broken after the first update. 

I guess using a weak reference  either here or on callsite is the way to go.

python/python/res/simulator/batch_simulator.py
```diff
@@ -65,7 +67,9 @@ def callback(*args, **kwargs):
         self.ert = EnKFMain(self.res_config)
         self.control_keys = tuple(controls.keys())
         self.result_keys = tuple(results)
-        self.callback = callback
+
+        # must be weakref to avoid circular references
+        self._callback = weakref.ref(callback) if callback else None
```
joakim-hove, line 72: I guess this could just be `self._callback = weakref.ref(callbak)` - if you have a weak reference to `None`that will not harm anyone?

andreabrambilla: Tried, and `TypeError: cannot create weak reference to 'NoneType' object`

andreabrambilla (approved):
:+1:

joakim-hove (commented):
Is this still on?

pgdr:
Yes

ghost:

markusdregi:
This was resolved by other measures right? Should we close it?

andreabrambilla:
Yes, the original issue was solved by making seba deleting the Simulator once it's done using it, so yes, we can close this.

However, I still find a bit unreasonable that a callback keeps an entire object alive, I find it legitimate that callbacks are stored by non-owning references

markusdregi:
Created issue #462

PR #275, opened 2018-04-05
Forward model status

joakim-hove:
The `start_time`and `end_time`attributes in the `status.json` file were never set.


python/tests/res/job_queue/test_jobmanager.py
```diff
@@ -376,6 +377,15 @@ def test_data_from_forward_model_json(self):
...
 
+    def test_complete(self):
+        with TestAreaContext("json_from_forward_model_NO_DATA_ROOT"):
+            with open("jobs.json", "w") as f:
+                f.write(JSON_STRING_NO_DATA_ROOT)
+
+            jobm = JobManager()
+            jobm.complete()
+            st = ForwardModelStatus.load(os.getcwd())
+            self.assertTrue( isinstance(st.end_time, datetime.datetime))
```
markusdregi, line 388: Looks good! Could you add a statement checking that `start_time` is also set? Perhaps also checking that they are close to `now`?

joakim-hove: OK - done.

markusdregi (approved):
Nice and clean fix. Merge when green.

PR #277, opened 2018-04-10
Expose LSF_RESOURCES and LSF_SERVER in keys

pgdr:
We expose two more LSF related keys in `ConfigKeys`
* `LSF_RESOURCES_KEY`
* `LSF_SERVER_KEY`

There was some internal confusion regarding `queue_name` and `queue_system`, that is, the functions named `queue_name` were used to get the `queue_system` information (e.g. LSF/LOCAL/TORQUE), but this has now been changed to `queue_system`.

Furthermore, `LSF_RESOURCE` was named `LSF_RESOURCES` several places, and one place, something is called `LSF_REQUEST`.  The _actual IBM concept_ is called **Resource requirements**.

I've tried to change most of them to `LSF_RESOURCE`.


joakim-hove (dismissed):

pgdr:
Ready for review

lib/include/ert/enkf/config_keys.h
```diff
@@ -102,8 +102,8 @@ extern "C" {
 #define  LOCAL_CONFIG_KEY                  "LOCAL_CONFIG"
 #define  LOG_FILE_KEY                      "LOG_FILE"
 #define  LOG_LEVEL_KEY                     "LOG_LEVEL"
-#define  LSF_QUEUE_KEY                     "LSF_QUEUE"
-#define  LSF_RESOURCES_KEY                 "LSF_RESOURCES"
+#define  LSF_QUEUE_KEY                     "LSF_QUEUE" // The queue name, e.g. "mr"
```
joakim-hove, line 105: Sorry - I suggest these keys are left unmodified; maybe with a deprecation warning. The keys which are actually accessed through the dictionary based api - and hence everest are [these keys](https://github.com/Statoil/libres/blob/master/lib/include/ert/job_queue/lsf_driver.h#L32).

python/res/enkf/queue_config.py
```diff
@@ -57,12 +58,25 @@ def max_submit(self):
...
     def driver(self):
-        return self._queue_driver(self.queue_name).setParent(self)
+        return self._queue_driver(self.queue_system).setParent(self)
 
     @property
     def job_script(self):
         return self._get_job_script()
+
+    @property
+    def lsf_resource(self):
```
joakim-hove: This should in my opinion explicitly ask for the LSF driver, and then return resource. The following situation is realistic:

1. The user has configured to use the `LOCAL` system.
2. When asking for the `lsf_resource` the queue config will look up the active - i.e. "LOCAL" system, and then query that for lsf resource -> Exception.

joakim-hove:
I realize this is quite messy in the current situation, and appreciate the cleanup resulting from your effort.

pgdr:
Thanks for the review, I had to fix a bug first, and now I will address your comments.

python/res/enkf/queue_config.py
pgdr: done

lib/include/ert/enkf/config_keys.h
pgdr: I think that's done now?  Can you check and see if this is satisfactory?  (Perhaps I've misunderstood something, though.)

joakim-hove (approved):

PR #279, opened 2018-04-18
Json env fix

joakim-hove:
Fix in outputtted json when the environment is set in forward models.

**Pre un-WIP checklist**


pgdr:
Perhaps we could consider using a JSON library instead of rolling our own?

joakim-hove:
> Perhaps we could consider using a JSON library instead of rolling our own?

We currently only write a very limited amount of json - no parsing; but sure - feel free to submit a PR; it will probably be merged.

PR #280, opened 2018-04-19
Add logging in `job_node`

pgdr:
During debugging Ert runs, I added some logging (info and debug levels) for job submissions and manipulations.

I also added a couple early returns.

### DIFF
--- a/lib/job_queue/job_node.c
+++ b/lib/job_queue/job_node.c
@@ -27,6 +27,7 @@
 #include <ert/util/util.h>
 #include <ert/util/arg_pack.h>
 #include <ert/util/type_macros.h>
+#include <ert/res_util/res_log.h>
 
 #include <ert/job_queue/job_node.h>
 
@@ -177,7 +178,7 @@ void job_queue_node_set_queue_index( job_queue_node_type * node , int queue_inde
   if (node->queue_index == INVALID_QUEUE_INDEX)
     node->queue_index = queue_index;
   else
-    util_abort("%s: internal error: atteeempt to reset queue_index \n",__func__);
+    util_abort("%s: internal error: attempt to reset queue_index \n", __func__);
 }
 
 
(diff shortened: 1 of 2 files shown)


lib/job_queue/job_node.c
```diff
@@ -271,66 +272,65 @@ job_queue_node_type * job_queue_node_alloc( const char * job_name ,
...
-      node->max_confirm_wait= 60*2; /* 2 minutes before we consider job dead. */
-    }
-
-    pthread_mutex_init( &node->data_mutex , NULL );
-    return node;
-  } else
+  if (!util_is_directory( run_path ))
     return NULL;
+
+  res_log_fdebug("alloc job node for %s", job_name);
```
joakim-hove: This is so much 'normal operation' that I would not say it warrants a message at all.

lib/job_queue/job_node.c
```diff
@@ -271,66 +272,65 @@ job_queue_node_type * job_queue_node_alloc( const char * job_name ,
...
+  node->progress_timestamp = time(NULL);
+  UTIL_TYPE_ID_INIT( node , JOB_QUEUE_NODE_TYPE_ID );
+
+  /* The data initialized in this block should *NEVER* change. */
+  node->job_name       = util_alloc_string_copy( job_name );
+
+  if (util_is_abs_path(run_path))
+    node->run_path = util_alloc_string_copy( run_path );
+  else
+    node->run_path = util_alloc_realpath( run_path );
```
joakim-hove: This could just be `node->run_path = util_alloc_realpath( run_path )` unconditionally.

lib/job_queue/job_node.c
```diff
@@ -271,66 +272,65 @@ job_queue_node_type * job_queue_node_alloc( const char * job_name ,
...
+  if (ok_file)
+    node->ok_file = util_alloc_filename(node->run_path , ok_file , NULL);
+  else
+    node->ok_file = NULL;
+
+  if (status_file)
+    node->status_file = util_alloc_filename(node->run_path , status_file , NULL);
+  else
+    node->status_file = NULL;
+  node->confirmed_running = false;
```
joakim-hove: The `node->confirmed_running = false` assignment could be moved out of the file assignments.

lib/job_queue/job_node.c
```diff
@@ -415,69 +415,89 @@ void job_queue_node_run_EXIT_callback( job_queue_node_type * node ) {
...
+                                             node->argc,
+                                             (const char **) node->argv);
+  if (job_data == NULL) {
+    /*
+      In this case the status of the job itself will be
+      unmodified; i.e. it will still be WAITING, and a new attempt
+      to submit it will be performed in the next round.
+    */
+    submit_status = SUBMIT_DRIVER_FAIL;
+    res_log_finfo("job data null on submit for %s(%d)",
```
joakim-hove: I guess a better log message would be something like: "Failed to submit job: %s(d)"?

lib/job_queue/job_node.c
```diff
@@ -415,69 +415,89 @@ void job_queue_node_run_EXIT_callback( job_queue_node_type * node ) {
...
+                  node->job_name,
+                  node->queue_index);
+    goto cleanup;
```
joakim-hove: That was a new one - but I like it!

lib/job_queue/job_node.c
```diff
@@ -513,53 +533,65 @@ static void job_queue_node_update_timestamp(job_queue_node_type * node) {
...
+  job_status_type current_status = job_queue_node_get_status(node);
+
+  bool confirmed = job_queue_node_status_update_confirmed_running__(node);
+
+  if ((current_status & JOB_QUEUE_RUNNING) && !confirmed) {
+    // it's running, but not confirmed running.
+    double runtime = job_queue_node_time_since_sim_start(node);
+    if (runtime >= node->max_confirm_wait) {
+      res_log_finfo("max_confirm_wait (%d) has passed since sim_start"
+                    "without success; %s(%d) is dead",
```
joakim-hove: Actually for these log messages it is misleading to include the integer `queue_index`, that index is an internal number which is semi-random, and not anything the user can relate to.

joakim-hove:
This is good - I like the early returns, although I miss some of my lexical scoping. I am happy to see the use of the log system.

Some minor comments - I guess [this](https://github.com/Statoil/libres/pull/280/files#r182877768) is the most important - address it if you agree.

joakim-hove (approved):
Nice!

pgdr:
I have addressed all the comments.

joakim-hove:
> I have addressed all the comments.

Good - then we let travis do it's job and merge.

PR #281, opened 2018-04-25
Some minor python 3 fixes.

stefoss23:
**Task**
Because the python 3 version of libecl/libres is currently on "Allowed Failures", python3-incompatible code can still be accepted by TRAVIS. Here are some fixes to that. 

**Approach**
tests/__init__.py: removed types.ClassType.
ui_return.py: added __bool__ method, replacing the __nonzero__ from python 2. 

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/python/res/util/ui_return.py
+++ b/python/res/util/ui_return.py
@@ -43,19 +43,13 @@ class UIReturn(BaseCClass):
         else:
             raise ValueError('Unable to construct UIReturn with status = %s' % str(status))
 
-
+    #For python 3, corresponds to __nonzero__
     def __bool__(self):
-        if self.status() == UIReturnStatusEnum.UI_RETURN_OK:
-            return True
-        else:
-            return False
+        return self.status() == UIReturnStatusEnum.UI_RETURN_OK
 
-
+    #For python 2
     def __nonzero__(self):
-        if self.status() == UIReturnStatusEnum.UI_RETURN_OK:
-            return True
-        else:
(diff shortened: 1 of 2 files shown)


python/res/util/ui_return.py
```diff
@@ -44,6 +44,13 @@ def __init__(self , status):
             raise ValueError('Unable to construct UIReturn with status = %s' % str(status))
 
 
+    def __bool__(self):
+        if self.status() == UIReturnStatusEnum.UI_RETURN_OK:
+            return True
+        else:
+            return False
```
pgdr: please just return `self.status() == UIReturnStatusEnum.UI_RETURN_OK` and rewrite `__nonzero__` to `self.__bool__()` or something appropriate

stefoss23: I'll do the first, but it needs both __nonzero__ (for py2) and __bool__(for py3), but I shall add some comments to that.

pgdr: Yes, but you don't need two different implementations. :)

I was thinking something like
```python
def __bool__(self):
    return self.status() == UIReturnStatusEnum.UI_RETURN_OK

def __nonzero__(self):
    return self.__bool__()
```

stefoss23: That would work better, yes.

joakim-hove (approved):

PR #289, opened 2018-05-06
Make python 3 work in ert. 

stefoss23:
**Task**
Make python3 work in statoil/ert.

**Approach**
Changes in the share directory. 

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/share/workflows/jobs/internal-gui/scripts/csv_export.py
+++ b/share/workflows/jobs/internal-gui/scripts/csv_export.py
@@ -2,7 +2,13 @@ import os
 import re
 
 import pandas
-from PyQt4.QtGui import QCheckBox
+import sys
+
+if sys.version_info[0] == 2:
+  from PyQt4.QtGui import QCheckBox
+else:
+  from PyQt5.QtWidgets import QCheckBox
+
 
 from res.enkf import ErtPlugin, CancelPluginException
 from res.enkf.export import SummaryCollector, GenKwCollector, MisfitCollector, DesignMatrixReader, CustomKWCollector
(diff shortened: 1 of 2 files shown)


share/workflows/jobs/internal-gui/scripts/csv_export.py
```diff
@@ -2,7 +2,13 @@
 import re
 
 import pandas
-from PyQt4.QtGui import QCheckBox
+import sys
+
+if sys.version_info[0] == 2:
+  from PyQt4.QtGui import QCheckBox
+else:
+  from PyQt5.QtWidgets import QCheckBox
```
joakim-hove: Instead of explicitly testing Python version I would hope this could be:
```
try:
   from PyQt4.QtGui import QCheckBox
except ImportError:
   from PyQt5.QtWidgets import QCheckBox
```

pgdr: Are we going to Qt5 in 2.3?

joakim-hove: As I see it this is mainly a change to *prepare* for Python 3; since Python 3 requries Qt5 we have this thing going. I had not thought we would go for Qt5 in 2.3.

JensGM: It might be better to use six here to check python version. It has two flags you can check for, I think they are `six.PY2` and `six.PY3`

joakim-hove (approved):

PR #294, opened 2018-05-09
Revert  Refactor job_node and add logging

pgdr:
Revert one of the changes in #280 that introduced a crash if CWD is not writable.


joakim-hove (approved):
Go

joakim-hove (approved):
Go

PR #310, opened 2018-05-29
[WIP] Correctly initializes nodes during updates

andreabrambilla:
@joakim-hove @lars-petter-hauge 

I guess this should fix the problem (please have a look). It builds on my machine, I haven't run the tests yet (but I guess that's pretty much pointless, otherwise we would have already caught the issue a long time ago)

**Pre un-WIP checklist**

### DIFF
--- a/lib/enkf/enkf_main.c
+++ b/lib/enkf/enkf_main.c
@@ -681,7 +681,8 @@ static int enkf_main_serialize_dataset( const ensemble_config_type * ens_config 
   return matrix_get_rows( A );
 }
 
-static void deserialize_node( enkf_fs_type * fs,
+static void deserialize_node( enkf_fs_type * target_fs,
+                              enkf_fs_type * src_fs,
                               const ensemble_config_type * ensemble_config,
                               const char * key ,
                               int iens,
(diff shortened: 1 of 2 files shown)


lib/enkf/enkf_main.c
```diff
@@ -691,10 +692,17 @@ static void deserialize_node( enkf_fs_type * fs,
...
   const enkf_config_node_type * config_node = ensemble_config_get_node( ensemble_config , key );
-  enkf_node_type * node = enkf_node_alloc( config_node );
+
   node_id_type node_id = {.report_step = target_step, .iens = iens  };
-  enkf_node_deserialize(node , fs , node_id , active_list , A , row_offset , column);
-  state_map_update_undefined(enkf_fs_get_state_map(fs) , iens , STATE_INITIALIZED);
+  enkf_node_type * node = enkf_node_alloc( config_node );
+
+  // If partly active, init node from source fs (deserialize will fill it only in part)
+  if(active_list_get_mode(active_list) == PARTLY_ACTIVE)
```
joakim-hove: I think this should be: `active_list_get_mode(active_list) != ALL_ACTIVE)` - it is not that difficult to envision the slightly pathological case where no elements are added to the vector and the mode is `INACTIVE` - which would fail now.

joakim-hove (approved):
One minor comment - otherwise this is ready! This PR shows an impressively high: `depth_of_code_change / time_spent_in_code` ratio!

lib/enkf/enkf_main.c
andreabrambilla: ha, I had it like that before and decided to change it right before committing xD
I'll fix it

PR #333, opened 2018-07-14
Add eclipse forward model to libres

joakim-hove:
Move the eclipse forward model code from ert-statoil to libres.


python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,200 @@
...
+                raise OSError("The mpirun argument: '{}' is not executable by user".format(executable))
+
+        self.mpirun = mpirun
+
+
+
+class EclConfig(object):
+    # The config_file is defined here as a class attribute to allow
+    # for monkey pathching in tests.
+    config_file = os.path.join(os.path.dirname(__file__), "ecl_config.yml")
```
andreabrambilla: I think "for testing purposes" is usually not the best reason for doing something.
What about calling this `DEFAULT_CONFIG_FILE`, and modify `__init__` as follows:
```Python
def __init__(self, config_file=None):
  if config_file is None:
    config_file = self.DEFAULT_CONFIG_FILE
  with open(config_file) as f: 
    ...
```

andreabrambilla:
I did not review this very thoroughly, as I presume this code is mostly copied as is from ert-statoil.
On Friday I pretty much started doing the same thing, and I stopped the moment I began reading `ecl_config.py`. That module has a number of module-level variables/constants that refer to paths that exists only in Equinor. 
Which is definitely practical for us, but I think they should not be in `libres`. A possibility could be a site-config file (and tehre is already something like that in ert/libres, right?)

joakim-hove:
> I did not review this very thoroughly, 

Well - thank you for looking anyway; I hope to have it ready from my hand later today.

> as I presume this code is mostly copied as is from ert-statoil.

Yes - that is quite true. The `EclRun` is copied 98% verbatim, but the `EclConfig` is new - that used to have configuration in code; now that is  yml formatted file. The intention is that a site specific config file `ecl_config.yml`is copied in as:
```bash
$prefix/lib/python{x.y}/site-packages/res/fm/ecl/ecl_config.yml
```
as part of/after the normal libres installation.


> That module has a number of module-level variables/constants that refer to paths that exists only in Equinor.

They have gone now - they were just part of WIP for editing convenience.

python/res/fm/ecl/ecl_config.py
joakim-hove: > I think "for testing purposes" is usually not the best reason

I agree - but maybe this time ..... The reason was reluctant to do as you suggest is for the testing of the `EclRun` class. That class creates a config object like:
```
class EclRun(object):

    def __init__(self, ...):
        self.config = EclConfig( )
```

but - OK; also that constructor could get an optional `config_file` argument; I guess that actually is a better solution - comment driven development.

andreabrambilla:
Awesome, let me know when this is ready and I'll take a second look at it ;)

python/res/fm/ecl/ecl_config.py
joakim-hove: What about an environment variable - `ECL_CONFIG_FILE`?

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+import sys
+
+
+def _replace_env_string(value):
+    while True:
+        dollar_pos = value.find("$")
+        if dollar_pos < 0:
+            break
+
+        end_pos = value.find(":", dollar_pos)
```
andreabrambilla: I don't fully understand why you are looking for the `':' ` character in the `value`, shouldn't you look for a space (or, actually, anything that is not alphanumeric)

EDIT: ok, after seeing the .yml file I got it. Given how this is used, I guess checking for `:` is enough. If we'll ever need to support stuff like `''$HOME/bin/ecl"` we can improve it later

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+# Small utility function will take a dict as input, and create a new dictionary
+# where all $VARIABLE in the values has been replaced with getenv("VARIABLE").
+
+def _replace_env(env):
+    new_env = {}
+    for key,value in env.items():
+        if "$" in value:
+            new_env[key] = _replace_env_string(value)
+        else:
+            new_env[key] = value
```
andreabrambilla: this if is unnecessary running replace on a string without `$` would return the string as is anyway

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+        self.env = env
+
+        if mpirun is None:
+            self.mpi = False
+        else:
+            self.mpi = True
+            if not os.access(mpirun, os.X_OK):
+                raise OSError("The mpirun argument: '{}' is not executable by user".format(executable))
+
+        self.mpirun = mpirun
```
andreabrambilla: `self.mpi` is a bit redundant, checking `self.mpirun is not None` is equally effective

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+class EclConfig(object):
+    """Represent the eclipse configuration at a site.
+
+    The EclConfig class internalizes information of where the various eclipse
+    programs are installed on site, and which environment variables need to be
+    set before the simulator starts. The class is based on parsing a yaml
+    formatted configuration file, the source distribution contains commented
+    example file.
+
+    """DEFAULT_CONFIG_FILE = os.path.join(os.path.dirname(__file__), "ecl_config.yml")
```
andreabrambilla: Not wrong, but I'd insert a `\n` after the triple quotes

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+        env.update( self._config.get("env", {} ))
+
+        mpi_sim = self._config["simulators"][sim][version][exe_type]
+        env.update( mpi_sim.get("env", {}))
+        return _replace_env(env)
+
+    def has_sim(self, sim, version):
+        try:
+            simulator = self._get_sim(sim, version, "scalar")
+            return False
```
andreabrambilla: I guess this should be return `True`

Still, from a design point of view, I see no point in have a `has_XXX`method which is as expensive as `get_XXX`, as they would lead to code like
```Python
if has_XXX(a, b, c):
  x = get_XXX(a, b, c)
```
which executes practically the same code twice

In a situation like this, I would simply get rid of the `has_XXX`, so the user will have to write

```Python
x = get_XXX(a, b, c)
if not x:    # or, if the get throws, this won't even be needed
  return
```

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+        except:
+            return False
+
+    def has_mpi_sim(self, sim, version):
+        try:
+            simulator = self._get_sim(sim, version, "mpi")
+            return False
```
andreabrambilla: Same here

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+    def _get_sim(self, sim, version, exe_type):
+        d = self._config["simulators"][sim][version][exe_type]
+        if exe_type == "mpi":
+            mpirun = d["mpirun"]
+        else:
+            mpirun = None
+        return Simulator(sim, version, d["executable"], self._get_env(sim, version, exe_type), mpirun = mpirun)
```
andreabrambilla, line 83: what about making the strings `env`, `executable`, `mpi`, `scalar`, `simulators` constants?

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+        """Will return a small struct describing the simulator.
+
+        The struct has attributes 'executable' and 'env'. Observe that the
+        executable path is validated when you instantiate the Simulator object;
+        so if the executable key in the config file points to non-existing file
+        you will not get the error before this point. You can check with
+        has_sim() first to be certain.
+
+        """
+        return self._get_sim(sim, version, "scalar")
```
andreabrambilla: I guess we could merge `sim`, `mpi_sim` and `get_sim` into something like `def sim(self, sim, version, use_mpi=False)`

python/res/fm/ecl/ecl_config.py
```diff
@@ -0,0 +1,149 @@
...
+                            sim = self._get_sim(name, version, exe_type)
+                        except Exception:
+                            sys.stderr.write("Failed to create simulator object for:{name} version:{version} {exe_type}\n".format(name=name,
+                                                                                                                                  version=version,
+                                                                                                                                  exe_type=exe_type))
+                            sim = None
+
+                    if sim:
+                        simulators.append(sim)
+        return simulators
```
andreabrambilla, line 119: This is just my opinion, feel free to disagree ;)
I think this method is overly complicated. I would expect `simulators` to give me a list of available simulators. If retrieving one specific simulator fails, I would simply exclude it from the returned list.
In other words, I'd get rid of `strict` and have only the `else` branch, even without the `write`

python/res/fm/ecl/ecl_run.py
```diff
@@ -0,0 +1,385 @@
...
+       run = EclRun(sys.argv)
+       run.runEclipse( )
+
+
+    Name of executable
+    ==================
+
+    Observe that the EclRun.__init__() initializer determines which simulator
+    to use by checking the value of sys.argv[0] - i.e. typically the name of
+    the executable. The currently recognized names are:
```
andreabrambilla, line 76: I know it was like this before and it works, but I honestly find this quite ugly.  Can we consider changing the constructor of EclRun to something like `def __init__(self, simulator, version, data_file, num_cpu = 1)` ?

python/res/fm/ecl/ecl_run.py
```diff
@@ -0,0 +1,385 @@
...
+                self.num_cpu = 1
+                self.sim = config.sim( simulator, self.version )
+            else:
+                self.num_cpu = int(argv[3])
+                self.sim = config.mpi_sim( simulator, self.version)
+
+        else:
+            raise ValueError("Must have three/four arguments")
+
+
```
andreabrambilla, line 157: I have only quickly skimmed through the rest of EclRun, as I guess it was not significantly modified wrt ert-statoil (and it also includes a good deal of implementation details I am not too familiar with). If you'd like me to check this in more detail let me know.

python/tests/__init__.py
```diff
@@ -67,6 +68,7 @@ def skip_wrapper(*args, **kwargs):
         return test_item
     return decorator
 
+simulator_test = statoil_test
```
andreabrambilla: This is just minor, but I'd get rid of `simulator_test`

`statoil_test` is well documented and communicates the requirements for running a test (ie, that statoil test data is available).
`simulator_test` I guess refers to what is being tested (the simulator). If someone does not know that it is just an alias for `statoil_test`, it does not provide much more information

python/tests/res/fm/test_ecl_config.py
```diff
@@ -0,0 +1,120 @@
...
+
+            sim = conf.sim("ecl100", "2015")
+            self.assertEqual(sim.executable, scalar_exe)
+            self.assertIsNone(sim.mpirun)
+
+            with self.assertRaises(Exception):
+                simulators = conf.simulators()
+
+            simulators = conf.simulators(strict = False)
+            self.assertEqual(len(simulators), 2)
```
andreabrambilla, line 120: This test is a bit "heavy". It would have been maybe clearer if we had some test data in some test folder instead of generating it on the fly. And if different functionalities were tested in different test functions. 
But I'd say that for now this is enough, this is going to be one of the most used forward models, possible bugs would pop up rather quickly anyway

andreabrambilla (approved):
I had a bunch of comments about style and minor design choices, but overall I think this looks good.

I would have preferred to make also the API a bit "cleaner", similarly to what we did for Everest. something along the lines of
```Python
sim_context = fm.ecl.simulate('ecl100', '2015.02', SPE1.DATA, num_cpu=27)
```
But that is a significant change, and, since this used only a forward model, I am not sure there would be any considerable advantage besides aesthetics 

I guess the next thing to solve is where to store and how to distribute the site config file to our users, but of course that will be addressed separately

Well done :+1:

python/res/fm/ecl/ecl_config.py
joakim-hove: Yes - I was totally focused on `:` separated path variables; I have added a fix.

python/res/fm/ecl/ecl_config.py
joakim-hove: Fixed

python/res/fm/ecl/ecl_config.py
joakim-hove: Removed

python/res/fm/ecl/ecl_config.py
joakim-hove: Yes - done

python/res/fm/ecl/ecl_config.py
joakim-hove: I have remived the `has_` - the get will throw.

python/res/fm/ecl/ecl_config.py
joakim-hove: Yes - I agree that might be nice; they should never be visible outside the `EclConfig` class; and they must still be manually syncronised with the values in the yml file - that was at least the reason I did not do it initially.

python/res/fm/ecl/ecl_config.py
joakim-hove: I ran out comment-writing-steam; but the yaml file is - on purpose - not validated before you actually ask for a specific simulator. The thought was that if you misspelled `executbAle` for one of the simulators, the rest should still be usable.

The `simulators( strict = True )` method would be a simple method spin through *all* the simulators and verify their sanity; I envision using that in a pre-install test of the configuration file.

python/tests/__init__.py
joakim-hove: I have no strong feelings here; the point was that `@simulator_test` requires a simulator; and could in principle be run outside of Statoil as well. BUt of course - the actual data used in the test are still hardwired to Statoil / Equinor. I'll remove it.

joakim-hove:
OK - thank you for review; I have addressed many of your comments. When it comes to infering the simulator from the name of the executable and the somewhat cleaner API suggestion I would prefer to let them be as they are. Mainly because as you say this is *the* most used forward model, so I would like to not rock the boat more than necessary in one go.

The convention of encoding simulator in the name of the executable has the advantage that one only needs to maintain one script, and can add symlinks to enable more simulators. But I agree it is ugly and would not oppose changing it in the future.

I agree that your suggested api is nicer; but this API should not really be used as an API as I see it, the most relevant "API" is the executable commands and their commandline arguments; but again no opposition to change it in the future.

python/res/fm/ecl/ecl_run.py
joakim-hove: That is fine - it has worked for several years.

joakim-hove:
I added an alternative `simulate()` function - that was quite simple. I'll merge this tomorrow if all tests are green.

andreabrambilla:
Great, feel free to merge when green :+1:

PR #344, opened 2018-08-14
Add rms forward model component

joakim-hove:
Add a class for configuration and running of RMS as forward model.


markusdregi (approved):
With my limited knowledge of _RMS_ this looks good as far as I can see. As I said before I like the approach.

PR #345, opened 2018-08-14
Job queue uses basename

stefoss23:
**Task**
- Name of job in the queue system should only be the basename.
  E.g. if defined ECLBASE ABC/NAME, name of the job will be NAME.
- LSF driver shall write job_id_nr to a file "lsf_info.json" in runpath.

**Approach**
- lib/job_queue/job_node.cpp: 
  job_queue_node_alloc(...): job_name path reduced to basename. 

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


lib/job_queue/job_node.cpp
```diff
@@ -281,7 +283,9 @@ job_queue_node_type * job_queue_node_alloc( const char * job_name ,
   UTIL_TYPE_ID_INIT( node , JOB_QUEUE_NODE_TYPE_ID );
 
   /* The data initialized in this block should *NEVER* change. */
-  node->job_name       = util_alloc_string_copy( job_name );
+  std::string path     = job_name;
+  std::string basename = path.substr(path.find_last_of("/\\") + 1);
```
joakim-hove, line 287: What happens if there is no `/` in the string?

joakim-hove (approved):
One minor thing/question; otherwise this is ready to be merged.

lib/job_queue/job_node.cpp
stefoss23: Then std::string_path = job_name

PR #349, opened 2018-08-22
Use rng_std_normal()

joakim-hove:
This should + https://github.com/Statoil/libecl/pull/456 should fix a rare every-one-and-then problem with `nan` values in GEN_KW.


ghost:

PR #371, opened 2018-09-18
WIP: modifications due to linkage problems w/ util.hpp.

stefoss23:
**Task**
This is related to https://github.com/Statoil/libecl/pull/471. Some linkage problems arose. 

**Approach**
Some include statements moved out of extern C. Some references to util.hpp changed to util.h.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#471

### DIFF
--- a/lib/analysis/modules/rml_enkf.c
+++ b/lib/analysis/modules/rml_enkf.c
@@ -22,7 +22,7 @@
 #include <string.h>
 #include <stdio.h>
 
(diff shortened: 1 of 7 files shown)


joakim-hove:
The moving of the C++ compile guards is fine; but are the `util.h -> util.hpp` changes necessary? Would actually prefer to just have `util.h` as - that is the simplest.

ghost:

PR #375, opened 2018-09-20
WIP: Move shell from ert statoil

stefoss23:
**Task**
shell.py is moved to libres. 

**Approach**
Files were moved. 

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/python/res/fm/shell/__init__.py
+++ b/python/res/fm/shell/__init__.py
@@ -1,0 +1,1 @@
+from .shell import Shell
--- a/python/res/fm/shell/__init__.py
+++ b/python/res/fm/shell/__init__.py
@@ -1,8 +1,5 @@
 from .shell import Shell
 
-def prefix_path(input_path):
-  Shell.prefix_path(input_path)
-
 def symlink(target , link_name):
   Shell.symlink(target , link_name)
 
--- a/python/res/fm/shell/__init__.py
+++ b/python/res/fm/shell/__init__.py
@@ -1,1 +1,27 @@
 from .shell import Shell
+
+def prefix_path(input_path):
+  Shell.prefix_path(input_path)
+
+def symlink(target , link_name):
+  Shell.symlink(target , link_name)
+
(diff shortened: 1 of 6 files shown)


python/res/fm/shell/__init__.py
```diff
@@ -0,0 +1 @@
+from .shell import Shell
```
joakim-hove, line 1: Good if you in this `__init__` file also add small functions like:
```python
def move_file(src,target):
      Shell.moveFile(src,target)
```

so that everything can be called as:
```python
from res.fm.shell import move_file
...
move_file(f1,f2)
```

python/res/fm/shell/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+def prefix_path(input_path):
```
joakim-hove: This should not be exposed in `__init__.py`

python/res/fm/shell/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+  Shell.prefix_path(input_path)
+
+def symlink(target , link_name):
+  Shell.symlink(target , link_name)
+
+def mkdir(path):
+  Shell.mkdir(path)
+
+def moveFile(src_file , target):
```
joakim-hove: Please use `snake_case` - i.e. `move_file(src_file, target)`

python/tests/res/fm/test_shell.py
```diff
@@ -0,0 +1,322 @@
+import unittest
+import os
+import os.path
+
+from ecl.util.test import TestAreaContext
+from tests import ResTest
+from res.fm.shell import Shell
+
+
```
joakim-hove: In this test please also import the small functions from `shell/__init__.py`:
```python
from res.fm.shell import *
```
and replace some `Shell.method()`calls with calls to the thin wrapper.

PR #380, opened 2018-09-26
Remove shaders from stale utility

joakim-hove:


JensGM:
There was OpenGL in libres? :smile:

joakim-hove:
Hehe lightning speed and viz effects...

PR #399, opened 2018-10-09
Remove python from RMS 2013 path

markusdregi:
**Task**
RMS was failing when using version 2013, because it picked up the Komodo Python from PATH, and hence also Python 2 packages.

**Approach**
If version is 2013, filter path for folders containing Python when running RMS.

**Pre un-WIP checklist**

I have fixed some more minor Python3 things + made sure rms mock executable can run without Python

### DIFF
--- a/python/res/fm/rms/rms_run.py
+++ b/python/res/fm/rms/rms_run.py
@@ -25,7 +25,9 @@ def _remove_python_from_path(path):
     path_elems = []
     for elem in path.split(os.pathsep):
         if os.path.isfile(os.path.join(elem, python_exec)):
-            print("**Warning: the path: {} is removed from $PATH to avoid conflict with RMS internal python interprete".format(elem))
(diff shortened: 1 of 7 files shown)


python/res/fm/rms/rms_run.py
```diff
@@ -18,6 +20,17 @@ def pushd(path):
     os.chdir(cwd0)
 
 
+def _remove_python_from_path(path):
+    python_exec = 'python'
+    path_elems = []
+    for elem in path.split(os.pathsep):
+        if os.path.isfile(os.path.join(elem, python_exec)):
+            print("**Warning: the path: {} is removed from $PATH to avoid conflict with RMS internal python interprete".format(elem))
```
markusdregi: `interpreter`

markusdregi:
@joakim-hove Are we getting ready to merge this one?

joakim-hove:
Yes

PR #419, opened 2018-11-02
Always drop komodo PYTHONPATH before starting RMS

joakim-hove:


markusdregi (approved):
Very good that you patched this :+1:

PR #422, opened 2018-11-05
IES2

geirev:
**Task**
Implementation of a numerically stable version of Iterative Ensemble Smoother.
Loaded as external analysis module.
The current implementation should work as long as the same observations are used in each iteration.

**Approach**
Implemented as ies_enkf.c and using ies_enkf_config.c

Current issues in order of priority:  
1. Measurement perturbations are resampled every iteration.  They should not be resampled. The initial matrix E should be used for all subsequent iterations.
2. It seems like the update from the first iteration is ignored.
3. The number of active observations may change between iterations due to a repeated screening of observations. For the iterative methods, it would make more sense to only screen observations once initially.

In general:
The following classes of data assimilation methods should be supported by ERT:
* Non-iterative methods
   - With perturbed observations (EnKF ++)
   - Without perturbed observations (SQRT, ++)
* Iterative methods with perturbed observations:
   - ESMDA: Perturbed observations are resampled in every iteration
   - IES: Perturbed observations are sampled initially and not changed during iterations.
* Iterative methods without perturbed observations
   -Bocquet and Sakov type iterative SQRT methods may be implemented in the future.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

image


lib/analysis/modules/ert_module
```diff
@@ -8,7 +8,8 @@ ert_root = os.path.realpath( os.path.join(os.path.dirname( os.path.realpath( os.
 
 #-----------------------------------------------------------------
 
-default_lib_list          = ["analysis"   , "ert_util"]
+default_lib_list          = ["res"   , "ecl"]
+#default_lib_list          = ["analysis"   , "ert_util"]
```
joakim-hove: Just remove the commented out code.

joakim-hove:
OK - this was a handful; will take some time to review++ on this.

It looks like the PR contains several files which should not be part of the PR, as a very first step it would be good if you could remove these files:
```
lamda0.txt
lib/analysis/modules/.exrc
lib/analysis/modules/.laminc.txt.swp
lib/analysis/modules/compile.sh
lib/analysis/modules/ies.F90
lib/analysis/modules/laminc.txt
my_enkf.*
```
in addition there are quote a lot changes in the `test-data/local/snake_oil` directory. I assume those changes should be reverted. Would be good if you [rewrite the git history](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) by squashing / rebasing and then subsequently force push the branch again. That way all traces of the files which should not have been there will be removed - that will simplify the review process!

One other immediate comment is that it can look like there is some duplication of code which is already in the repository; at least I though of the logging functionality. I would hope and think it should be possible to only use the  existing `rml_log` functionality - possible with some mild refactoring. 



```

**Comment by ghost** [comment]

**Comment by joakim-hove** [comment]
ok to test

**Comment by geirev** [comment]
Ok, I have fetched, merged and rebased and the geirev/IES2 branch should now be cleaner.

**Comment by joakim-hove** [comment]
OK good - if you do:
```
git fetch upstream
git rebase upstream/master
git push -f origin IES2
```
it will be *even cleaner*!

**Comment by joakim-hove** [comment]
Very good - thank you.

**Comment by markusdregi** [comment]
Hi Even.

How would you feel about continuing the initial reviewing here at GitHub and then, after that, I can come by your offices one day and we can try to bring it into a mergeable state?

**Comment by geirev** [comment]
I have updated IES2 to be consistent with upstream/master and it now only contains 3 new files and one modified file.   Note there is an issue with duplicate files that I had to delete.
	deleted:    share/ert/forward-models/res/eclipse100
	deleted:    share/ert/forward-models/res/flow
	deleted:    test-data/local/config/simulation_batch/jobs/copy_ext_param
These files exist both in lower and upper case with identical content and that creates case conflicts.

**Comment by joakim-hove** [comment]
@geirev : with this: https://github.com/Statoil/libres/pull/457 there is an `obs_mask` passed to the `analysis_init` function. You must implement that function, internalize the obs_mask pointer - `obs_vector_alloc_copy()` - and then implement the book-keeping of active/inactive observations

**Comment by markusdregi** [comment]
@geirev Just wanted to mention that your changes can easily be deployed in _Equinor_ for testing without the code being merged into the repository for early user testing. I've previously discussed this with Stephanie. Let me know if this becomes relevant.

**Comment by geirev** [comment]
The current version is stable and well tested. Joakim will review code and help create additional tests before changes are merged.

File: lib/analysis/modules/ies_enkf_config.c [hunk 2]
```diff
@@ -0,0 +1,174 @@
...
+  int       ies_inversion;              // Controlled by config key: DEFAULT_IES_INVERSION
+  char    * ies_logfile;           // Controlled by config key: DEFAULT_IES_LOGFILE
+  bool      ies_debug;             // Controlled by config key: DEFAULT_IES_DEBUG
+};
+
+
+ies_enkf_config_type * ies_enkf_config_alloc() {
+  ies_enkf_config_type * config = util_malloc( sizeof * config );
+  UTIL_TYPE_ID_INIT( config , IES_ENKF_CONFIG_TYPE_ID );
+  config->ies_logfile = util_alloc_string_copy( "tmp.log" ); // for DEBUG of alloc vs realloc
```
joakim-hove: You will achieve the same with `config->ies_logfile = NULL`

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+       costfd += matrix_column_column_dot_product(D,i,D,i);
+   }
+   costfp=costfp/ens_size;
+   costfd=costfd/ens_size;
+   costf= costfp + costfd;
+   fprintf(fp," %d %12.5e %12.5e %12.5e %12.5e %12.5e %12.5e %d %d\n", itnr-1, diff1W, diff2W, costfp, costfd, costf, rcond, nrsing , nrobs ) ;
+   fclose(fp);
+}
+
+void tecfld(matrix_type * W,
```
joakim-hove: Some general comments:

1. This function (probably along with many of the others here ...) are clearly only meant to be used *inside* this file; then they should be marked as `static`.
2. The `matrix` argument should not be modified inside this function - specify it as `const matrix_type *W`.

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+   int ic;
+
+   if (izone == 1){
+      fp2 = fopen(fname, "w");
+      fprintf(fp2, "TITLE = \"%s\"\n",fname);
+      fprintf(fp2, "VARIABLES = \"i-index\" \"j-index\" \"%s\"\n",vname);
+      fprintf(fp2, "ZONE F=BLOCK, T=\"Iteration %d\", I=%d, J=%d, K=1\n",izone, ii, jj);
+      ic=0;
+      for (int j=0;j<ii;j++){
+         for (int i=0;i<jj;i++){
```
joakim-hove: It can look like you have mixed up `ii` and `jj` here? Maybe slightly more informative names.

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+            ic=ic+1;
+            fprintf(fp2,"%d ",i);
+            if (ic==30){
+               ic=0;
+               fprintf(fp2,"\n");
+            }
+         }
+      }
+
+      ic=0;
```
joakim-hove: Here you are reusing the variable `ic` for two different (although similar ...) purposes; such variable reuse is in my opinion bad form. I know matters are divided on this, but personally I like to create lexical scopes for situations like this:

```C
{
    int ic;
    for (int i, ...
         for (int j, ... 
}

{
    int ic;
    for (int j, ...
          for (int j, ...
}
```
no big deal for such a simple example as this function - but I think it is a good principle to avoid variable reuse.

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+  free( data );
+}
+
+void teccost(matrix_type * W,
+             matrix_type * D,
+             const char * fname,
+             int ens_size,
+             int izone)
+{
+   FILE *fp2;
```
joakim-hove: No reason to use `fp2` - just use `fp`?

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+  return data;
+}
+
+void ies_enkf_data_free( void * arg ) {
+  ies_enkf_data_type * data = ies_enkf_data_safe_cast( arg );
+  ies_enkf_config_free( data->config );
+  free( data );
+}
+
+void teccost(matrix_type * W,
```
joakim-hove: OK - there seems to be several of these `tecxxx` functions; I would recommend moving to a separate file `tecplot.cpp` or something similar - put the corresponding header file in `libres/lib/private-include/` - we do not want to export `tecplot` functionality.

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+   data->ies_inversion        = ies_enkf_config_get_ies_inversion( data->config );
+   data->ies_logfile          = ies_enkf_config_get_ies_logfile( data->config );
+   data->ies_debug            = ies_enkf_config_get_ies_debug( data->config );
+
+   data->truncation           = ies_enkf_config_get_truncation( data->config );
+   data->subspace_dimension   = ies_enkf_config_get_enkf_subspace_dimension( data->config );
+
+   data->gauss_newton_conv    = ies_enkf_config_get_gauss_newton_conv( data->config );
+
+   if (data->iteration_nr > 3) data->ies_steplength=data->ies_steplength/2;
```
joakim-hove: This seems like a very aggressive shortening of step length - for iteration_nr > 6 the step_length will be reduced with a factor of 1/8 for every step - is that right?

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+   if (data->iteration_nr > 6) data->ies_steplength=data->ies_steplength/4;
+
+
+   FILE *fp;
```
joakim-hove: I have not seen the `freopen()` function before - but it seems to "steal" the `stdout` stream in a highly non-local way; I don't like it :-(

I would prefer if you instead:
```
data->log_fp = fopen(data->ies_logfile, "w/a");
..
fprintf(data->log_fp, ".....", ...);
...
fclose(data->log_fp);
```

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+   }
+
+   fprintf(stdout,"\n\n\n***********************************************************************\n");
+   fprintf(stdout,"IES Iteration   = %d\n", data->iteration_nr);
+   fprintf(stdout,"----ies_steplength  = %f\n", data->ies_steplength);
+   fprintf(stdout,"----ies_inversion   = %d\n", data->ies_inversion);
+   fprintf(stdout,"----ies_debug       = %d\n", data->ies_debug);
+   fprintf(stdout,"----truncation      = %f %d\n", data->truncation, data->subspace_dimension);
+   fprintf(stdout,"----ies_logfile     = %s\n", data->ies_logfile);
+   bool dbg = ies_enkf_config_get_ies_debug( data->config ) ;
```
joakim-hove: Maybe instead of a `bool` for debug it could be a log_level: https://github.com/Statoil/libres/blob/master/lib/include/ert/res_util/log.hpp#L31- then you can do 
```
message_level_type log_level = ies_config_get_ies_log_level( data->config );
if (log_level <= LOG_DEBUG)
     // This is debug message which only comes when you have set the config log level to DEBUG (the lowest value).
     fprintf( ... );

if (log_level <= LOG_CRITICAL)
   // Complete f...up - this should always come ...
```

That way all the nice debug messages you have configured can be retained from some time?

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+
+   if (data->iteration_nr > 3) data->ies_steplength=data->ies_steplength/2;
+   if (data->iteration_nr > 6) data->ies_steplength=data->ies_steplength/4;
+
+
+   FILE *fp;
+   if (data->iteration_nr == 1){
+      fp=freopen(data->ies_logfile, "w", stdout);
+   } else {
+      fp=freopen(data->ies_logfile, "a", stdout);
```
joakim-hove: I would suggest allowing both directory part, and a `%d` - which will be replaced with iteration number - in the filename for the logfile - then this should work:
```C
// Assuming data->ies_logfile = "ies_log/iter%d";
{
     char * filename = util_alloc_sprintf( data->ies_logfile, data->iteration_nr );
     data->log_fp = util_mkdir_fopen( filename , "w");
     free( filename );
}
```

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+    if (strcmp( var_name , IES_LOGFILE_KEY) == 0)
+      ies_enkf_config_set_ies_logfile( module_data->config , value );
+    else
+      name_recognized = false;
+
+    return name_recognized;
+  }
+}
+
+bool ies_enkf_get_string( void * arg , const char * var_name ) {
```
joakim-hove: This looks weird - I guess the return type should be `const char*` - not `bool`?

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+  ies_enkf_data_type * module_data = ies_enkf_data_safe_cast( arg );
+  {
+    if (strcmp( var_name , IES_LOGFILE_KEY) == 0)
+      return ies_enkf_config_get_ies_logfile( module_data->config );
+
+    else
+       return false;
```
joakim-hove: `return NULL;`?

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,1059 @@
...
+   bool_vector_type * obs_mask0;      // Initial observation mask for active measurements
+   bool_vector_type * obs_mask;       // Current observation mask
+   matrix_type * W;                   // Coefficient matrix used to compute Omega = I + W (I -11'/N)/sqrt(N-1)
+   matrix_type * A0;                  // Prior ensemble used in Ei=A0 Omega_i
+   matrix_type * E;                   // Prior ensemble of measurement perturations (should be the same for all iterations)
+   bool      converged;               // GN has converged
+   ies_enkf_config_type * config;     // This I don't understand but I assume I include data from the ies_enkf_config_type defined in ies_enkf_config.c
+   bool      AAprojection;            // For including the AAprojection of Y
+};
+
```
joakim-hove: It seems that the division between the `ies_enkf_config` object and the `enkf_ies_data` struct is not really clear. There are of course no hard rules here - but an approach which I think would make sense (and I *think* that is how it is in the rml module):

**ies_config**: This object is for various settings which can be configured/set by the user. The values in the `ies_config`object should *not* be modified by the algorithm.

**ies_data**: This is (temporary) data needed by the algorithm while running. 

In some cases - the initial/default value can be provided by the `ies_config` object, and then subsequently updated while the algorithm runs. Then the `ies_config` should hold the initial value, and the `ies_data` object hold the running - active - value.

lib/analysis/modules/ies_enkf.c
geirev: Well, I was trying different schemes for the step length. 
I actually wanted to reduce to quarter and not 1/8, but now I think the 1/2 after three steps is ok.
The optimal steplengths are something we will continue to tune with experience from real field cases.

lib/analysis/modules/ies_enkf.c
geirev: teccost is useful since it plots the cost function for all individual realizations as a function of the iteration step. Thus it helps to identify good realizations as well as checking that "most" realizations are reducing the cost function in each iteration step. The output format is just columns of data, but with a two-line text header on top. Maybe we could plot this in ERT instead?

tecfld is only for debugging and will be removed.

teclog is generating a log file that should be accessible to users. Again I added a header for quick plotting in tecplot, but maybe a better alternative would be to plot it directly in ERT.

Lets discuss how to best continue here.

geirev:
I fixed all the issues mentioned above except:
TEC_OUTPUT  - Se comment above.
DEBUG - Will do but too tired now.
DIR/IES_LOGFILE: I support the idea of a separate catalog for the outputs. However, I found it useful to specify the log_filename since it will otherwise be overwritten by the next run. Also, I found it useful to collect the whole run in one file since it is easier to look for differences between iteration steps.

Let's discuss how to proceed.

joakim-hove:
> Let's discuss how to proceed.

Sure.

I have some suggestions in a PR on your PR here [here](https://github.com/geirev/libres/pull/3) - that might be interesting context also for those who will eventually merge this(@markusdregi).

joakim-hove:
@geirev : 

1. Can you rebase this on `upstream/master` again  - some additional changes have been merged.
2. You posted a good list of tests at some stage, but now I can not find it, could you post that again?

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,803 @@
...
+   }
+
+   if (dbg) matrix_pretty_fprint_submat(W0,"Updated W","%11.5f",log_fp,0,m_ens_size,0,m_ens_size) ;
+
+
+
+/* Store active realizations from W0 to data->W */
+   {
+      int i=-1;
+      int j;
```
joakim-hove, line 511: @geirev : I think maybe there is a bug in the code below here; when looking at the code it seems like `dataW` should have dimensions `[ens_size_msk, ens_size_msk]` - but when [allocating `data->W`](https://github.com/equinor/libres/pull/422/files#diff-198bf63d37b336e1e6e5ed266b7ad034R173) the size used is `ens_size` which amount to the currently *active* ensemble size?

geirev: Yes this sounds correct. I never got to test with failed realizations when this bug would hit hard.

joakim-hove: Good - then I understand.

geirev:
**Suggest to verify the PR and that it is merged into Libres.** 

This PR now comprise a complete although comprehensive add-on package IES that is ready for merge.
The PR is nearly a perfect add-on and does not interfere with the existing code.
The PR also comes with an extensive set of tests.

The code compiles into a new library ies.so which is loaded by ert using 
ANALYSIS_LOAD IES_ENKF   /usr/local/lib/ies.so
ANALYSIS_SELECT IES_ENKF

It is expected that, after some further testing and verification in an operational environment, the IES package will become the standard iterative smoother for ensemble history matching.

The further testing will be done by me and Stephanie Agostini in collaboration Eivind Smørgrav and IT, in an Equinor project that has just started.



Variables in IES can be changed from default in the ert menu defined in ert PR IES2 #227
Alternatively they can be defined in the ert datafile as:
ANALYSIS_SET_VAR IES_ENKF ENKF_SUBSPACE_DIMENSION -1
ANALYSIS_SET_VAR IES_ENKF ENKF_TRUNCATION 0.999
ANALYSIS_SET_VAR IES_ENKF IES_STEPLENGTH 0.6
ANALYSIS_SET_VAR IES_ENKF IES_INVERSION 0
ANALYSIS_SET_VAR IES_ENKF IES_DEBUG true
ANALYSIS_SET_VAR IES_ENKF IES_AAPROJECTION false
ANALYSIS_SET_VAR IES_ENKF IES_LOGFILE "testIES0.log"

lib/analysis/modules/tests/ies_iteration.cpp
```diff
@@ -0,0 +1,235 @@
...
+  rng_free(rng);
+}
+
+
+void cmp_std_ies(const res::es_testdata& testdata) {
+  int num_iter = 100;
+  rng_type * rng = rng_alloc( MZRAN, INIT_DEFAULT );
+  matrix_type * A1 = testdata.alloc_state("prior");
+  matrix_type * A2 = testdata.alloc_state("prior");
+  matrix_type * S  = testdata.alloc_state("S");
```
joakim-hove: This has given you a matrix of the same size and shape as A - that is not correct; just use plain matrix_alloc

geirev: yes, lots of new notation in these routines... fixed below.

joakim-hove:
The travis failure is due to failed `apt-get install` - hopefully that is only a temporary glitch; would be good if someone with sufficient credentials can rerun travis.

lars-petter-hauge:
test this please

lars-petter-hauge:
Ah, my bad, you said Travis.. and that has been run..

but apparently it's not running on jenkins..

lars-petter-hauge:
but apparently it's not running on jenkins..

lars-petter-hauge:
ok to test

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,821 @@
...
+  fprintf(log_fp,"%10s: ",name);
+  for (int i = 0; i < bool_vector_size(mask); i++){
+    fprintf(log_fp,"%d", bool_vector_iget(mask, i));
+    if ((i+1)%10  == 0)  fprintf(log_fp,"%s"," ");
+    if ((i+1)%100 == 0)  fprintf(log_fp,"\n %9d  " ,i+1);
+  }
+  fputs("\n", log_fp);
+}
+
+/*
```
markusdregi: Please remove commented out code.

lib/analysis/modules/ies_enkf.c
```diff
@@ -0,0 +1,821 @@
...
+
+
+/*
+ * -------------------------------------------------------------------------------------------------------------
+ * I E n K S
+ * IEnKS update (IES searching the solution in ensemble subspace)
+ * -------------------------------------------------------------------------------------------------------------
+*/
+
+void ies_enkf_updateA( void * module_data,
```
markusdregi, line 112: This function is very long... Would it be possible to split it up? In particular I see that you have blocks of code with a comment above giving it a purpose.

lib/analysis/modules/ies_enkf_config.c
```diff
@@ -0,0 +1,172 @@
...
+    config->subspace_dimension = INVALID_SUBSPACE_DIMENSION;
+}
+
+/*------------------------------------------------------------------------------------------------*/
+/* SUBSPACE_DIMENSION -> TRUNCATION */
+int ies_enkf_config_get_enkf_subspace_dimension( const ies_enkf_config_type * config ) {
+  return config->subspace_dimension;
+}
+
+void ies_enkf_config_set_enkf_subspace_dimension( ies_enkf_config_type * config , int subspace_dimension) {
```
markusdregi, line 102: Are all of these setters necessary? In general immutability is a good thing, especially for configurations. So if we could instead make a constructor taking in all these values and then not change them afterwards that would be much better...

lib/analysis/modules/ies_enkf_config.h
```diff
@@ -0,0 +1,80 @@
+/*
+   Copyright (C) 2015  Statoil ASA, Norway.
```
markusdregi: I think you need to update your copyright headers ;)

lib/analysis/modules/tecplot.c
```diff
@@ -0,0 +1,141 @@
+void static teccost(const matrix_type * W,
+             const matrix_type * D,
+             const char * fname,
+             const int ens_size,
+             const int izone)
+{
+   FILE *fp;
+   float costJ[100];
```
markusdregi: Should this not be the size of the ensemble?

lib/enkf/ensemble_config.cpp
```diff
@@ -812,6 +812,24 @@ bool ensemble_config_has_impl_type(const  ensemble_config_type * config, const e
   return false;
 }
 
+
+bool ensemble_config_GEN_DATA_require_summary(const  ensemble_config_type * config) {
```
markusdregi: This a probably a failed rebase? This function should be removed... I see that the header still exists. That one should of course also be removed (we can fix that in a separate PR).

geirev:
It is still in use (include in ies_enkf.c) but mostly for debugging by myself. The main reason why it is not removed is that it outputs the ensemble of cost function values to file for plotting and this functionality should be extended to plotting within ERT.  The plot below would be very useful for monitoring progress if it is updated after every iteration and displayed within ERT.  I suggest the file is kept for now.
image

markusdregi:
> The plot below would be very useful for monitoring progress if it is updated after every iteration and displayed within ERT.

Could you make an issue in _ERT_ for this?

markusdregi (commented):
Sorry for the long waiting time! To my defense, this is a big PR ;)

When that is said this is truly nice work! It is well tested and the work you have done together with @joakim-hove to make the changes in existing code separate PR's really helped!

Had some minor comments, but we should be able to get this in soon....

lib/enkf/ensemble_config.cpp
markusdregi: The header was removed in https://github.com/equinor/libres/pull/552#pullrequestreview-230228388. So a rebase together with removing this function again should do the trick.

lib/analysis/modules/ies_enkf.c
geirev: Done!

lib/analysis/modules/ies_enkf.c
geirev: It is possible. Would you move code blocks into separate functions defined within the same file, or do you foresee a separate file containing the new functions? 

The reason for not making a collection of functions were many; 1. that there is no reuse of code in the current function or elsewhere; 2. there is a large number of variables that must be communicated for each such function; 3. I was really scared of introducing new bugs in the process;  4. The use of many small functions may reduce the readability of the code, and 5. with the use of the matrix utilities there is already a lot of compaction of the code. 

I will come by your office and we can discuss how to best proceed. For now, I will force push with all the other corrections.

lib/analysis/modules/ies_enkf_config.c
geirev: I think the setters are all used. I agree that the system is a bit messy, but it supports the changing of default values for the variables either through the ERT config file or through the IES submenu in ERT.

lib/analysis/modules/ies_enkf_config.h
geirev: ehh, yes, I guess to  2019 Equinor for now.

lib/analysis/modules/tecplot.c
geirev: correct(ed)

lib/enkf/ensemble_config.cpp
geirev: Yes, I remember this one. It relates to an issue https://github.com/equinor/libres/issues/543.
I have checked out the master file to replace lib/enkf/ensemble_config.cpp, and now it compiles.

markusdregi (approved):
Excellent work! I'll create a follow-up issue for splitting the algorithm into multiple steps.

markusdregi:
Could you squash all the commits into a single one named `New iterative ensemble smoother implementation` or the similar before we merge?

geirev:
Squashed !

Issue #416, opened 2018-10-30
Add forward model job: CAREFULL_COPY


joakim-hove:
This job already exists in ert-statoil. Should only copy if target does not already exist.

markusdregi:
The job in _ert-statoil_ is located here: https://github.com/Statoil/ert-statoil/blob/master/etc/ERT/Scripts/carefull_copy.py

Should go here similarly to the other shell jobs. https://github.com/Statoil/libres/tree/master/share/ert/forward-models/shell

PR #424, opened 2018-11-06
added careful copy job with corresponding test.

ManInFez:
**Task**
#416 

**Approach**

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/python/res/fm/shell/shell.py
+++ b/python/res/fm/shell/shell.py
@@ -212,3 +212,9 @@ class Shell(object):
                 shutil.copyfile( src , target_file )
         else:
             raise IOError("Input argument:'%s' does not correspond to an existing file" % src)
+
+    @staticmethod
+    def carefulCopyFile(src , target = None):
+        if os.path.exists(target):
+            raise IOError("Input argument:'%s' already exist" % target)
+        Shell.copyFile(src,target)
--- a/python/res/fm/shell/shell.py
+++ b/python/res/fm/shell/shell.py
@@ -216,5 +216,6 @@ class Shell(object):
     @staticmethod
     def carefulCopyFile(src , target = None):
         if os.path.exists(target):
-            raise IOError("Input argument:'%s' already exist" % target)
+            print("File: {} already present - not updated".format(target))
+            return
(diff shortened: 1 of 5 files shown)


ghost:

joakim-hove:
Ok to test

python/res/fm/shell/shell.py
```diff
@@ -212,3 +212,9 @@ def copyFile(src , target = None):
                 shutil.copyfile( src , target_file )
         else:
             raise IOError("Input argument:'%s' does not correspond to an existing file" % src)
+
+    @staticmethod
+    def carefulCopyFile(src , target = None):
+        if os.path.exists(target):
+            raise IOError("Input argument:'%s' already exist" % target)
```
joakim-hove: That the target file exists is perfectly OK - not exception.

joakim-hove: That the target file exists is perfectly OK - not exception.

share/ert/forward-models/shell/CAREFUL_COPY_FILE
```diff
@@ -0,0 +1,9 @@
+STDOUT      careful_copy_file.stdout
+STDERR      careful_copy_file.stderr
```
joakim-hove, line 2: The std file descriptors are redirected by default - not necessary to do it explicitly.

joakim-hove (changes requested):
Some minor changes - then she goes in. Good

python/res/fm/shell/shell.py
ManInFez: Should the function fail silently? Would it not be a good idea to let the user know if the copy failed?

joakim-hove: Well - if the file is already present the copy should not proceed, and that is the expected operation of this function - i.e. *not* a failure. Could add a message like:
```
print("File: {} already present - not updated".format(target_file))
```

ManInFez: Very good, I will change it accordingly.

joakim-hove (approved):

PR #433, opened 2018-11-12
Let state_map create mask

markusdregi:
**Task**
Let state map create a mask based on internal data.

Statoil/ert#168


markusdregi:
Pair programmed with @joakim-hove. Merging it..

PR #435, opened 2018-11-13
Bugfixes in enkf_linalg.cpp and matrix.cpp

geirev:
**Task**
Corrected an erroneous index j to i in enkf_linalg_lowrankE.
Also, replace several ugly tabs to spaces.

Then, in matrix.cpp the function matrix_pretty_fprint_submat has a minor bug where it does not print the last indicated column and row.  Also, a \n after the matrix name makes the output even more pretty.

**Approach**

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/lib/analysis/enkf_linalg.cpp
+++ b/lib/analysis/enkf_linalg.cpp
@@ -149,10 +149,10 @@ static int enkf_linalg_num_significant(int num_singular_values , const double * 
     double running_sigma2  = 0;
     for (int i=0; i < num_singular_values; i++) {
       if (running_sigma2 / total_sigma2 < truncation) {  /* Include one more singular value ? */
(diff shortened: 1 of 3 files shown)


ghost:

joakim-hove:
There is a test failure due to the matrix fixup; if you are certain that your approach is correct - please run the test locally and update until it passes before you push a new version of the code.

geirev:
The test was wrong: the returned X from test_initX_subspace_inversion_algorithm and test_initX_enkf_linalg_lowrank_EE should be identical. Now they are.

joakim-hove:
ok to test

joakim-hove (approved):

PR #436, opened 2018-11-13
Value export of JSON updated to create real JSON using c++

### LINKED ISSUE #432: Dump parameters and definitions to runpath as json
Have a look at `value_export` in https://github.com/Statoil/libres/blob/master/lib/enkf/gen_kw.cpp

ManInFez:
included cJSON library

**Task**
resolves #432 

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/lib/enkf/gen_kw.cpp
+++ b/lib/enkf/gen_kw.cpp
@@ -259,14 +259,15 @@ void gen_kw_export_values(const gen_kw_type * gen_kw, value_export_type * export
     char * export_key         = util_alloc_sprintf("%s:%s" , key, parameter);
     double value              = gen_kw_config_transform( gen_kw->config , ikw , gen_kw->data[ikw] );
 
-    value_export_append( export_value, export_key , value );
+    value_export_append( export_value, key, parameter , value );
+
     free( export_key );
 
     if (gen_kw_config_should_use_log_scale(gen_kw->config, ikw)) {
       double log_value = log10(value);
-      char * log_export_key = util_alloc_sprintf("LOG10_%s:%s", key , parameter);
-      value_export_append( export_value, log_export_key , log_value );
-      free( log_export_key );
+      char * log_key = util_alloc_sprintf("LOG10_%s", key);
+      value_export_append( export_value, log_key, parameter , log_value );
+      free( log_key );
     }
   }
 }
--- a/lib/enkf/gen_kw.cpp
+++ b/lib/enkf/gen_kw.cpp
@@ -256,12 +256,10 @@ void gen_kw_export_values(const gen_kw_type * gen_kw, value_export_type * export
   for (int ikw = 0; ikw < size; ++ikw) {
     const char * key          = gen_kw_config_get_key(gen_kw->config);
     const char * parameter    = gen_kw_config_iget_name(gen_kw->config , ikw);
-    char * export_key         = util_alloc_sprintf("%s:%s" , key, parameter);
+
     double value              = gen_kw_config_transform( gen_kw->config , ikw , gen_kw->data[ikw] );
 
     value_export_append( export_value, key, parameter , value );
-
-    free( export_key );
 
     if (gen_kw_config_should_use_log_scale(gen_kw->config, ikw)) {
       double log_value = log10(value);
@@ -273,7 +271,7 @@ void gen_kw_export_values(const gen_kw_type * gen_kw, value_export_type * export
 }
 
 void gen_kw_write_export_file(const gen_kw_type * gen_kw , const char * filename) {
-  value_export_type * export_value = value_export_alloc( NULL, filename );
+  value_export_type * export_value = value_export_alloc("", filename );
   gen_kw_export_values( gen_kw, export_value );
   value_export_txt__( export_value , filename );
   value_export_free( export_value );
--- a/lib/enkf/tests/value_export.cpp
+++ b/lib/enkf/tests/value_export.cpp
@@ -31,7 +31,7 @@
 
 void test_create() {
   test_work_area_type * work_area = test_work_area_alloc("value_export");
-  value_export_type * export_value = value_export_alloc( NULL, "parameters");
+  value_export_type * export_value = value_export_alloc( "", "parameters");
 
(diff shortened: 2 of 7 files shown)


ManInFez:
run manual succeeded

lib/enkf/gen_kw.cpp
```diff
@@ -259,14 +259,15 @@ void gen_kw_export_values(const gen_kw_type * gen_kw, value_export_type * export
     char * export_key         = util_alloc_sprintf("%s:%s" , key, parameter);
```
joakim-hove: The `export_key` variable can be completely removed?

lib/enkf/tests/value_export.cpp
```diff
@@ -47,21 +56,43 @@ test_work_area_type * work_area = test_work_area_alloc("value_export");
...
   value_export_json( export_value );
+
   test_assert_true( util_file_exists( "path/parameters.json" ));
+
+  std::ifstream f("path/parameters.json");
+  std::string strJSON((std::istreambuf_iterator<char>(f)),
+                  std::istreambuf_iterator<char>());
+
+  cJSON *json = cJSON_Parse(strJSON.c_str());
+  test_assert_false(json==NULL);
```
joakim-hove: You can use `test_assert_not_NULL()`

lib/enkf/value_export.cpp
```diff
@@ -46,86 +48,147 @@ static void backup_if_existing(const char * filename) {
...
-  value->base_name = util_alloc_string_copy( base_name );
 
-  value->keys = stringlist_alloc_new( );
-  value->values = double_vector_alloc(0,0);
+  if(directory == NULL)
+    value->directory = "";
+  else
+    value->directory = directory;
+  if(base_name == NULL)
+    value->base_name = "";
```
joakim-hove: We have no intention of handling `base_name == NULL` - that is exception material!

lib/enkf/value_export.cpp
```diff
@@ -18,21 +18,23 @@
...
 #define VALUE_EXPORT_TYPE_ID     5741761
 
 
 struct value_export_struct {
   UTIL_TYPE_ID_DECLARATION;
-  char * directory;
-  char * base_name;
+  std::string directory;
+  std::string base_name;
+  std::map<std::string,std::map<std::string, double>> values;
```
joakim-hove, line 36: In additiom to the map-of-maps `values` I would duplicate the information in a `std::map<std::string,double> composite_values`which you can update with:
```
this->composite_values[ key + ":" + subkey]  = value;
```
in the `export_append()` method. That way the export based on composite keys becomes quite a lot simpler.

lib/enkf/value_export.cpp
```diff
@@ -46,86 +48,147 @@ static void backup_if_existing(const char * filename) {
...
-  backup_if_existing(filename);
-  value_export_txt__( value, filename );
-  free( filename );
+  std::string filename = value->directory +"/" + value->base_name + ".txt";
+  backup_if_existing(filename.c_str());
+  value_export_txt__( value, filename.c_str() );
+
+}
+
+void generate_hirarchical_keys(const value_export_type * value, FILE * stream)
```
joakim-hove: This should be `static` - i.e. visibility should be limited to this file.

lib/enkf/value_export.cpp
```diff
@@ -46,86 +48,147 @@ static void backup_if_existing(const char * filename) {
...
+            fprintf(stream, "\n");
+        }
+
+        fprintf(stream, "},\n");
+
+    }
+
+}
+
+void generate_comosite_keys(const value_export_type * value, FILE * stream)
```
joakim-hove: `static`

lib/enkf/value_export.cpp
```diff
@@ -46,86 +48,147 @@ static void backup_if_existing(const char * filename) {
...
 
-void value_export_append( value_export_type * value, const char * key , double double_value) {
-  stringlist_append_copy( value->keys, key );
-  double_vector_append( value->values, double_value );
-}
 
+void value_export_append(value_export_type * value, const std::string key, const std::string subkey, double double_value){
+
+  if(value->values.find(key) == value->values.end()){
+    value->values.insert(std::make_pair(key, std::map<std::string, double>()));
```
joakim-hove: You can just do: `value->values[key][subkey] = value`. The `[]` operator on a map will auto-create an instance and return a reference to it.

lib/enkf/tests/value_export.cpp
```diff
@@ -78,26 +109,29 @@ void test_export_txt() {
...
   test_assert_true( util_file_exists( "path/parameters__.txt" ));
   test_assert_true( util_files_equal( "path/parameters__.txt", "path/parameters.txt"));
   {
     FILE * stream = util_fopen("path/parameters.txt", "r");
-    char key1[100],key2[100];
+    char key1[100],key2[100], subkey1[100], subkey2[100];
     double v1,v2;
 
-    fscanf( stream, "%s %lg %s %lg" , key1, &v1, key2, &v2);
+    fscanf( stream, "%[^:]:%s %lg %[^:]:%s %lg" , key1, subkey1, &v1, key2, subkey2, &v2);
```
joakim-hove, line 128: That is black magic `fscanf` tricks :-)

lib/enkf/value_export.cpp
ManInFez: Certain tests and calls gives NULL as directory. NULL handling was added to keep original code working. This was acceptable in the old version as it was stored as a pointer. I can change the function calls to use string if that is desired.

lib/enkf/value_export.cpp
ManInFez: I agree it would make output easier, but I think we should aim keep data-structure as simple as possible, rather than  export functions?

joakim-hove: Wel; I consider the export functionality the essential part of this. The implementation of the `struct export_value` is purely a detail - that data structure will never be used for anything else than this particular export.

joakim-hove (approved):
OK - I had some minor comments; nothing important - merge when you see fit.

PR #439, opened 2018-11-13
Take timestep 0 into account

markusdregi:
**Task**
Previously timestep 0 has not been taken into account when loading observed data.


joakim-hove:
Small PR - large impact!

Since this is a potentially important change; can we have a commit message which more clearly communicates the potential impact and importance - e.g.:
```
Start scan at report_step == 0 when looking for active observations.
```

joakim-hove (approved):

PR #440, opened 2018-11-14
WIP: obs_data_active_mask.

stefoss23:
**Task**
obs_vector will return a bool mask of active positions. 

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/lib/enkf/obs_data.cpp
+++ b/lib/enkf/obs_data.cpp
@@ -70,6 +70,7 @@ Matrices: S, D, E and various internal variables.
 #include <ert/util/vector.h>
 #include <ert/res_util/matrix.hpp>
 #include <ert/util/rng.h>
+#include <ert/util/bool_vector.h>
 
 #include <ert/enkf/obs_data.hpp>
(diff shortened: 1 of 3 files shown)


lib/enkf/obs_data.cpp
```diff
@@ -760,3 +771,16 @@ void obs_data_fprintf( const obs_data_type * obs_data , FILE * stream) {
   }
   fprintf(stream , "\n");
 }
+
+
+const bool * obs_data_get_active_mask( const obs_data_type * obs_data ) {
+  int total_size = obs_data_get_total_size( obs_data );
```
joakim-hove, line 781: At the site where this will be used we will need size of mask - so although it is on the way out; I want this to be implemented with `bool_vector_type`.

Furthermore the `obs_data`structure must hold on to the `obs_mask`, and free it when the `obs_data` is discarded. It should not be the responsability of the calling scope to free the mask.

joakim-hove (approved):

PR #443, opened 2018-11-16
Template parser

### LINKED ISSUE #437: Create job for filling in parameter template file
Given a template file of the following format:
```
FILENAME
F1 {{key1.subkey1}}
FX {{key1:subkey2}}
F1_T {{key2.subkey1}}
```
and a Json object like:
```
{
    "key1": {
                "subkey1": 1999.22,
                "subkey2": 200
            },
    "key2": {
                "subkey1": 300
            },
    "key1:subkey1":1999.22,
    "key1:subkey2":200,
    "key2:subkey1":300
}
```

The new job should create a new file by replacing in the key in the template file with the corresponding values from the Json object.

Based on #432

ManInFez:
**Task**
resolves #437

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


.travis.yml
```diff
@@ -79,7 +79,7 @@ install:
     - bash miniconda.sh -b -p $HOME/miniconda
     - export CONDA_HOME="$HOME/miniconda"
     - export PATH="$CONDA_HOME/bin:$PATH"
-    - pip install cwrap
+    - pip install -r requirements.txt
```
joakim-hove, line 82: The family `libecl - libres - ert` are built together on travis, based on this https://github.com/Statoil/ert/blob/master/travis/build_total.py

joakim-hove (commented):
Have not got time now - but I have some requests regarding this review. The loading of parametrs.json should be in a separete (public in res.fm.util) class.

python/res/fm/util/template_render.py
```diff
@@ -0,0 +1,79 @@
...
+
+import jinja2
+import json
+import os
+import yaml
+
+#default export file name found in enkf_defaults.hpp
+DEFAULT_GEN_KW_EXPORT_NAME = "parameters.json"
+
+class TemplateRender:
```
markusdregi: No reason to keep this in a class.. I suggest to remove the class and move the functions out of the `util` submodule and into a `templating` submodule. Then the submodule provides the same "namespacing" as the class did earlier, but in a more pythonic way ;)

python/res/fm/util/template_render.py
```diff
@@ -0,0 +1,79 @@
...
+            except Exception as yaml_err:
+                pass
+
+            try:
+                return json.load(fin)
+            except Exception as json_err:
+                pass
+
+        err_msg = '%s is neither yaml (err_msg=%s) nor json (err_msg=%s)'
+        raise IOError(err_msg % (filename, str(yaml_err), str(json_err)))
```
markusdregi: Please white space separate functions. In general, if a function is inside a class separate them by one line, if outside a class then separate by two newlines..

python/tests/res/fm/test_shell.py
```diff
@@ -68,7 +68,7 @@ def test_symlink(self):
 
             symlink("path/target" , "link")
             self.assertTrue( os.path.islink( "link" ))
-            self.assertTrue( os.path.isfile( "path/target" ))
+            self.   assertTrue( os.path.isfile( "path/target" ))
```
markusdregi: Inserted whitespace by accident?

python/tests/res/fm/test_util.py
```diff
@@ -0,0 +1,73 @@
...
+        with TestAreaContext("util/template") as tac:
+            with open("template", 'w') as template_file:
+                template_file.write(
+                    "FILENAME\n"+
+                    "F1 {{parameters.key1.subkey1}}\n"+
+                    "FX {{parameters.key1.subkey2}}\n"+
+                    "F1_T {{parameters.key2.subkey1}}")
+
+            with open("parameters.json", 'w') as json_file:
+                json_file.write('{\n'+ \
```
markusdregi: I think this would be more readable if specified as a `dict` and then dumped with the `json` library..

python/res/fm/util/template_render.py
```diff
@@ -0,0 +1,79 @@
+gen_kw_export_file = "parameters";
+
+import jinja2
+import json
+import os
+import yaml
+
+#default export file name found in enkf_defaults.hpp
+DEFAULT_GEN_KW_EXPORT_NAME = "parameters.json"
```
markusdregi: Did you consider prototyping the name in `enkf_defaults.hpp`, so that you do not have to specify it twice?

python/tests/res/fm/test_util.py
```diff
@@ -0,0 +1,73 @@
+
+from ecl.util.test import TestAreaContext
+from tests import ResTest
+from res.fm.util import *
+
+import subprocess
+import os
+import sys
+
+class UtilTest(ResTest):
```
markusdregi: The test coverage for the similar job in _Everest_ is quite a bit better. Please revise the cases that you are not currently covering here. In particular having multiple input files and richer template files utilising iteration.

python/res/fm/util/template_render.py
joakim-hove: Whether it is implemented in a class or not is not that important to me; but I would would like the *load dictionary from `parameters.json`* functionality to be in separate, testable+++ unit. That way we can fasilitate a two level approach to creating input files from the `parameters.json` file:

1. The simplest will be plain template instantiation - as provided by this PR.

2. The second will be the ability to create files based on the content of `parameters.json` in an arbitrary way[1]; then people can create forward model executable like this:

```
#!/usr/bin/env python
from res.util import parameters   

value1 = parameters["MAJOR"]["val1"]
value2 = parameters["MAJOR"]["val2"]

if value1 > value2:
    alternative1()
else:
    alternative2()
```
The important point is that we a simple way to get the `parameters.json` loaded into a dictionary structure; from there on people write their own script.



[1]: Yes - I know the jinja2 templates are turing complete - that does not necessary imply that they will always be the most convenient way to make your files.

markusdregi: > [1]: Yes - I know the jinja2 templates are turing complete - that does not necessary imply that they will always be the most convenient way to make your files.

In the same way as most programming languages being turing complete, and that we still have numerous of them. So no need for a discussion here, you are definitively right ;)

markusdregi: > The important point is that we a simple way to get the parameters.json loaded into a dictionary structure; from there on people write their own script.

I agree, but I think we should only make one promise here. Either the API is that we dump the parameters as _json_ to `parameters.json`, which is quite easy to load in all programming languages. Or we say that we can give you the parameters in a data structure in _Python_ and you can use that. Which is slightly easier in _Python_, but less general..? Also, I would in the later case hope that we then expose a function `load_parameters` that they can call?

joakim-hove: > Or we say that we can give you the parameters in a data structure in Python and you can use that. Which is slightly easier in Python, but less general..? Also, I would in the later case hope that we then expose a function load_parameters that they can call?

Hehe - so much for clear communication; I have argued for the second alternative you mention - including to expose well defined functionality to load parameters; to ensure that the `load_parameters()` functionality is published in some way - be it a a class (method) or a function - has been the whole purpose of my commenting in this thread.

markusdregi: Hehe! That is fine by me :) Just not in a class please.. Let us save the dummy classes with no state for the _Java_-developers.

joakim-hove: > with no state for the Java-developers.

That hypothetical class would have state - the name of the file in question; i.e. `parameters.json`.

markusdregi: Yeah, but that is immutable right? So might as well be an member of the submodule? Anyway, I think we agree on this, so lets just see how it turns out :+1:

joakim-hove: >Yeah, but that is immutable right? 

yes - it is immutable; if we fetch it from the compiled code (which we ideally should) - there will be some fencing involved.

> So might as well be an member of the submodule? 

Yes

> Anyway, I think we agree on this, so lets just see how it turns out +1

Yes +1

.travis.yml
ManInFez: Så meine du at det er feil å installere det som er i requirements.txt? Når ein legger til bruk av nye bibliotek så må travis få beskjed på eit vis.

joakim-hove: Feil og feil; men poenget er at for en vilkårlig PR i ett av repoene libecl, libres og ert så bygges alltid *hele kjeden*, basert på ett lite Python script som ligger i ert repoet. Det er derfor det scriptet som må oppdateres: https://github.com/Statoil/ert/blob/master/travis/install_python_packages.py

Så også for en libecl PR må vi få travis til å innstallere `jinja2` - og da hjelper det ikke i legge inn `pip install -r requirements.txt` slik du har gjort.

ManInFez: (I guess I should have kept this in english...) Sorry, but I didn't  understand what you where aiming at. I can make the changes in the install_python_packages script.

joakim-hove: > I can make the changes in the install_python_packages script.

That is exactly what I am aiming at.

andreabrambilla: Sorry in advance for polluting the review of this PR. I find it a bit unfortunate that `build_total.py` uses a dedicated `install_python_packages.py` instead of pip installing the `requirements.txt` from the 3 repos. Was there a particular reason for doing that?

joakim-hove: Don't know - at some time `pip install` was not very efficient/friendly with binary packages like `numpy`. Maybe it could be replaced with three `pip install -r $repo/requirements.txt` calls now.

ManInFez: There are still some things not available in pip, like pyqt4

python/res/fm/templating/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .template_render import render_template
+from .template_render import load_data
+
+def load_parameters():
+    return load_data('parameters.json')
```
markusdregi: Please add a new line to the files with the no-newline symbol displayed here at _Github_ ;)

python/res/fm/templating/template_render.py
```diff
@@ -0,0 +1,76 @@
...
+            raise ValueError('Input file: %s, does not exist..' % input_file)
+
+    if not os.path.isfile(template_file):
+        raise ValueError('Template file: %s, does not exist..' % template_file)
+
+    if not isinstance(output_file, str):
+        raise TypeError('Expected output path to be a string')
+
+
+def render_template(input_files, template_file, output_file):
```
markusdregi, line 69: A doc string would be nice

python/tests/res/fm/test_templating.py
```diff
@@ -0,0 +1,184 @@
...
+            'PROD1 takes value 0.2, implying off\n',
+            'PROD2 takes value 0.4, implying off\n',
+            '----------------------------------\n',
+            'INJ1 takes value 0.8, implying on\n',
+            'INJ2 takes value 0.6, implying on\n',
+            'INJ3 takes value 0.4, implying off\n',
+            'INJ4 takes value 0.2, implying off'
+        ]
+
+
```
markusdregi: Please keep to at most one newline separating code inside a function

python/res/fm/templating/template_render.py
```diff
@@ -0,0 +1,76 @@
...
+
+    if isinstance(input_files, str) and input_files:
+        input_files = [input_files,]
```
markusdregi: When creating a list, it is not necessary to add the comma. Or alternatively you can switch to tuples ;)

python/res/fm/templating/template_render.py
```diff
@@ -0,0 +1,76 @@
...
+
+    all_input_files = [DEFAULT_GEN_KW_EXPORT_NAME+".json",]
+
+    if input_files:
+        all_input_files += input_files
```
markusdregi: If `input_files` are provided as a tuple here, then this will not work. I suggest making `all_input_files` a tuple and then doing `all_input_files += tuple(input_files)`. Then you are ok as long as the user is passing in an iterable collection :+1:

markusdregi (approved):
Had some minor comments regarding white spaces and the API of the templating module. Besides that, this is good to go ;) Also, please clean up your git history in one way or another before it goes in. Good work :+1:

markusdregi:
Ship it :+1:

PR #448, opened 2018-11-21
enkf_obs_scale_correlated_std will return 1 if only 1 obs.

stefoss23:
**Task**

**Approach**

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/lib/enkf/enkf_obs.cpp
+++ b/lib/enkf/enkf_obs.cpp
@@ -1233,6 +1233,9 @@ double enkf_obs_scale_correlated_std(const enkf_obs_type * enkf_obs,
 
   enkf_obs_get_obs_and_measure_data( enkf_obs , fs , local_obsdata , ens_active_list,
                                      meas_data , obs_data );
+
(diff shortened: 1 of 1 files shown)


lib/enkf/enkf_obs.cpp
```diff
@@ -1231,6 +1231,9 @@ double enkf_obs_scale_correlated_std(const enkf_obs_type * enkf_obs,
   obs_data_type * obs_data = obs_data_alloc( 1.0 );
   double scale_factor = 1.0;
 
+  if (local_obsdata_get_size(local_obsdata) == 1)
+    return scale_factor;
+
   enkf_obs_get_obs_and_measure_data( enkf_obs , fs , local_obsdata , ens_active_list,
```
joakim-hove, line 1234: No - the test should be *after* the `enkf_obs_get_obs_and_measure_data()` and should be on the `obs_data` type:

```C++
if (obs_data_get_active_size( obs_data ) <= 1)
   return 1.0;
```

PR #449, opened 2018-11-22
cv_enkf_init_update takes extra arg obs_mask.

stefoss23:
**Task**
cv_enkf_init_update takes extra arg obs_mask.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


joakim-hove:
In addition an update is required here: https://github.com/Statoil/libres/blob/master/lib/enkf/enkf_main.cpp#L1109

stefoss23:
Why?

joakim-hove:
> Why?

Because that function is the frontend to actually call the functions you have changed.

joakim-hove:
This is continued & merged here: https://github.com/Statoil/libres/pull/457

PR #467, opened 2018-11-30
Support for test of FM script consistency.

### LINKED ISSUE #466: Verify that the forward models shipped with libres are internally consistent
In particular - check that the executable confiigured exists as an executable.

xjules:
**Issue**
Resolves #466 

**Approach**
parsing FM scripts for executables and testing whether they exist and are indeed executables. Currently I used only fixed folder containing such scripts - this needs to be double checked.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#

### DIFF
--- a/python/tests/res/fm/test_fm_config.py
+++ b/python/tests/res/fm/test_fm_config.py
@@ -1,0 +1,39 @@
+import os
+import unittest
+from tests import ResTest
+
+class TestFMValidity(ResTest):
+
+    def setUp(self):
+        pass
+
+    def _extract_executable(self,filename):
+        executable_list = []
+        with open(filename,'r') as f:
+            for line in f.readlines():
+                l = [a.strip() for a in line.split(' ') if a not in ['', ' ', '  ']]
+                if l[0]=='EXECUTABLE':
+                    executable_list.append(l[1])
+        return executable_list
+
+    def _file_exist_and_is_executable(self,file_path):
+        return os.path.isfile(file_path) and os.access(file_path, os.X_OK)
+
+    def test_validate_scripts(self):
+        fm_path = ['share/ert/forward-models/res',
+                   'share/ert/forward-models/shell',
+                   'share/ert/forward-models/templating']
+        for fdir in fm_path:
+            fpath = os.path.join(self.SOURCE_ROOT,fdir)
+            files = os.listdir(fpath)
+            for fn in files:
+                fn = os.path.join(fpath,fn)
+                if os.path.isfile(fn):
+                    executable_script = self._extract_executable(os.path.join(fdir,fn))
+                    valid_executable = [self._file_exist_and_is_executable(os.path.join(fpath,es))
+                                        for es in executable_script]
+                    self.assertFalse(False in valid_executable)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/python/tests/res/fm/test_fm_config.py
+++ b/python/tests/res/fm/test_fm_config.py
@@ -7,32 +7,33 @@ class TestFMValidity(ResTest):
     def setUp(self):
         pass
 
-    def _extract_executable(self,filename):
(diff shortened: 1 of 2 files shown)


python/tests/res/fm/test_fm_config.py
```diff
@@ -0,0 +1,39 @@
...
+class TestFMValidity(ResTest):
+
+    def setUp(self):
+        pass
+
+    def _extract_executable(self,filename):
+        executable_list = []
+        with open(filename,'r') as f:
+            for line in f.readlines():
+                l = [a.strip() for a in line.split(' ') if a not in ['', ' ', '  ']]
```
markusdregi: This can be replaced by line.strip().split()

python/tests/res/fm/test_fm_config.py
```diff
@@ -0,0 +1,39 @@
...
+        fm_path = ['share/ert/forward-models/res',
+                   'share/ert/forward-models/shell',
+                   'share/ert/forward-models/templating']
+        for fdir in fm_path:
+            fpath = os.path.join(self.SOURCE_ROOT,fdir)
+            files = os.listdir(fpath)
+            for fn in files:
+                fn = os.path.join(fpath,fn)
+                if os.path.isfile(fn):
+                    executable_script = self._extract_executable(os.path.join(fdir,fn))
```
markusdregi: The length of executable_script should be exactly one.

python/tests/res/fm/test_fm_config.py
```diff
@@ -0,0 +1,39 @@
...
+    def _extract_executable(self,filename):
+        executable_list = []
+        with open(filename,'r') as f:
+            for line in f.readlines():
+                l = [a.strip() for a in line.split(' ') if a not in ['', ' ', '  ']]
+                if l[0]=='EXECUTABLE':
+                    executable_list.append(l[1])
+        return executable_list
+
+    def _file_exist_and_is_executable(self,file_path):
```
markusdregi: Please add a white space between function arguments

python/tests/res/fm/test_fm_config.py
```diff
@@ -0,0 +1,39 @@
...
+        return os.path.isfile(file_path) and os.access(file_path, os.X_OK)
+
+    def test_validate_scripts(self):
+        fm_path = ['share/ert/forward-models/res',
```
markusdregi: I would prefer if this was a listdir of the forward model directory. There are definitely coming more folders here in the future and at some point someone will forget to update this test..

markusdregi (commented):
Please resolve the comments.

python/tests/res/fm/test_fm_config.py
xjules: I touhg about this in the first place, but was not sure about how this will handle irregular spaces, but sure can change it.

python/tests/res/fm/test_fm_config.py
xjules: Ok (here I wasn't sure that such a script could be composed of only one executable)

python/tests/res/fm/test_fm_config.py
xjules: OK

python/tests/res/fm/test_fm_config.py
xjules: Ok, sounds legit

python/tests/res/fm/test_fm_config.py
markusdregi: It will handle tabs and what not. It is the default splitting in _Python_

python/tests/res/fm/test_fm_config.py
markusdregi: It should be exactly one script ;)

markusdregi (approved):

PR #469, opened 2018-12-04
Added DGESVX from lapack.

geirev:
**Issue**
Resolves #my_issue
I needed LU solver that returns condition number to check numerical stability.

**Approach**
Implemented interface to dgesvx (which is an extension of dgesv already in matrix_lapack.cpp)

**Pre un-WIP checklist**
Tested in real application and output is identical to dgesv. 
There are no built-in tests in libres for these routines.
No other ERT routines are currently calling dgesvx.

**Depends on**
* Statoil/libecl#

### DIFF
--- a/lib/res_util/tests/ert_util_matrix_lapack.cpp
+++ b/lib/res_util/tests/ert_util_matrix_lapack.cpp
@@ -82,11 +82,79 @@ void test_det2() {
   rng_free( rng );
 }
 
+void test_dgesvx() {
+  matrix_type * m1 = matrix_alloc(3  , 3 );
+  matrix_type * m2 = matrix_alloc(3  , 3 );
+  matrix_type * b1 = matrix_alloc(3  , 5 );
+  matrix_type * b2 = matrix_alloc(3  , 5 );
+  matrix_type * b3 = matrix_alloc(3  , 5 );
+  rng_type * rng = rng_alloc(MZRAN , INIT_DEV_URANDOM );
+  double rcond;
+  double epsilon=0.0000000001;
+  double diag=1.0;
+  {
+     matrix_diag_set_scalar(m1, diag);
+     matrix_random_init( b1 , rng );
+     matrix_assign(b2 , b1);
+     matrix_dgesv( m1 , b1  );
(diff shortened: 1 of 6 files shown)


ghost:

markusdregi:
test this please

markusdregi:
@geirev Could you please write a small test in `lib/res_util/tests/ert_util_matrix_lapack.cpp` demonstrating that the function works on a small case? Besides that, this is ready to be merged.

geirev:
Ok, I will try :-)

markusdregi:
test this please

markusdregi:
Thanks for the updated version. When Jenkins also have tested, we can merge this.

lib/res_util/tests/ert_util_matrix_lapack.cpp
```diff
@@ -82,11 +82,52 @@ void test_det2() {
...
+     matrix_random_init( m1 , rng );
+     matrix_assign(m2 , m1);
+
+     matrix_random_init( b1 , rng );
+     matrix_assign(b2 , b1);
+
+     matrix_dgesv( m1 , b1  );
+     matrix_dgesvx( m2 , b2, &rcond  );
+
+     test_assert_true( matrix_equal( b1 , b2 ) );
```
markusdregi: This is the assertion failing the test. Does it pass locally?

geirev:
With repeated runs, the original test failed some of the times. The problem is that there is no tolerance in the matrix_equal function (as there is in the python tests) and sometimes the results are different at the level of numerical truncation errors. Thus, the test should compare if || b1 - b2 || < epsilon where epsilon is a number related to the numerical precision and the condition number of the matrix that is inverted. I don't know how to get around this with the existing tests. However, I made a routine, matrix_similar that compares two matrices within a given tolerance, and I think this can also be useful later to test more complex numerical schemes where truncation errors make a difference. I added a test for matrix_similar as well in lib/res_util/tests/ert_util_matrix_lapack.cpp.

markusdregi (approved):
Very good! Thank you for your contribution.

PR #470, opened 2018-12-06
Fix bug when loading predictions (FMU tutorial)

ManInFez:
**Issue**
Resolves https://github.com/Statoil/ert/issues/173

**Approach**

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


everapparatus:
I think we should hold this off until we have some more information. I suggest that we sit down together and continue this work @ManInFez

markusdregi:
I think we should hold this off until we have some more information. I suggest that we sit down together and continue this work @ManInFez

markusdregi (approved):
Good job!

PR #476, opened 2018-12-13
Site config installation

joakim-hove:
**Issue**
When building libres the path to a site wide configuration file is embedded in the compiled code (for good and for bad ...). The path to this file can be given with `-DSITE_CONFIG_FILE=/path/to/som/file`; if that option is not given the path will be `${CMAKE_INSTALL_PREFIX}/share/ert/site-config`.

Prior to this PR the cmake variable `SITE_CONFIG_FILE` was initialized/set the the first time cmake was run, and then *not updated* on subsequent invocations. The consequence of this was that if you reran with an update install prefix, i.e. a sequence like this:
```
bash% cmake .. -DCMAKE_INSTALL_PREFIX=/preifx1 
...
bash% cmake .. -DCMAKE_INSTALL_PREFIX=/prefix2
```
the embedded path would be `/prefix1/share/ert/site-config`.  

**Approach**
Split the handling of the `SITE_CONFIG_FILE` in one CACHE variable which is set by the user - and takes presedence if it is set; and one local variable  `__SITE_CONFIG_FILE` which is reset to the current value of `CMAKE_INSTALL_PREFIX` for every run.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,22 +175,22 @@ if (RST_DOC)
 endif()
 
 foreach(share_dir workflows gui forward-models)
-  install(DIRECTORY "share/ert/${share_dir}" DESTINATION share/ert)
+  install(DIRECTORY "share/ert/${share_dir}" DESTINATION ${SHARE_DIR})
 endforeach()
-install_example("share/ert/site-config" "share/ert")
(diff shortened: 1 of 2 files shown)


markusdregi (approved):
LGTM!

Issue #471, opened 2018-12-10
More active use of flow 


joakim-hove:
#### Use flow
Testing of the intended use of libres, typically with a forward model including reservoir simulation is not trivial because the test environment typically does not have Eclipse installed; to a lesser extent this also applies to developers who do maybe do not have access to Eclipse - or at least must configure it correctly locally. The opm/flow reservoir simulator has matured significantly, and is in my opinion well suited to be used in test and development of libres - in particular there are [binary packages](https://opm-project.org/?page_id=36) which can be used in a CI setup.

If there is a wish to utilize `flow` more actively in libres testing and development I see the following sub tasks:

##### 1. Configure `ecl_config.yml` by poking in the system

The intended task here is to let `cmake` find the `flow` executable and configure the `ecl_config.yml`file correspondingly if it is found. An initial PR for this is: #472

##### 2. Default version

For Eclipse people generally have quite strong feelings about *versions*; i.e. quite often they specifically run exactly version 2017.1 or 2015.2. With that background it makes sense that Eclipse version has a quite *in-your-face* behavior in the user "api", however for `flow` - at least for the simple test and development addressed here - I assume  exact `flow` version will be of less relevance. I therefor suggest some changes/additions in the handling of versions:

1. Add a alias capability to  `ecl_config.py / ecl_config.yml` - so that version `alias` in the example below can be used as an alternative to the `2018.10` version:
```yml
simulators:
  ecl100:
    '2018.10':  
      scalar:
        executable: /path/ecl/2015.2/linux/x86_64/bin/eclipse100
      mpi:
        executable: /path/ecl/2015.2/linux/x86_64/bin/eclipse100_mpi
        ....

    alias: '2018.10'
```
2. Give semantic meaning to e.g. the version string `default` - so that if no version was explicitly requested the `default` version would be used.

3. The code to actually run the eclipse/flow forward model should shuffle the arguments a bit, so that it can be invoked without an explicit version.

markusdregi:
> If there is a wish to utilize flow more actively in libres testing and development I see the following sub tasks.

I think this is a very good initiative!

markusdregi:
For the default versions, should we consider making it three different config files before we continue? One each for `ecl100`, `ecl300` and `flow`?

Also I find a full scale `alias` mechanism a bit overkill, would not a single `default_version` key do the trick?

joakim-hove:
> For the default versions, should we consider making it three different config files before we continue? One each for ecl100, ecl300 and flow?

[...]

PR #477, opened 2018-12-17
Split ecl config

joakim-hove:
This is part of #471.

What has been done:

1. The configuration file for eclipse family simulators has been split in three separate files for ecl100, flow and ecl300.

2. The construction of `EclConfig` is extracted *outside* of `EclRun`, and small simulator object is passed into the `EclRun` class.

3. The `EclConfig` class has a notion of a *default* version.

4. The changes are propagated to the forward model files.

This *should* not introduce any braking changes...


joakim-hove:
OK - now this *works for me* - and it also works for Travis; however the Equinor internal tests are not satisfied. Would be grateful if someone could post a Jenkins error log.

markusdregi:
I just skimmed through the changes and the initial impression is very good 👍  Unfortunately I have not had time the last days to take a closer look... I hope to do that soon!

joakim-hove:
>I just skimmed through the changes and the initial impression is very good

Cool - hopefully the impression will not falter under more scrutiny!

python/res/fm/ecl/ecl_config.py
```diff
@@ -59,68 +59,131 @@ class EclConfig(object):
...
+    def __init__(self, config_file):
         with open(config_file) as f:
             try:
                 config = yaml.load(f)
             except:
                 raise ValueError("Failed parse: {} as yaml".format(config_file))
 
         self._config = config
+        self._config_file = os.path.abspath(config_file)
+        self.default_version = None
```
markusdregi: I think keeping information only once is a good thing, so how about making a property `default_version` that returns `self._config.get('default_version')` instead?

markusdregi: Also, please consider defining the string `default_version` somewhere.

python/res/fm/ecl/ecl_config.py
```diff
@@ -59,68 +71,134 @@ class EclConfig(object):
...
+
+    @property
+    def default_version(self):
+        if Keys.default_version in self._config:
+            return self._config[Keys.default_verson]
+        else:
+            return None
+
+
+    def _get_version(self, version_arg):
```
markusdregi, line 97: Should we raise an exception also if the config does not contain the requested version?

python/res/fm/ecl/ecl_config.py
```diff
@@ -59,68 +71,134 @@ class EclConfig(object):
...
+        if version_arg is None:
```
markusdregi: To me
```python
if version_arg in [None, Keys.default]:
   version = self.default_version
else:
   version = version_arg
```
would read nicer

python/res/fm/ecl/ecl_config.py
```diff
@@ -59,68 +71,134 @@ class EclConfig(object):
...
 
         self._config = config
+        self._config_file = os.path.abspath(config_file)
+
+
+    def __contains__(self, version):
+        if version in self._config[Keys.versions]:
+            return True
+
+        if self.default_version is None:
```
markusdregi: `return self.default_version is not None and version in  [None, Keys.default]` to make the function a bit more compact?

python/res/fm/ecl/ecl_run.py
```diff
@@ -238,6 +183,7 @@ def initRedirect(self):
 
     def execSerialEclipse(self):
         os.execv( self.sim.executable , [ self.sim.executable , self.base_name ])
+        irint("cwd: {}  calling {}  {}".format(os.getcwd(), self.sim.executable, self.base_name))
```
markusdregi: Either this is never tested or the tests will fail...

python/res/fm/ecl/ecl_config.py
joakim-hove: OK - I have added the property. Have also added a small "namespace" class `Keys` which has all the magic strings used.

share/ert/forward-models/res/ECLIPSE100
```diff
@@ -1,2 +1,4 @@
 EXECUTABLE script/ecl100
-ARGLIST <VERSION> <ECLBASE> <NUM_CPU>
\ No newline at end of file
+DEFAULT <VERSION> version
```
markusdregi, line 2: I honestly did not know of this feature! That is very useful...

markusdregi (commented):
This looks very good @joakim-hove :+1: And you are fast at updating!

python/res/fm/ecl/script.py
```diff
@@ -0,0 +1,22 @@
+from argparse import ArgumentParser
+from .ecl_run import EclRun
+
+
+
+
+def run(config, argv):
+    parser = ArgumentParser()
+    parser.add_argument("ecl_case")
+    parser.add_argument("--version", dest="version", type=str)
```
markusdregi, line 10: `default=Keys.default`?

markusdregi:
I've been thinking of whether we should encourage lower case job names for `simulation jobs` and then not bring the older `FORWARD_MODEL` syntax along if we move to `yaml` format. I'm mentioning this as you also created the jobs `flow` etc. If you also think this is a good idea, I would suggest that we remove the `ARGLIST` from the lower cased job descriptions...

joakim-hove:
> I would suggest that we remove the ARGLIST from the lower cased job descriptions...

Done

python/res/fm/ecl/script.py
joakim-hove: I thought we should not expose the `Keys` class - current code will pass `None` for version as default; and that should be picked up as good as the `Keys.default` string. The main purpose of the `Keys.default` string is to support the default in the old style FORWARD_MODEL file.

python/res/fm/ecl/ecl_run.py
joakim-hove: It might actually "work" - because due to the `execv()` call the line should never be executed - test or not. Have removed it.

python/res/fm/ecl/ecl_config.py
joakim-hove: That will raise a `KeyError` in the `_get_sim()` method.

python/res/fm/ecl/ecl_config.py
joakim-hove: OK

python/res/fm/ecl/ecl_run.py
joakim-hove: OK

markusdregi (approved):
Excellent work! When the tests are passing, I'm ready to merge this...

Issue #478, opened 2019-01-07
Clean up status file when restarting ensemble run


ManInFez:
status file is not removed when restarting ensemble until it is over written.

PR #479, opened 2019-01-08
Expose iens in run_arg

ManInFez:
Clear status.json when creating run.json
Improved ForwardModelStatus.load

Required for Statoil/ert#175

### DIFF
--- a/lib/job_queue/forward_model.cpp
+++ b/lib/job_queue/forward_model.cpp
@@ -45,6 +45,7 @@ struct forward_model_struct {
 };
 
 #define DEFAULT_JOB_JSON     "jobs.json"
+#define DEFAULT_STATUS_JSON  "status.json"
 #define DEFAULT_JOB_MODULE   "jobs.py"
 #define DEFAULT_JOBLIST_NAME "jobList"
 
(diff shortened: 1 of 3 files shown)


python/res/enkf/run_arg.py
```diff
@@ -78,3 +79,7 @@ def runpath(self):
     @property
     def iter_id(self):
         return self._get_iter()
+
+    @property
+    def iens(self):
+        return self._get_iens()
```
markusdregi: Missing new line at the end of the file..

markusdregi (dismissed):
Since you are doing two very different things, could you please split the PR into two commits?

lib/job_queue/forward_model.cpp
```diff
@@ -208,6 +209,11 @@ static void forward_model_json_fprintf(const forward_model_type * forward_model,
   fprintf(stream, "}\n");
   fclose(stream);
   free(json_file);
+
+  char * status_file = (char*)util_alloc_filename(path , DEFAULT_STATUS_JSON, NULL);
```
lars-petter-hauge, line 213: I find it a bit awkward that there is removing of a different file within the creation of jobs.json function, but I understand the predicament of having the cleaning placed in the same location as the creation (since the file isn't created before the first job has been run..) - noted in https://github.com/Statoil/libres/issues/478

lars-petter-hauge (approved):
I'll approve the PR regardless of the outcome of the comment. Should be fixed when resolving #478

PR #481, opened 2019-01-11
Facilitate for testing and merging of #422

joakim-hove:
This PR is just a small step to facilitate/enable the merging of #422. Currently it contains two completely independent commits:

1. The `poly_example` case from ert is added
2. The `ert_module` script which is used to build external modules gets some minor cleanup


lib/analysis/modules/ert_module
```diff
@@ -148,19 +148,28 @@ Options summary:
 
   -I/include : Include the path /include in the compiler include path.
 
+<<<<<<< 72d2645279bbc39213c407bb68e98414574fd917:lib/analysis/modules/ert_module
```
markusdregi: Looks like a somewhat broken rebase... ;)

joakim-hove: Yes - fixed

markusdregi (approved):

PR #486, opened 2019-01-17
Fix bug in matrix_fscanf_data()

joakim-hove:
While working to add tests to: #422 a severe bug in the `matrix_fscanf_data()` function was discovered; it was also discovered that a test file: `ert_util_matrix.cpp` was not compiled at all.


markusdregi (approved):
Well spotted!

PR #487, opened 2019-01-18
Assert mask size

joakim-hove:
Another issue discovered while working on adding tests to #422 

The main content of this PR is the hard check added in function `analysis_module_init_update()`, which is basically to protect against possible incorrect use of the API. As part of adding this test some other issues where discovered:

1. A missing argument in the Python `AnalysisModule.initUpdate()` method (first commit).
2. A test file with extension `.c` was renamed `.cpp` and the `CMake` setup for adding tests of the analysis code was slightly refactored (second commit).


markusdregi (approved):
Sanity checks are always welcome :+1:

PR #488, opened 2019-01-18
Use pkill to kill processes spawned by job_dispatch.py

### LINKED ISSUE #417: Local QUEUE; unable to stop/kill jobs
ERT seems to be unable to stop jobs submitted with 'Local QUEUE' . User reported this as a problem for the second time.  The problem is unrelated to the job and has been confirmed with rms and with flow.

To reproduce: submit job with local queuing system, kill the job from GERT and observe that thye process still runs on the machine.

ManInFez:
Resolves #417 

Uses "pkill -P" to kill child processes


xjules (approved):
Tested and explained. Good job!

Issue #238, opened 2018-02-08
A review of the programmatic libres configuration


markusdregi:
**Context**

When implemented, the structure of the dictionary configuring _libres_ was inspired by an early version of the configuration of _Everest_. I've later realized that this configuration was very immature. In addition, we've learned some lessons regarding the configuration of _libres_ lately and I want to put those in use. I hence propose a complete refactoring of the dictionary passed to `ResConfig` to configure _libres_.

**General thoughts**

- I propose that we give ourselves complete freedom in this process. There is no reason to deviate from the current file based configuration just to be different, but where improvements are possible I suggest that we aim at improvement over compatibility.

- If an invalid configuration is passed, we expect the code to say so. In some cases, this is likely to happen in `C`, but in general, we should aim for `Python` exceptions. Later, it would be nice to support some kind of validation of the configuration, but we'll probably save that for later.

- I suggest that we pick a subset of the full space of possible _libres_-configurations and support only those programmatically for now. Then we can expand on this down the line on a on-demand basis.

- Disclaimer: I'm a bit torn on this one! I suggest that we continue to pass the configuration as a Python dictionary for now. I would really like the syntax `config.xxx` instead of `config['xxx']`. In addition I'm a big fan of the immutability of for instance `namedtuples`. However, it is slightly more cumbersome to create something along the lines of a `namedtuple` for the user, no matter if they are to set up a simple configuration of _libres_ in a test, use it in production or load a configuration as `yaml` or similar. If any of you have some thoughts on this one they are very much welcome. But until we have a better solution, I propose sticking with dicts.

- Would it be an idea to expose all the configuration keys in `Python` as lower case versions of the `C` keyword. Are these keys used by anyone else but the configuration part of _libres_, like `ConfigContent`, `ConfigParser` and `ResConfig` (and its ~11 minions)? If not, the only change this would enforce is to lower case all keys in `ConfigKeys.py` and upper case them again when parsing the dictionary in `ResConfig`.

- We should also keep in mind that this is *NOT* an ERT-configuration, but a _libres_ configuration, meaning that if there are settings that are only interpreted by ERT, or possibly should only be a concern for ERT, than I think that is should not en up in the configuration below.

- Also, if there are keys that we plan to deprecate or that is not much used, please point that out and we can leave then out of the loop for now..

**What should be configurable?**
Input on this one is highly appreciated!
- jobname
- job_script
- Installing jobs
- Configuring the forward model (Should we support both `simulation_job` and `forward_model`?)
[...]

PR #491, opened 2019-01-25
Fix arguments in Forward Model jobs

### LINKED ISSUE #490: Arguments ignored in the ECLIPSE100 forward model?
It looks like the arguments for version and num-cpu in the [ECLIPSE100](https://github.com/equinor/libres/blob/master/share/ert/forward-models/res/ECLIPSE100) is ignored when the job is added.
`ARGLIST <ECLBASE> --version=<VERSION> --num-cpu=<NUM_CPU>`

lars-petter-hauge:
**Issue**
Resolves #490

**Approach**
Change how the arguments are input in the forward model job, otherwise they're ignored as comments. This is still a WIP as the other FM jobs must be updated as well, and the `--ignore-errors` isn't handled yet, i.e. it can't be toggled in the config file.

### DIFF
--- a/share/ert/forward-models/res/ECLIPSE300
+++ b/share/ert/forward-models/res/ECLIPSE300
@@ -1,4 +1,4 @@
-EXECUTABLE script/ecl100
+EXECUTABLE script/ecl300
 DEFAULT <VERSION> version
 DEFAULT <NUM_CPU> 1
 ARGLIST <ECLBASE> "--version=<VERSION>" "--num-cpu=<NUM_CPU>" "--ignore-errors"
--- a/share/ert/forward-models/res/FLOW
+++ b/share/ert/forward-models/res/FLOW
@@ -1,4 +1,4 @@
 EXECUTABLE script/flow
 DEFAULT <VERSION> default
 DEFAULT <NUM_CPU> 1
-ARGLIST <ECLBASE> --version=<VERSION> --num-cpu=<NUM_CPU>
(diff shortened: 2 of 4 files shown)


lars-petter-hauge:
Can't say I'm a big fan of how it looks just now, but I do like that it's python argparse that's used in the `run` function.

joakim-hove:
Observe that the comment syntax - i.e. the "--" in this case is a parameter in the c function which parses the job description file, I. E it can be changed

lars-petter-hauge:
Sure, but that would result in every forward model job expecting it to be treated as a comment would throw. Not an issue for internal jobs of course.

There's changes coming towards the [ert config file](https://github.com/equinor/libres/issues/238) itself, perhaps we could evaluate the possibility for these jobs as well to get some deserved attention. Here we could keep in mind the syntax for comments..

markusdregi:
> There's changes coming towards the ert config file itself, perhaps we could evaluate the possibility for these jobs as well to get some deserved attention. Here we could keep in mind the syntax for comments..

I would say that they are a quite natural candidate for `yaml` as well when that train starts to roll... I suggest that we work around the commenting issue until then.

markusdregi:
@lars-petter-hauge Still WIP?

lars-petter-hauge:
retest

lars-petter-hauge:
@markusdregi I've just restarted travis and jenkins now, get those through and I think we're good to go!

markusdregi (approved):
LGTM

PR #492, opened 2019-01-27
Use pointer to const for all the matrices in the analysis update

joakim-hove:
**Issue**
The analysis modules take pointers to five matrices as arguments, these matrices should "in general" not be modified by the analysis modules - however they are passed as pointers to mutable objects, and the main calling scope does not access this memory afterwards, so "some changes" have been safe. For a better communication of intent - and for testing, it would be better to use pointers to const matrices.

**Approach**
The function signatures in the analysis modules have changed from `matrix_type *` to `const matrix_type *`. A few functions actually modified the input argument, for those a local copy is created.


markusdregi (approved):
Good!

PR #497, opened 2019-01-30
Fix condition for obs vector to actual implementation type

### LINKED ISSUE #244: Plot list not correctly tagging GEN_DATA plots with observations
Can be seen in ploy_example

ManInFez:
resolves equinor/ert#244


lars-petter-hauge (approved):
Silly bug.. Python2 really does accept anything.. :)

PR #506, opened 2019-02-06
Update travis with mutltibuild

### LINKED ISSUE #505: Make travis run tests for different python versions on OSX
Update travis similarly to libecl https://github.com/equinor/libecl/issues/571

xjules:
**Issue**
Resolves #505 

**Approach**
We make use of multibuild (https://github.com/matthew-brett/multibuild), which handles different python versions wrt. various OSs well.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libecl#


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "multibuild"]
+	path = multibuild
+	url = https://github.com/matthew-brett/multibuild.git
```
andreabrambilla: As we discussed, it is not worth having this as a submodule, we can simply have a `git clone` in the `.travis` file

xjules: Ok, updated and tested

andreabrambilla (approved):

PR #507, opened 2019-02-06
Adapt to libecl removal of test_work_are_set_store()

joakim-hove:
As part of https://github.com/equinor/libecl/pull/562 the ability set the `store_area` flag of a `test_area` was removed, you had to pass the `store_area` flag when the test_area was created. This PR just follows up on the required API changes in libres.

~~I have added a commit which targets libecl 2.4.rc1 - where this change is; I don't know if that is the way to go?~~

Hmmm: Looking at the [network graph for libecl](https://github.com/equinor/libecl/network) it seems that C++ test_area changes will come in the next version after 2.4? In the final commit I have set the LIBECL_VERSON to `master`- I realize that is not the way it should be; but I had no other idea - please advise.

### DIFF
--- a/.libecl_version
+++ b/.libecl_version
@@ -1,1 +1,1 @@
-export LIBECL_VERSION=2.4.rc1
+export LIBECL_VERSION=2.4.rc2
--- a/lib/enkf/ert_test_context.cpp
+++ b/lib/enkf/ert_test_context.cpp
@@ -44,7 +44,7 @@ struct ert_test_context_struct {
 UTIL_IS_INSTANCE_FUNCTION( ert_test_context , ERT_TEST_CONTEXT_TYPE_ID )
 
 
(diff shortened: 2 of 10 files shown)


lars-petter-hauge:
I've run Jenkins manually with libecl master and can confirm that the tests pass.

yeh, we shouldn't change the .LIBECL version to master.. The better way would be to release a new tag on libecl to build against - we'll get that fixed as well. I understand the challenge though..

.libecl_version
```diff
@@ -1 +1 @@
-export LIBECL_VERSION=2.4.rc1
+export LIBECL_VERSION=master
```
markusdregi: Please set this to `2.4.rc2`

markusdregi:
Please view my comment on which _libecl_ version to use ;) Besides that, this one is good to go in.

.libecl_version
joakim-hove: Done

markusdregi (approved):

PR #508, opened 2019-02-08
Minor changes to prepare for IES2

joakim-hove:
**Issue**
 Attempt to extract changes to existing code from #422 - so that the #422 will - as much as possible - only be completely new code.

### DIFF
--- a/lib/analysis/modules/rml_enkf.c
+++ b/lib/analysis/modules/rml_enkf.c
@@ -620,6 +620,7 @@ void rml_enkf_updateA(void * module_data,
 
 void rml_enkf_init_update(void * arg,
                           const bool_vector_type * ens_mask,
+                          const bool_vector_type * obs_mask,
                           const matrix_type * S,
                           const matrix_type * R,
                           const matrix_type * dObs,
--- a/lib/enkf/enkf_main.cpp
+++ b/lib/enkf/enkf_main.cpp
@@ -1160,6 +1160,9 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
         int * row_offset = (int *)util_calloc( local_dataset_get_size( dataset ) , sizeof * row_offset  );
         local_obsdata_type   * local_obsdata = local_ministep_get_obsdata( ministep );
 
+        // The enkf_main_serialize_dataset() function will query the storage
+        // layer and fetch data which is serialized into the A matrix which is
+        // buried deep into the serialize_info structure.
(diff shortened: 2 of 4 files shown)


joakim-hove:
I have one more minor commit coming.

joakim-hove:
I am trying to reduce the number of moving parts with respect to #422 - would be good to have this merged.

lib/enkf/enkf_main.cpp
```diff
@@ -70,6 +70,7 @@
 #include <ert/res_util/res_util_defaults.hpp>
 #include <ert/res_util/subst_func.hpp>
 #include <ert/res_util/res_log.hpp>
+#include <ert/res_util/es_testdata.hpp>
```
markusdregi: Are there any programmatic changes to this file? If not, I suggest to remove the newly added scope to keep the changes small... Please correct me if I'm wrong!

The improved comments are of course very much welcome 👍

lib/analysis/modules/rml_enkf.c
```diff
@@ -620,6 +620,7 @@ void rml_enkf_updateA(void * module_data,
 
 void rml_enkf_init_update(void * arg,
                           const bool_vector_type * ens_mask,
+                          const bool_vector_type * obs_mask,
```
markusdregi, line 623: I guess I can read this change as this code actually having zero tests?

markusdregi (commented):
This looks good to me!

Please consider to remove the block as pointed out above if that is the single cause of the "big" diff.

lib/analysis/modules/rml_enkf.c
joakim-hove: Not even compiled by default ... :-(

lib/enkf/enkf_main.cpp
joakim-hove: > I suggest to remove the newly added scope to keep the changes small.

Everyone complain about my scopes; when I become dictator then there will be scopes - yessir; but for now I have removed it :-)

joakim-hove:
Give this a Jenkins kick?

lib/analysis/modules/rml_enkf.c
markusdregi: Scary stuff; robots from the future will come for us soon!

lib/enkf/enkf_main.cpp
markusdregi: Scopes went out of fashion in C99... ;)

markusdregi:
Test this please

markusdregi (approved):

PR #512, opened 2019-02-15
Add methods to deactivate in es_testadata

joakim-hove:
This is another small step to enable proper testing of #422. 

In the ensemble smoother in general, and in the iterated ensemble smoother in particular, the book-keeping of realizations and observations which are killed off/deactivated for various reasons is quite complex. With this PR the class `es_testdata` has got the ability to deactivate observations and realizations, to test this functionality.

### DIFF
--- a/lib/res_util/es_testdata.cpp
+++ b/lib/res_util/es_testdata.cpp
@@ -80,7 +80,13 @@ std::array<int,2> load_size() {
   int active_ens_size, active_obs_size;
 
   FILE * stream = fopen("size", "r");
-  fscanf(stream, "%d %d", &active_ens_size, &active_obs_size);
+  if (!stream)
+    throw std::invalid_argument("Could not find file: size with ens_size, obs_size information in the test directory.");
+
+  int read_count = fscanf(stream, "%d %d", &active_ens_size, &active_obs_size);
+  if (read_count != 2)
+    throw std::invalid_argument("Failed to read ens_size obs_size from size file");
+
   fclose(stream);
 
   return {active_ens_size, active_obs_size};
@@ -97,6 +103,12 @@ matrix_type * safe_copy(const matrix_type * m) {
     return matrix_alloc_copy(m);
 
   return nullptr;
+}
+
+
+void matrix_delete_row_column(matrix_type * m1, int row_column) {
+  matrix_delete_row(m1, row_column);
+  matrix_delete_column(m1, row_column);
 }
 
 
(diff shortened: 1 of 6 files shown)


joakim-hove:
This works for me - obvisously; and it works for Travis. If the failure in Equinor/Jenkins is for real it would be very valuable if you could post an error message here.

lib/res_util/tests/es_testdata.cpp
```diff
@@ -49,6 +70,71 @@ res::es_testdata make_testdata(int ens_size, int obs_size) {
...
+  res::es_testdata td1 = make_testdata(ens_size, obs_size);
+  res::es_testdata td2 = make_testdata(ens_size, obs_size);
+
+
+
+  test_assert_throw(td1.deactivate_obs(10), std::invalid_argument);
+  td1.deactivate_obs(del_iobs);
+
+  for (int i1=0; i1 < matrix_get_rows(td1.R); i1++) {
+    int i2 = i1;
```
markusdregi: Would prefer this and the similar construction below to read:
`int i2 = i1 + (i1 >= del_iobs ? 1 : 0);`

(or if you insist `int i2 = i1 + (i1 >= del_jobs);`, but I personally prefer the first one).

lib/res_util/tests/es_testdata.cpp
```diff
@@ -61,9 +147,34 @@ void test_basic() {
...
+  test_assert_int_equal( bool_vector_size(td1.obs_mask), obs_size );
+  test_assert_int_equal( bool_vector_count_equal(td1.obs_mask, true), obs_size );
+
+  test_assert_int_equal( bool_vector_size(td1.ens_mask), ens_size );
+  test_assert_int_equal( bool_vector_count_equal(td1.ens_mask, true), ens_size );
+
   test_work_area_free(work_area);
 }
 
+void test_size_problems() {
```
markusdregi, line 148: This test never runs?

lib/res_util/es_testdata.cpp
```diff
@@ -100,6 +106,51 @@ matrix_type * safe_copy(const matrix_type * m) {
 }
 
 
+matrix_type * matrix_delete_column(const matrix_type * m1, int column) {
```
markusdregi: Is not the `libres` convention that these functions should have `alloc` in their names?

lib/res_util/es_testdata.cpp
```diff
@@ -100,6 +106,51 @@ matrix_type * safe_copy(const matrix_type * m) {
...
+
+matrix_type * matrix_delete_row_column(const matrix_type * m1, int row_column) {
+  matrix_type * m2 = matrix_delete_column(m1, row_column);
+  matrix_type * m3 = matrix_delete_row(m2, row_column);
+  matrix_free(m2);
+  return m3;
+}
+
+
+matrix_type * swap_matrix(matrix_type * old_matrix, matrix_type * new_matrix) {
```
markusdregi: This is not really a `swap`-function, but more of an overwrite function. We should find it a better name...

markusdregi:
The overall content of this PR looks good to me 👍

markusdregi:
How about if we removed the `swap`-function and made `matrix_delete_column(A, row)` equivalent to the current `A = swap(A, matrix_delete_column(S, row)`? Would still float the boat of the implementation without breaking naming conventions...

markusdregi:
Retest this please

markusdregi:
I'll inspect _Jenkins_ if it fails, a bit easier than going back in time trying to find the correct build...

joakim-hove:
> How about if we removed the `swap`-function and made `matrix_delete_column(A, row)` equivalent to the current `A = swap(A, matrix_delete_column(S, row)`? Would still float the boat of the implementation without breaking naming conventions...

I'll look into this; you are absolutely right about the current code breaking the naming convention. It was in an anonymous namespace and I could not really come up with a good name; but anyway - I will look into your suggestion.

lib/res_util/tests/es_testdata.cpp
joakim-hove: Ehhh - well spotted :blush:

lib/res_util/tests/es_testdata.cpp
joakim-hove: Ternary operator - that is beyond my pay grade; but I'll Copy&Paste your suggestion.

joakim-hove:
OK - I think I have addressed all your comments. The new `matrix_delete_row()` and `matrix_delete_column()` are much better now. I would suggest to move them to the `matrix` implementation - but I'll let that decision be up to you.

markusdregi:
> I would suggest to move them to the matrix implementation - but I'll let that decision be up to you.

I don't mind; feel free to do so!

markusdregi:
Retest this please

markusdregi:
The jenkins failure was an error on our side and should be fixed now...

joakim-hove:
OK - this has been completely reshuffeled git wise; but I think all comments have been addressed.

markusdregi (approved):
Looks good :+1:

joakim-hove:
Thank you for merging.

PR #513, opened 2019-02-19
Secure abs calls

ManInFez:
This secures all calls to abs and other math calls. The function abs in c++ does not support double, we need to include cmath in all cpp files. In addition, we should secure other math calls in c++ by specifically referring to std namespace.

The new test values for ExportJoinTest and MisfitCollectorTest has been calculated manually with the following script.
```
from res.enkf.export import GenDataCollector
from res.enkf import ResConfig
from res.enkf import EnKFMain
from ecl.summary import EclSum
from res.enkf.export import SummaryCollector

rc = ResConfig('snake_oil.ert')
main = EnKFMain(rc)

e = EclSum('refcase/SNAKE_OIL_FIELD')
obs = main.getObservations()


def chi(val1, val2, std):
    result = (val1-val2)/std
    return result*result


def chi_from_fopr_obs(realization, key, storage):
    history = key+'H'
    data = e.numpy_vector(key=history, report_only=True)
    res = 0
    for v1, v2, i in zip(storage[key][realization], data, range(0, len(data) - 1)):
        res += chi(v1, v2, max(0.1, 0.1 * abs(v2)))
    return res


def chi_from_obs_key(obskey, realization, sum_storage):
    split_key = obskey.split('_')
    key = split_key[0]+':'+split_key[1]
    time_step = eval(split_key[2])

    std = obs[obskey].getNode(time_step).getStandardDeviation()
    value = obs[obskey].getNode(time_step).getValue()

    return chi(sum_storage[key][realization][time_step - 1], value, std)


def chi_from_sum_obs(realization, sum_storage):
    res = 0
    local_obs = obs.getAllActiveLocalObsdata()
    for x in local_obs:
        key = x.key()
        if key == 'FOPR' or key == 'WPR_DIFF_1':
            continue
        res += chi_from_obs_key(key, realization, sum_storage)
    return res


def chi_from_gen_obs(realization, gen_storage):
    timesteps = [400, 800, 1200, 1800]
    stds = [0.1, 0.2, 0.15, 0.05]
    values = [0.0, 0.1, 0.2, 0.0]

    res = 0
    for timestep, value, std in zip(timesteps, values, stds):
        diff = gen_storage[realization][timestep]
        res += chi(diff, value, std)
    return res


def get_total(realization, sum_storage, gen_storage):
    total = 0
    total += chi_from_fopr_obs(realization, 'FOPR', sum_storage)
    total += chi_from_sum_obs(realization, sum_storage)
    total += chi_from_gen_obs(realization, gen_storage)
    print total


gen_storage_ = GenDataCollector.loadGenData(main, 'default_0', 'SNAKE_OIL_WPR_DIFF', 199)
sum_storage_ = SummaryCollector.loadAllSummaryData(main, 'default_0')
get_total(0, sum_storage_, gen_storage_)
get_total(24, sum_storage_, gen_storage_)

gen_storage_ = GenDataCollector.loadGenData(main, 'default_1', 'SNAKE_OIL_WPR_DIFF', 199)
sum_storage_ = SummaryCollector.loadAllSummaryData(main, 'default_1')
get_total(0, sum_storage_, gen_storage_)
get_total(24, sum_storage_, gen_storage_)

```

### DIFF
--- a/lib/res_util/regression.cpp
+++ b/lib/res_util/regression.cpp
@@ -95,7 +95,7 @@ double regression_unscale(const matrix_type * beta , const matrix_type * X_norm 
    Performs an ordinary least squares estimation of the parameter
    vector beta.
 
-   beta = inv(X'·X)·X'·y
+   beta = inv(X' . X) . X' . y
 */
 
 void regression_augmented_OLS( const matrix_type * X , const matrix_type * Y , const matrix_type* Z, matrix_type * beta) {
--- a/python/tests/res/enkf/export/test_export_join.py
+++ b/python/tests/res/enkf/export/test_export_join.py
@@ -78,12 +78,12 @@ class ExportJoinTest(ResTest):
             self.assertEqual(result["EXTRA_INT_COLUMN"][1][last_date], 225)
             self.assertEqual(result["EXTRA_STRING_COLUMN"][1][last_date], "OFF")
 
-            self.assertFloatEqual(result["MISFIT:FOPR"][0][last_date], 489.191069)
-            self.assertFloatEqual(result["MISFIT:FOPR"][24][last_date], 1841.906872)
+            self.assertFloatEqual(result["MISFIT:FOPR"][0][last_date], 457.491003)
+            self.assertFloatEqual(result["MISFIT:FOPR"][24][last_date], 1630.774198)
 
-            self.assertFloatEqual(result["MISFIT:TOTAL"][0][first_date], 500.170035)
-            self.assertFloatEqual(result["MISFIT:TOTAL"][0][last_date], 500.170035)
-            self.assertFloatEqual(result["MISFIT:TOTAL"][24][last_date], 1925.793865)
+            self.assertFloatEqual(result["MISFIT:TOTAL"][0][first_date], 468.469969)
+            self.assertFloatEqual(result["MISFIT:TOTAL"][0][last_date], 468.469969)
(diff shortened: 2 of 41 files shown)


lib/res_util/regression.cpp
```diff
@@ -95,7 +95,7 @@ double regression_unscale(const matrix_type * beta , const matrix_type * X_norm
    Performs an ordinary least squares estimation of the parameter
    vector beta.
 
-   beta = inv(X'�X)�X'�y
+   beta = inv(X'�X)�X'�y
```
markusdregi: I would say that this is not an improvement ;)

python/tests/res/enkf/export/test_export_join.py
```diff
@@ -54,7 +54,6 @@ def test_join(self):
             gen_kw_data = GenKwCollector.loadAllGenKwData(ert, "default_1")
             misfit = MisfitCollector.loadAllMisfitData(ert, "default_1")
             dm = DesignMatrixReader.loadDesignMatrix("DesignMatrix.txt")
-
```
markusdregi: Please, keep the diff of this one to a minimum.

markusdregi (approved):
Good work! I had to minor comments that should be addressed. In addition, please remove the commits from Joakim and then we can either merge it immediately or alternatively wait for master to be updated. But I can't really see how you would depend on that?

PR #515, opened 2019-02-20
Expose ITER as magic string

markusdregi:
This PR exposes the iteration number as a magic string similarly to how `GEO_ID` and `IENS` already is exposed. This implies that you can pass the iteration number to your forward model, encompass it in paths etc.

Cleaned up the runpath substitution tests while I was at it.


oysteoh (approved):
Great!

PR #517, opened 2019-02-20
Add std:: to isnan() function call

joakim-hove:
This is required to build on my Ubuntu 16.04 / gcc 5.4.0 setup.


markusdregi (approved):
Good initiative!

PR #518, opened 2019-02-21
Osx support

ManInFez:
ValGrind still not supported on osx 10.14, travis should test this on older version


lib/enkf/ert_workflow_list.cpp
```diff
@@ -196,52 +196,9 @@ const workflow_job_type * ert_workflow_list_get_job( const ert_workflow_list_typ
...
 }
 
-/**
-   This function will create the most specific matching filename
-   corresponding to the @root_name input. I.e. it will look for
-   filename in this order:
 
-     ROOT@X.Y.Z, ROOT@X.Y, ROOT@X, ROOT
-
-   And return the first name corresponding to an existing file. If no
```
joakim-hove, line 206: Gee - this was much work at one stage; I would not have heart to kill it myself - but it sure is good to see it go :+1:

lib/enkf/tests/enkf_workflow_job_test_directory.cpp
```diff
@@ -21,58 +21,34 @@
...
   {
     const char * path = argv[1];
     ert_workflow_list_type * workflows = ert_workflow_list_alloc_empty(NULL);
     ert_workflow_list_add_jobs_in_directory( workflows , path );
 
+
     // The CONF1 only exists as default - unversioned
     test_assert_true( ert_workflow_list_has_job( workflows , "CONF1"));
 
     // The CONF2 exists as the default - which is invalid and will not load,
```
oysteoh: guess since the assertions is swapped the comments are outdated and should be changed / removed as well?

oysteoh:
Looks good to me  - but would prefer someone else with more knowledge in libres putting their approval stamp on it ! :smiley:

.travis.yml
```diff
@@ -95,6 +94,7 @@ script:
   - python -c "import sys; print('\n'.join(sys.path))"
   - mkdir build
   - pushd build
+  - ulimit -n 2048
```
andreabrambilla, line 97: Is this required at compile time or when running the tests?
We had issues about "too many files open" when using libres from everest (it was the first usage where a lot of simulations were run on the same ert instance). And the fix was to hunt down the files that were left open. 
If this is required for compiling, it's fine, if it is for running the tests, then maybe we should also open an issue saying that there are still files left open somewhere

lib/CMakeLists.txt
```diff
@@ -431,6 +430,16 @@ foreach(name job_loadOK
      target_link_libraries(${name} res)
 endforeach()
 
+if (NOT APPLE)
```
andreabrambilla, line 433: I would add a comment here saying why this is not run on OSX

lib/CMakeLists.txt
```diff
@@ -620,7 +625,7 @@ foreach(test enkf_magic_string_in_workflows
              enkf_main
              enkf_select_case_job
              enkf_plot_gen_kw_fs
-             enkf_workflow_job_test_version
+             enkf_workflow_job_test_directory
```
andreabrambilla, line 628: This looks suspicious, just double check it is correct (you removed the _version_ thinkgy and added a new directory _thingy_ which are totally unrelated, right?)

.travis.yml
ManInFez: This is for running the some of the tests. The default is 256 on osx and 1024 on linux

lib/enkf/ert_workflow_list.cpp
```diff
@@ -252,38 +209,35 @@ static char * ert_workflow_list_alloc_version_name( const char * path , const ch
...
-        break;
-    }
-    closedir( dirH );
-  } else
+  if (!dirH) {
     fprintf(stderr, "** Warning: failed to open workflow/jobs directory: %s\n", path);
+    return;
+  }
+  while (true) {
+    struct dirent * entry = readdir( dirH );
```
andreabrambilla, line 220: minor: Since we are modernizing the code a bit, I guess this declaration can be changed to just `dirent* entry = readdir( dirH );` or even `auto const* entry = readdir( dirH );`

lib/enkf/ert_workflow_list.cpp
```diff
@@ -252,38 +209,35 @@ static char * ert_workflow_list_alloc_version_name( const char * path , const ch
...
+    if (entry == NULL)
+      break;
+
+    if ((strcmp(entry->d_name , ".") == 0) || (strcmp(entry->d_name , "..") == 0))
+      continue;
 
+    char * root_name = entry->d_name;
```
andreabrambilla, line 227: minor: we don't need the local `root_name`variable any more

lib/enkf/tests/enkf_workflow_job_test_directory.cpp
andreabrambilla: Since the versioning is gone, I guess all the comments should be revised. And maybe also the tests should be revised as well:
- Take for instance the CONF3 case. That workflow exists as `CONF3@1.2.3`. And, as it is now, we check that `CONF3` is not found. It would be more reasonable to check that `CONF3@1.2.3` is found instead.
- versioning is gone, we have removed a variable from the equations. I guess that makes some of the test cases redundant, so we could get rid of them

andreabrambilla (approved):
A few minor comments, overall looks good, nice work :+1:

lib/CMakeLists.txt
ManInFez: The test was renamed, as it is now just testing adding workflow jobs by by directory

PR #520, opened 2019-02-27
Raise userwarning if workflows can't be run

lars-petter-hauge:
**Issue**
If a workflow configuration was invalid, libres would not raise any warning. 

This does not completely solve issue: https://github.com/equinor/ert/issues/194, but it improves.

Handling issue 194 involves a bit more, e.g. if the config file is to include a stdout/stderror like the forward models, then parsing the file must be updated. It remains to find out if that is worth doing in c (as it is now), or if we should implement parsing yml config in python while retaining the old parsing for backwards compatibility.

### DIFF
--- a/python/res/job_queue/workflow.py
+++ b/python/res/job_queue/workflow.py
@@ -60,17 +60,21 @@ class Workflow(BaseCClass):
         """
         self.__running = True
         success = self._try_compile(context)
+        if not success:
+            msg = "** Warning: The workflow file {} is not valid - "\
+                  "make sure the workflow jobs are defined accordingly\n"
+            sys.stderr.write(msg.format(workflow.src_file))
 
-        if success:
-            for job, args in self:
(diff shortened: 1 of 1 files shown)


lars-petter-hauge:
Need to do some more checking before this is merged; e.g. if a workflow raises error in between iterations, do we want complete stop? Believe so, but not completely sure.

lars-petter-hauge:
So, new suggestion for the implementation -> write to `sys.stderror` instead, so similar error handling as if the `runpath_list_file` does not exist.

lars-petter-hauge:
So now I think this is in the proper place. I moved it from `HookManager`, as it's more suited within the workflow itself, closer to where it actually happends. 

Also if one wants to use `Workflow` on it's own it should be able to raise the warning.

ManInFez (approved):
LGTM

PR #522, opened 2019-02-28
Keep stdout stderr

lars-petter-hauge:
**Issue**
Continues the work on https://github.com/equinor/ert/issues/194

Does include commits from #520, but I'll rebase and fix everything.. The #520 will probably be merged first anyway.

**Approach**
First off, use `communicate` from `Popen` when using `PIPE` in order to collect and make sure that the program does not deadlock.

A bit unsure how the error should propagate upwards on this one, and who should be responsible for storing the result. I think I made the proper call on where to put it. 

This isn't complete, using `PIPE` also means that the stdout and stderr isn't sent to terminal, I'll fix that before it's merged so this doesn't change current behaviour

### DIFF
--- a/python/res/job_queue/ert_script.py
+++ b/python/res/job_queue/ert_script.py
@@ -19,6 +19,22 @@ class ErtScript(object):
 
         self.__is_cancelled = False
         self.__failed = False
+        self._stdoutdata = ""
+        self._stderrdata = ""
+
+    @property
+    def stdoutdata(self):
+        """ @rtype: str """
+        if isinstance(self._stdoutdata, bytes):
+            self._stdoutdata = self._stdoutdata.decode()
+        return self._stdoutdata
+
+    @property
+    def stderrdata(self):
+        """ @rtype: str """
+        if isinstance(self._stderrdata, bytes):
+            self._stderrdata = self._stderrdata.decode()
+        return self._stderrdata
 
     def isVerbose(self):
         return self.__verbose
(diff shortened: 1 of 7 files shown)


python/res/job_queue/ert_script.py
```diff
@@ -19,6 +19,28 @@ def __init__(self, ert):
...
+    def stdout(self, value):
+        """ @type: str """
+        self.__stdout = value
+
+    @property
+    def stderr(self):
+        """ @rtype: str """
+        return self.__stdout
+
+    @stderr.setter
```
joakim-hove: Having public read&write properties with names `stdout` and `stderr` is in my opinion a bit misleading, it is not `stdout` and `stderr`as such you are manipulating here after all, but rather the string content which has been written to the respective streams.

lars-petter-hauge: yeh, agreed.

Did a quick test on how it feels when using the existing `defaultStackTrace` in `ErtScript`. That might be the most "how the rest works", but I think `ErtScript` still should keep the results and have it available.

lars-petter-hauge:
Alright, a bit back and forth here.. I've renamed the variables to stdoutdata and stderrdata, the same names as the python documentation has and removed the setters, should be less misleading?

Added some tests as well.

joakim-hove:
> should be less misleading?

I think so yes.

lars-petter-hauge:
Did some final changes to this one.

1) The stdout is now piped to local variable using the `communicate` function from the `Popen` object. The `communicate` will however wait until the command has completed and thus the `stdout` will be flushed completely after the process has completed.. So there's no "live" output from workflows anymore - this might be a problem.

2) The `defaultStackTrace` didn't use the `error` argument at all, I've changed it around a bit, but I guess now it's likely to be not None, and thus it'll always be used. 

3) Instead of returning the stacktrace from the function call (and then having it printed by the caller), it's printed within `ErtScript` and returning `None`

ManInFez (approved):
Good job!

PR #523, opened 2019-03-07
Use c++ TestArea implementation

joakim-hove:
Similar to:  https://github.com/equinor/libecl/pull/583

Two issues:

1. When creating an `ErtTestContext` from Python the underlying test_area is created independently from Python - this implies that the `ert_test_context`implementation is still based on the C api.
2. The utility function `install_file()` was previously removed from the  `TestArea` implementation; the tests where that was used a copy of the function is inlined.

### DIFF
--- a/lib/enkf/tests/enkf_forward_init_FIELD.cpp
+++ b/lib/enkf/tests/enkf_forward_init_FIELD.cpp
@@ -21,7 +21,7 @@
 #include <unistd.h>
 
 #include <ert/util/test_util.h>
-#include <ert/util/test_work_area.h>
+#include <ert/util/test_work_area.hpp>
 #include <ert/util/util.h>
 #include <ert/res_util/arg_pack.hpp>
 #include <ert/util/rng.h>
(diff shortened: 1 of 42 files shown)


lib/enkf/tests/enkf_forward_init_FIELD.cpp
```diff
@@ -46,15 +46,43 @@ void create_runpath(enkf_main_type * enkf_main, int iter) {
...
+
+    if (util_file_exists( src_file.c_str() )) {
+      char * target_file   = util_alloc_filename( ta.test_cwd().c_str(), input_src_file, NULL );
+      util_copy_file( src_file.c_str() , target_file );
+      free( target_file );
+    }
+
+    free(src_path);
+  }
+}
```
andreabrambilla, line 69: I have a feeling this was already discussed somewhere, but why `install_file` is not a method in `TestArea`?

EDIT: there must be some very good reason, since this code is copy pasted several times in this PR

lib/enkf/tests/enkf_forward_init_FIELD.cpp
```diff
@@ -46,15 +46,43 @@ void create_runpath(enkf_main_type * enkf_main, int iter) {
...
 int main(int argc , char ** argv) {
   enkf_main_install_SIGNALS();
   const char * root_path = argv[1];
   const char * config_file = argv[2];
   const char * init_file = argv[3];
   const char * forward_init_string = argv[4];
-  test_work_area_type * work_area = test_work_area_alloc(config_file );
-  test_work_area_copy_directory_content( work_area , root_path );
-  test_work_area_install_file( work_area , init_file );
+  ecl::util::TestArea ta("FIELD");
```
andreabrambilla: not sure about the implication of this, but the value passed to the `TestArea` constructor has changed here

lib/enkf/tests/enkf_gen_data_config_parse.cpp
```diff
@@ -152,7 +149,7 @@ enkf_config_node_type * parse_alloc_GEN_DATA_result( const char * config_string
 
 
 void test_parse_gen_data_result() {
-  test_work_area_type * work_area = test_work_area_alloc("GEN_DATA_RESULT_parse");
+  ecl::util::TestArea ta("GEN:_DATA_RESULT_parse");
```
andreabrambilla: there is a colon in the name, I think some OSs are not too happy with having a colon in a folder name (and if the tests succeeded, maybe it's worth investigating this further)

andreabrambilla (approved):
Overall looks good, but please have a look at my second comment before merging. Maybe it's not serious, but it smells bad

lib/enkf/tests/enkf_forward_init_FIELD.cpp
joakim-hove: Hehe; there was a discussion at some point: https://github.com/equinor/libecl/pull/583 - I lost that argument ...

lib/enkf/tests/enkf_gen_data_config_parse.cpp
joakim-hove: Colon removed in fixup

joakim-hove:
Thank you for review - I have removed the misplaced `:` and pushed an updated version; I can not merge myself ...

PR #524, opened 2019-03-13
Re-internalize data using an instance of run_context

andreabrambilla:
Used to support re-internalization in the everest project

see https://github.com/equinor/everest/issues/762

### DIFF
--- a/lib/enkf/enkf_main.cpp
+++ b/lib/enkf/enkf_main.cpp
@@ -28,6 +28,7 @@
 #include <pwd.h>
 #include <unistd.h>
 #include <sys/types.h>
+#include <thread>
 
 #define HAVE_THREAD_POOL 1
 #include <ert/util/rng.h>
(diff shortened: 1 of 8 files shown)


lib/enkf/enkf_main.cpp
```diff
@@ -2283,61 +2283,81 @@ int enkf_main_load_from_forward_model(enkf_main_type * enkf_main, int iter , boo
...
+      enkf_main_type * enkf_main,
+      ert_run_context_type * run_context,
+      stringlist_type ** realizations_msg_list,
+      enkf_fs_type * fs) {
+   auto const ens_size = enkf_main_get_ensemble_size( enkf_main );
+   auto const * iactive = ert_run_context_get_iactive(run_context);
+
+   int result[ens_size];
+   arg_pack_type ** arg_list = (arg_pack_type **) util_calloc( ens_size , sizeof * arg_list ); // CXX_CAST_ERROR
+   thread_pool_type * tp     = thread_pool_alloc( 4 , true );  /* num_cpu - HARD coded. */
```
ManInFez: What happens if available cpu is less than 4?

joakim-hove: Nothing in particular, performance will of course be the best if the number of workers (i.e. the 4 here) and the number of cpu is the same - but correctness is not affected.

The C++11 function `std::thread::hardware_concurrency()` could probably be used instead of the hardcoded 4?

ManInFez (approved):
Looks good to me!

Good job Andrea :)

PR #526, opened 2019-03-22
Added std-files to fm status

ManInFez:
required for https://github.com/equinor/ert/issues/274


xjules (approved):
Tested and looks legit  - good work!

PR #546, opened 2019-04-12
Create config content in resconfig python constructor

### LINKED ISSUE #531: Construct config_content in python
Move the construction of config_content into the python layer of ResConfig and construct all child-configs with the python config_content

ManInFez:
Initially fetch out ConfigContent into python in order to pass them to the child-configs of ResConfig

resolves #531

### DIFF
--- a/python/res/enkf/res_config.py
+++ b/python/res/enkf/res_config.py
@@ -30,8 +30,6 @@ from res.enkf import (SiteConfig, AnalysisConfig, SubstConfig, ModelConfig, EclC
 class ResConfig(BaseCClass):
     TYPE_NAME = "res_config"
 
-    _alloc_load = ResPrototype("void* res_config_alloc_load(char*)", bind=False)
-    _alloc      = ResPrototype("void* res_config_alloc(config_content)", bind=False)
     _free       = ResPrototype("void res_config_free(res_config)")
     _alloc_full = ResPrototype("void* res_config_alloc_full("
                                "char*, " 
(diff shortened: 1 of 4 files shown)


python/res/enkf/res_config.py
```diff
@@ -76,60 +74,67 @@ def __init__(self, user_config_file=None, config=None, throw_on_error=True):
...
+            plot_config,
+            ecl_config,
+            ensemble_config,
+            model_config,
+            log_config,
+            config_content]
+
+        c_ptr = None
+
+        if not self.errors or not throw_on_error:
```
andreabrambilla: I know you just copied this code from what was there before, bu since we are at it, we could make this if a bit more sensible.
If I understood it correctly, `self.errors` can be populated only by the `build_content` function.
So I would invert the `if` and move it up to line 84. ```
if self.errors and throw_on_error:
  raise ASuitableException('A sensible message')
```

To be fair, it is not entirely clear to me what is the expected behavior when `throw_on_error` is `False`. Does it create an invalid config object? Or?

andreabrambilla (approved):
Just have a look at my comment/question, overall the PR looks good, nice job! :+1:

python/res/enkf/res_config.py
ManInFez: Yeah, I agree it could be moved up there. As to what "not throw on error" means, I am not sure. There is a lot of undefined behavior laying around.

PR #552, opened 2019-04-23
Remove already removed function from header

### LINKED ISSUE #551: Remove header for ensemble_config_GEN_DATA_require_summary
The function `ensemble_config_GEN_DATA_require_summary` was removed in #496, but the header is still around. Should be removed...

ManInFez:
resolves #551


markusdregi (approved):
LGTM

PR #558, opened 2019-04-28
Updated the documentation in the eclxxx_config.yml files

joakim-hove:
Initially all three simulators ecl100, ecl300 and flow were configured in the same configuration file `ecl_config.yml` - but as part of https://github.com/equinor/libres/pull/477 the configuration was split across three files. The documentation has not been updated to reflect this.


markusdregi (approved):
Looks good! Thank you for the documentation update :+1:

PR #565, opened 2019-05-10
Move queue config from site to res config

ManInFez:


ManInFez:
There is a lot of legacy code still in site_config, deprecated and unused keywords like MAX_RUNNING_LSF should be removed from the config parser.

The old tests for queue_config should also be changed to use QUEUE_OPTION more

lib/enkf/tests/enkf_queue_config.cpp
```diff
@@ -23,6 +23,36 @@ void test_empty() {
    queue_config_free(queue_config_copy);
 }
 
+void test_job_script() {
+  ecl::util::TestArea ta("script");
+  {
```
markusdregi: Please remove the scopes and some of the white spaces when you are touching this code anyway ;)

markusdregi: Also, you need to add it to the main body for it to be executed...

python/res/enkf/queue_config.py
```diff
@@ -37,8 +38,23 @@ class QueueConfig(BaseCClass):
     _queue_driver          = ResPrototype("driver_ref queue_config_get_queue_driver(queue_config, char*)")
     _get_num_cpu           = ResPrototype("int queue_config_get_num_cpu(queue_config)")
 
-    def __init__(self, user_config_file=None):
-        c_ptr = self._alloc(user_config_file)
+    def __init__(self, user_config_file=None, config_content=None):
+        if user_config_file and config_content:
```
markusdregi: Be explicit about not being `None`

python/res/enkf/queue_config.py
```diff
@@ -100,3 +116,16 @@ def lsf_server(self):
     @property
     def num_cpu(self):
         return self._get_num_cpu()
+
+    def __eq__(self, other):
```
markusdregi, line 120: Please make sure that all data in the C struct is indeed compared. `num_cpu` is among others missing...

python/tests/res/enkf/test_programmatic_res_config.py
```diff
@@ -336,8 +336,8 @@ def test_minimum_config(self):
             self.assertEqual(loaded_res_config.model_config.getEnspath(),
                              prog_res_config.model_config.getEnspath())
 
-            self.assertEqual(loaded_res_config.site_config.queue_config.job_script,
-                             prog_res_config.site_config.queue_config.job_script)
+            self.assertEqual(loaded_res_config.queue_config.job_script,
```
markusdregi: Is this not superseeded by the equality of the two `queue_configs` below?

python/tests/res/enkf/test_site_config.py
```diff
@@ -32,14 +32,7 @@ def test_invalid_user_config(self):
     def test_init(self):
         with TestAreaContext("site_config_init_test") as work_area:
             work_area.copy_directory(self.case_directory)
-            site_config = SiteConfig()
```
markusdregi: Why are you removing this test although you have not changed the site config constructor?

markusdregi (commented):
Looks quite good :+1: Please address my comments.

python/tests/res/enkf/test_site_config.py
ManInFez: This belongs to the site_config PR, will remove from this one

python/tests/res/enkf/test_programmatic_res_config.py
ManInFez: equality of queue configs in test_large_config

python/res/enkf/queue_config.py
ManInFez: All but the driver are tested for.

python/res/enkf/queue_config.py
```diff
@@ -37,8 +38,23 @@ class QueueConfig(BaseCClass):
     _queue_driver          = ResPrototype("driver_ref queue_config_get_queue_driver(queue_config, char*)")
     _get_num_cpu           = ResPrototype("int queue_config_get_num_cpu(queue_config)")
 
-    def __init__(self, user_config_file=None):
-        c_ptr = self._alloc(user_config_file)
+    def __init__(self, user_config_file=None, config_content=None):
+        if user_config_file is not None and config_content is not None:
+            raise IOError("Trying to create QueueConfig with multiple configurations!")
```
markusdregi: I would expect both of these errors to be `ValueError`. I did also check some of the other configs, and it seems that they (myself 2 years ago) agree :smile:

python/tests/res/enkf/test_res_config.py
```diff
@@ -261,23 +261,23 @@ def assert_analysis_config(self, analysis_config, config_data):
...
+            ('QUEUE_SYSTEM', queue_config.driver.name),
+            ('MAX_RUNNING', queue_config.driver.get_option('MAX_RUNNING')),
         ]
         for key, act in corresponding:
             exp = config_data[key]
             errmsg = 'Error for key {key}, expected: "{exp}", was: "{act}".'
             errmsg = errmsg.format(key=key, act=act, exp=exp)
             self.assertEqual(exp, act, msg=errmsg)
 
+    def assert_site_config(self, site_config, config_data, working_dir):
```
markusdregi, line 280: `umask` should be checked in this assert now

python/res/enkf/queue_config.py
markusdregi: And this behavior should be tested...

markusdregi (approved):
Had some minor comments. Please address them.

Besides that, this looks very good now :+1:

PR #570, opened 2019-05-15
Facilitate extra options res forward models

### LINKED ISSUE #542: Toggle `ignore errors` for ECLIPSE100 not possible
Although some of the arguments handling was added, the `--ignore-errors` was forgotten in https://github.com/equinor/libres/pull/491

Not completely sure how to best fix this, a default `ignore_errors` that's just an empty string in the bash script perhaps.

lars-petter-hauge:
Make it possible for the user to include additional arguments for the
forward models defined in share/ert/res. This is specifically ment for
the user to include --ignore-errors, the only argument not already
handled.

**Issue**
Resolves #542 

**Approach**
Add another magic string...

### DIFF
--- a/share/ert/forward-models/res/ECLIPSE100
+++ b/share/ert/forward-models/res/ECLIPSE100
@@ -1,4 +1,5 @@
 EXECUTABLE script/ecl100
 DEFAULT <VERSION> version
 DEFAULT <NUM_CPU> 1
-ARGLIST <ECLBASE> "--version=<VERSION>" "--num-cpu=<NUM_CPU>" "--ignore-errors"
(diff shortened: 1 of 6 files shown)


lars-petter-hauge:
Give me a second to verify that it works

markusdregi (approved):
LGTM; please backport to 2.5

lars-petter-hauge:
I've added some more info in the docstring to help the users in this case, as I don't think it's obvious how one should include `--ignore-errors`
https://github.com/equinor/ert/pull/370

lars-petter-hauge:
I've checked locally that this works with and without ignoring errors

PR #566, opened 2019-05-14
WIP: Introduce indexed control variables

oysteoh:
Add support for indices on control variables

PR #571, opened 2019-05-15
Support suffixes in libres 

andreabrambilla:
continuation of @oysteoh PR
This is still unfinished and is a continuation of #566


lib/enkf/enkf_config_node.cpp
```diff
@@ -360,36 +360,6 @@ enkf_config_node_type * enkf_config_node_alloc_summary( const char * key , load_
...
-                             output_file,
-                             NULL,
-                             NULL);
-
-    free( output_file );
-
-    return config_node;
-  } else
-    return NULL;
-}
```
andreabrambilla, line 389: This is entirely implemented in python now

lib/enkf/ext_param.cpp
```diff
@@ -77,54 +90,115 @@ double ext_param_key_get( const ext_param_type * param, const char * key) {
...
-    util_abort("%s: invalid index:%d - range: [0,%d) \n",__func__ , index , param->size);
+  if (index >= param->data.size())
+    util_abort("%s: invalid index:%d - range: [0,%d) \n",__func__ , index , param->data.size());
 
   if (index < 0)
-    util_abort("%s: invalid index:%d - range: [0,%d) \n",__func__ , index , param->size);
+    util_abort("%s: invalid index:%d - range: [0,%d) \n",__func__ , index , param->data.size());
 
-  return param->data[index];
+  return param->data[index].front();
```
markusdregi, line 147: I would expect `ext_param_iget` usage on a control with suffixes to be misuse. If you agree, should we raise instead of defaulting to first element for these methods?

andreabrambilla: This was not intended as a default, this is the expected behavior in case the `ext_param` has no suffixes. But you are right, it can be misused.
I'd suggest to raise if this method is called but the `ext_param` has suffixes (and do the same in similar methods)

andreabrambilla: No wait, we are in the C part, we cannot throw exceptions here. We could opt for a `util_abort`, but that is a bit extreme. I would prefer to do those checks only in python before calling this method, and in case raise a `RuntimeError`

markusdregi: We could throw right? Because we are int the `C++` part ;)

I would support raising as you suggested both in the `C++` and Python code. Opting for the `C++` exceptions to never be raised via the Python bindings as long as the bindings do not propagate exceptions (Who knows, maybe pybind will drop by one day).

markusdregi:
This looks very good by the way. I'll make a second go through, but I like what I see 👍

lib/enkf/ext_param.cpp
andreabrambilla: I'm not entirely sure this would work. The implementation is now based on C++, but the API itself is pure C. Making an exception propagate outside of a public method would be rather inconsistent (unless we decide we are officially switching to C++ also in the API)

Also, I guess exceptions don't het along too well with cwrap, don't know what happens if an exception is thrown from a c++ piece of code that was called from python. I'll give it a try anyway and see what happens

markusdregi (approved):
Good work!

PR #573, opened 2019-05-15
Facilitate extra options res forward models

lars-petter-hauge:
Make it possible for the user to include additional arguments for the
forward models defined in share/ert/res. This is specifically ment for
the user to include --ignore-errors, the only argument not already
handled.

The empty default argument is further stripped when running each script.
This will be handled differently when the forward model descriptive
files are rewritten.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-
+dist: xenial
 python:
   - 2.7
   - 3.6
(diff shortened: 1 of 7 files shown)


oysteoh (approved):
lgtm!

lars-petter-hauge:
Dist must be changed in order for the tests on linux to pass. This change has already been implemented as part of a commit in libres master

PR #577, opened 2019-05-21
Update version libres and libecl

lars-petter-hauge:
**Issue**
Update version of libres and libecl dependency

**Depends on**
https://github.com/equinor/libecl/pull/620
(to be fair, it depends on new tag in libecl)


lars-petter-hauge:
Jenkins test this please

lars-petter-hauge:
Jenkins test was restarted internally with the correct branch and all green (even though travis tests should be plenty in this case)

PR #578, opened 2019-05-22
Add dict constructor for queue_config

### LINKED ISSUE #575: Python constructor for queue config
Move constructor logic to python layer

ManInFez:
Create test for dict constructor

Comparison function is not using get_option on driver as local driver does bot have get_option function. 
resolves #575


python/res/enkf/queue_config.py
```diff
@@ -38,25 +39,42 @@ class QueueConfig(BaseCClass):
     _queue_driver          = ResPrototype("driver_ref queue_config_get_queue_driver(queue_config, char*)")
     _get_num_cpu           = ResPrototype("int queue_config_get_num_cpu(queue_config)")
 
-    def __init__(self, user_config_file=None, config_content=None):
-        if user_config_file is not None and config_content is not None:
-            raise ValueError("Trying to create QueueConfig with multiple configurations!")
+    def __init__(self, user_config_file=None, config_content=None, config_dict=None):
```
xjules, line 42: why not to make as a named dict?
E.g., configs = {'user_config_file': filename} and then just ask len(configs)==1

ManInFez: That would be a possibility, but this way, we are not breaking the old functionality.

xjules: Make sense.

xjules (approved):
Looks ok to me!

PR #579, opened 2019-05-22
Fix crash in test ert_util_matrix_lapack and remove memory leaks

mortalisk:
An invalid use of fprintf caused fprintf to sometimes use up all memory
on the system, then crash, and sometimes flood stdout with characters.

The large memory usage happened 80 percent of the time when running a
Debug build. The flooding of characters happened consistently when
running through valgrind. When actually debugging no errors would occur,
making the bug hard to track down.

Fixed by editing format string. The formatting string used a * to
specify that the number of decimal points should be given in a separate
argument, which was not given. It seemed to work most of the time,
except for certain double values (e.g. 0.0).  Fixed it by specifying a
fixed number of decimal points.

Also fixes some memory leaks in both the test and library code.

### DIFF
--- a/lib/res_util/matrix_lapack.cpp
+++ b/lib/res_util/matrix_lapack.cpp
@@ -236,6 +236,7 @@ void matrix_dgesvx(matrix_type * A , matrix_type * B, double * rcond) {
     free(ferr);
     free(berr);
     free(ipivot);
+    free(iwork);
   }
 }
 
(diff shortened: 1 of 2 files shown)


ghost:

markusdregi:
Jenkins, add to whitelist

markusdregi (approved):
Well done :clap:

PR #584, opened 2019-05-31
Add comparison function for ert_workflow_list

### LINKED ISSUE #580: Workflow list python constructor and comparison function
Create python comparison function and dict constructor for workflow list

ManInFez:
Add python dict constructor for ert_workflow_list

Add constructor test for ert_workflow_list

resolves #580

### DIFF
--- a/lib/enkf/ert_workflow_list.cpp
+++ b/lib/enkf/ert_workflow_list.cpp
@@ -53,7 +53,6 @@
 
 struct ert_workflow_list_struct {
   UTIL_TYPE_ID_DECLARATION;
-  stringlist_type         * path_list;
   hash_type               * workflows;
   hash_type               * alias_map;
   workflow_joblist_type   * joblist;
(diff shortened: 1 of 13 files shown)


lib/enkf/ert_workflow_list.cpp
```diff
@@ -67,7 +66,7 @@ static void ert_workflow_list_init(ert_workflow_list_type * workflow_list, const
 ert_workflow_list_type * ert_workflow_list_alloc_empty(const subst_list_type * context) {
   ert_workflow_list_type * workflow_list = (ert_workflow_list_type *)util_malloc( sizeof * workflow_list );
   UTIL_TYPE_ID_INIT( workflow_list , ERT_WORKFLOW_LIST_TYPE_ID );
-  workflow_list->path_list  = stringlist_alloc_new();
+  //workflow_list->path_list  = stringlist_alloc_new();
```
andreabrambilla: remove commented out line

python/res/enkf/ert_workflow_list.py
```diff
@@ -2,34 +2,85 @@
...
+                    print("WARNING: Unable to open job directory {}".format(job_path))
+                    continue
+
+                files = os.listdir(job_path)
+                for file_name in files:
+                    full_path = os.path.join(job_path, file_name)
+                    try:
+                        new_job = WorkflowJob.fromFile(config_file=full_path, parser=parser)
+                        new_job.convertToCReference(None)
+                        workflow_joblist.addJob(new_job)
```
andreabrambilla, line 61: It's just a minor thing, but it would be "safer" to first add the job, then convert it to a C reference. (same is true for lines 48-49 and 67-69)

python/res/job_queue/workflow.py
```diff
@@ -121,4 +121,8 @@ def createCReference(cls, c_pointer, parent=None):
         workflow.__current_job = None
         return workflow
 
+    def __ne__(self, other):
+        return not (self == other)
 
+    def __eq__(self, other):
+        return self.src_file == other.src_file
```
andreabrambilla: I guess this comparison will fail if you haver relative paths, or if you are going through symlinks. Maybe use something like `os.path.realpath()`

andreabrambilla (approved):
Added a few minor comments, overall looks good :+1:

PR #592, opened 2019-06-06
Python constructor for ecl config

### LINKED ISSUE #587: Python constructor and comparison function for ecl_config
Make python constructor and comparison function for ecl_config

ManInFez:
Remove schedule file from ecl config as it is not being read

Add dict constructor for ecl_config

Add non-equinor constructor test for ecl_config

Add ert config file for ecl_config test

resolves #587


lib/enkf/enkf_state.cpp
```diff
@@ -758,19 +758,19 @@ void enkf_state_init_eclipse(const res_config_type * res_config,
...
+  //   util_make_path(schedule_file_target_path);
+  //   free(schedule_file_target_path);
 
-    sched_file_fprintf(ecl_config_get_sched_file(ecl_config), schedule_file_target);
+  //   sched_file_fprintf(ecl_config_get_sched_file(ecl_config), schedule_file_target);
 
-    free(schedule_file_target);
-  }
+  //   free(schedule_file_target);
+  // }
```
xjules: Is there a reason to left it commented?

ManInFez: nope just missed it.. will clean up

python/tests/res/enkf/test_ecl_config.py
```diff
@@ -121,4 +120,24 @@ def test_refcase( self ):
         refcaseName = ec.getRefcaseName() + ".DATA"
         self.assertEqual( dfile , refcaseName )
 
+    def test_ecl_config_constructor(self):
+        config_dict = {
```
xjules, line 104: Please check INIT_SECTION_KEY since it parsed in ecl_config

ManInFez: After discussing with Markus we decided on marking this keyword as deprecated and not include in the tests.

xjules: ok

lib/enkf/config_keys.cpp
```diff
@@ -146,14 +142,49 @@ const char * config_keys_get_enspath_key() {
   return ENSPATH_KEY;
 }
 
+
+/* ************* ECL config  ************* */
```
xjules, line 146: Shouldn't this be gone?

python/res/enkf/config_keys.py
```diff
@@ -141,16 +155,26 @@ class ConfigKeys:
...
     DATA_KW_KEY      = _data_kw_key()
     RUNPATH          = _runpath()
     RUNPATH_FILE     = _runpath_file()
     RUNPATH_LIST_FILE = ".ert_runpath_list"
+    
+    # ************* ECL config  *************
     ECLBASE          = _eclbase()
-    NUM_REALIZATIONS = _num_realizations()
-    ENSPATH          = _enspath()
+    SCHEDULE_FILE    = _schedule_file()
```
xjules: `_schedule_file ` should be gone too?

xjules (approved):
Hopefully all the small things were collected. Good to go!

xjules:
I checked this config and `ecl_config.ert` does not have the same values as the dict in `test_ecl_config.py` - I mean the paths.

PR #607, opened 2019-06-24
Switch to Jenkins Pipelines

mortalisk:


testjenkins.sh
```diff
@@ -0,0 +1,177 @@
...
+clone_repos () {
+	echo "Cloning into $LIBRES_ROOT"
+	git clone . $LIBRES_ROOT
+
+	source ./.libecl_version
+
+	echo "Cloning into $LIBECL_ROOT"
+	git clone https://github.com/equinor/libecl $LIBECL_ROOT
+	pushd $LIBECL_ROOT
+	git checkout tags/$LIBECL_VERSION
```
oyvindeide: Can consider just checkout out the specific tag, for example like this:
git clone -b $LIBECL_VERSION --single-branch https://github.com/equinor/libecl $LIBECL_ROOT

testjenkins.sh
```diff
@@ -0,0 +1,177 @@
...
+
+create_directories () {
+	mkdir $INSTALL
+	mkdir $LIBECL_BUILD
+	mkdir $LIBRES_BUILD
+}
+
+clone_repos () {
+	echo "Cloning into $LIBRES_ROOT"
+	git clone . $LIBRES_ROOT
```
oyvindeide, line 121: Will this disregard local changes? Shouldn't be relevant for Jenkins though.

mortalisk: Yes. I will add a comment to clarify that in the top of the script

python/res/job_queue/ext_job.py
```diff
@@ -76,7 +76,10 @@ def __init__(self, config_file, private, name = None , license_root_path = None
             raise IOError('No such config file "%s".' % config_file)
 
     def __repr__(self):
-        return self._create_repr('%s, config_file = %s' % (self.name() , self.get_config_file()))
+        if self._address():
+            return self._create_repr('%s, config_file = %s' % (self.name() , self.get_config_file()))
```
oyvindeide: Since you're already changing this, you could change to .format instead of %, but not very important.

oyvindeide (approved):
Could perhaps add a more extensive commit message, but otherwise very good :+1:

PR #612, opened 2019-06-26
Enable pytest to run from source/python instead of the build folder

oyvindeide:
Because several tests were dependent on compiled values, it was not
possible to run python tests from the source folder. Removing the
copy of python test files to the build folder should also speed
up the build process. The test checking that all test files were
copied to the build folder is no longer valid. Because of
`python/__init__.py` the tests were not testing the source python
, but the installed version.


oyvindeide:
Note that jenkins is currently expected to fail as it tries to run the tests from the build folder, where there are no tests.

oyvindeide:
Another thing to note is that previously, there was internal logic to search for the libraries, first LD_LIBRARY_PATH was searched, then the build folder was searched. That second option has been removed and the libraries have to be on the LD_LIBRARY_PATH, this was because it would default to the LD_LIBRARY_PATH even if a secondary path was specified (the search order was reversed in https://github.com/equinor/cwrap/commit/f000d05b76ec3d1bd50f849128e89e5c838a4be7). This was done to avoid creating potentially confusing behavior.

oyvindeide:
retest this please

python/res/enkf/site_config.py
```diff
@@ -147,7 +147,10 @@ def __eq__(self, other):
         self_job_list = self.get_installed_jobs()
         other_job_list = other.get_installed_jobs()
 
-        if other_job_list.getAvailableJobNames() != self_job_list.getAvailableJobNames():
+        if len(other_job_list.getAvailableJobNames()) != len(self_job_list.getAvailableJobNames()):
+            return False
+
+        if not all(x in other_job_list.getAvailableJobNames() for x in self_job_list.getAvailableJobNames()):
```
andreabrambilla: This does not check for equality, it checks if `self_job_list` is a subset of `other_job_list`. What about checking for `set(other_job_list.getAvailableJobNames()) != set(self_job_list.getAvailableJobNames())`

andreabrambilla (approved):
Looks good :+1:

PR #617, opened 2019-06-28
Include equinor tests in test runner and change to pytest mark

oyvindeide:
Equinor tests were not being run in because of a broken symlink,
yet they were marked as passed in ctest. Changed to pytest.mark
to simplify the implementation.

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -1,0 +1,32 @@
+import pytest
+import unittest
+import resource
+import functools
+import os
+
+def source_root():
+    path_list = os.path.dirname(os.path.abspath(__file__)).split("/")
+    while len(path_list) > 0:
+        git_path = os.path.join(os.sep, "/".join(path_list), ".git")
+        if os.path.isdir(git_path):
+            return os.path.join(os.sep, *path_list)
+        path_list.pop()
+    raise RuntimeError('Cannot find the source folder')
+
+
+def has_equinor_test_data():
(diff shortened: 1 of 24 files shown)


xjules (approved):
Tested and worked - good job!

python/tests/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+    if has_equinor_test_data() and not all(x > 2000 for x in resource.getrlimit(resource.RLIMIT_NOFILE)):
+        new_limit = 2048
+        print("Running Equinor tests, and maximum open file handles is: {}, increasing to: {}".format(
+            min(resource.getrlimit(resource.RLIMIT_NOFILE)), new_limit))
+        resource.setrlimit(resource.RLIMIT_NOFILE, (new_limit, new_limit))
+
+
+def pytest_runtest_setup(item):
+    if item.get_closest_marker("equinor_test") and not has_equinor_test_data():
+        pytest.skip("Test requires Equinor data")
```
andreabrambilla, line 31: Just a minor thing. Why not replacing this method with something like
```Python
equinor_test = pytest.mark.skipif(lambda: not has_equinor_test_data())
```

This way it is explicit that we are creating a new marker/decorator

oyvindeide: Considered doing something like that, the reason I chose mark was if we wanted the flexibility to just run the Equinor tests with pytest -m "equinor_test", but if that is less readable I'll be happy to change to your suggestion.

PR #619, opened 2019-07-01
Correct doc-string and remove ERT_LIBRARY_PATH from res/__init__.py

oyvindeide:
Because cwrap now first reads LD_LIBRARY_PATH, https://github.com/equinor/cwrap/commit/f000d05b76ec3d1bd50f849128e89e5c838a4be7 the build path and the ERT_LIBRARY_PATH would only be used if no library was found on LD_LIBRARY_PATH, which could cause confusion to which library was actually being used. The docstring was thus incorrect. Changed this to only read LD_LIBRARY_PATH.

### DIFF
--- a/python/res/__init__.py
+++ b/python/res/__init__.py
@@ -27,28 +27,6 @@ function uses the standard methods of the operating system,
 i.e. standard locations configured with ld.so.conf and the environment
 variable LD_LIBRARY_PATH.
 
-To avoid conflict with other application using the res libraries the
-Python code should be able to locate the shared libraries without
-(necessarily) using the LD_LIBRARY_PATH variable. The default
-behaviour is to try to load from the library ../../lib64, but by using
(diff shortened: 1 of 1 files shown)


andreabrambilla:
Run this on top of my libecl and libres PRs for OSX ( equinor/libecl#640 and #624  ), and everything works fine. As soon as those PRs are in we expect this PR to pass on travis

markusdregi:
Please rebase and make ready for review again when tests are passing

lars-petter-hauge:
The changes are good, we want to respect `OS`'s installation path and behave as expected.

But the commit title is a bit misleading, right? You're not removing `res_lib_path` per se, you're just not modifying it based on the environment variable `ERT_LIBRARY_PATH` anymore.

lars-petter-hauge (approved):

PR #624, opened 2019-07-16
Build and install libres on macosx using rpath

andreabrambilla:
Analogous to Equinor/libecl#640

**Depends on**
* Equinor/libecl#640


markusdregi (approved):
LGTM

PR #625, opened 2019-07-19
Backup existing parameters.XXX files with date before writing

### LINKED ISSUE #391: parameters.txt.old ->parameters.txt.old_ss.MM.hh.dd.mm.yyyy or similar
Add timestamp to parameters.txt.old (eg parameters.txt.old.ss.mm.hh.dd.mm.yyy or something similar) when renamed so that it becomes unique and not lost the second time it is overwritten.

Requested from user at ERT customer voice meeting.

andreabrambilla:
**Issue**
Resolves https://github.com/equinor/ert/issues/391

### DIFF
--- a/lib/enkf/tests/value_export.cpp
+++ b/lib/enkf/tests/value_export.cpp
@@ -19,14 +19,63 @@
 #include <stdbool.h>
 #include <stdio.h>
 
+#include <algorithm>
 #include <iostream>
 #include <fstream>
+#include <string>
+#include <vector>
 
 #include <ext/json/cJSON.h>
 
+#include <ert/util/util.h>
 #include <ert/util/test_util.h>
 #include <ert/util/test_work_area.hpp>
 #include <ert/enkf/value_export.hpp>
+
+
+namespace {
+
+// This is super ugly but I could not find any way to do this without using a
+// a global variable. Of course the correct thing to do would be to implement
+// a new walk_directory that either collects first and return (include
+// max_depth, remove callbacks), or behaves like an iterator. But that would
+// require quite some efforts, and it's probably not worth it for just one
+// test. Especially considering that C++17 includes the filesystem library that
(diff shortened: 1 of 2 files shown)


ManInFez (approved):
Ship it!!

lib/enkf/tests/value_export.cpp
```diff
@@ -19,16 +19,65 @@
...
 
 
+namespace {
+
+// This is super ugly but I could not find any way to do this without using a
+// a global variable. Of course the correct thing to do would be to implement
+// a new walk_directory that either collects first and return (include
+// max_depth, remove callbacks), or behaves like an iterator. But that would
+// require quite some efforts, and it's probably not worth it for just one
+// test. Especially considering that C++17 includes the filesystem library that
```
markusdregi, line 43: Considering that `libres` is now moving towards having a Python-only API; I would say that we can use whatever compiler floats our boat ;)

lib/enkf/tests/value_export.cpp
```diff
@@ -132,6 +203,28 @@ void test_export_txt() {
...
+      std::begin(tree),
+      std::end(tree),
+      [](std::string const& l, std::string const& r) -> bool {
+          return l.size() < r.size();
+      });
+
+  test_assert_string_equal(tree[0].c_str(), "path/parameters.txt");
+  test_assert_string_equal(
+      tree[2].substr(0, 29).c_str(),
+      "path/parameters.txt_backup_20"); // Fix this in 80 years
```
markusdregi, line 224: Already looking forward to it 😎

Would it be possible to have a slightly more exact test of this?

andreabrambilla: given how painful it is to deal with time and date in c++, I'd rather wait until we move everything to python (which hopefully will happen before this test start failing :) )

PR #626, opened 2019-07-19
WIP: Remove outdated version of Eclipse/Flow jobs

andreabrambilla:
In `share/ert/forward-models/res` there are 2 versions of the jobs `eclipse100` `eclipse300` and `flow`: a version with the name all lowercase, another version with the name all uppercase
The two versions are similar but not identical, the uppercase version seems to be more up-to-date, so we remove the lowecase version.

@joakim-hove @markusdregi you guys might offer some insight on this. We tried digging in the git log but could not figure out why there are 2 versions of these jobs. If there is actually a good reason to keep them both, please let us know (FYI, they make a pretty bad mess of OSs that do not distinguish between uppercase and lowercase)

### DIFF
--- a/share/ert/forward-models/res/eclipse100
+++ b/share/ert/forward-models/res/eclipse100
@@ -1,1 +1,0 @@
-EXECUTABLE script/ecl100
--- a/share/ert/forward-models/res/eclipse300
+++ b/share/ert/forward-models/res/eclipse300
@@ -1,1 +1,0 @@
-EXECUTABLE script/ecl300
--- a/share/ert/forward-models/res/flow
+++ b/share/ert/forward-models/res/flow
@@ -1,1 +1,0 @@
-EXECUTABLE script/flow


markusdregi:
@andreabrambilla The idea was to use lower case for simulation jobs/future yaml configuration and keep the all caps around for compatibility. Your comment regarding OSs not distinguishing between upper and lowercase is a good point!

andreabrambilla:
I see the point. Since there is no distinction at the moment (well, the two versions are actually out of sync, but I guess they shouldn't, at least for now), what about having only one version in the repo, and automatically create the uppercase and lowercase versions on install?

markusdregi:
I have no problem with that! As you say; they should probably be equal now, but might deviate later down the road.

andreabrambilla:
Closing this, PR #645 takes care of those jobs

PR #633, opened 2019-08-01
Add support for unknown status when lsf commands are unavailable

### LINKED ISSUE #386: Investigate lsf failure handling
Users have reported issues when the lsf system and/or disk goes down temporarily. In particular, the `lsf` commands are not available. We could mock this scenario and test how ERT behaves to try to increase the robustness.

We should also investigate how killing jobs behave in this scenario.

ManInFez:
**Issue**
Resolves equinor/ert#386

In order to prevent the queue to resubmit already submitted jobs after reconnecting with lsf, unknown is treated as a separate status.


lib/job_queue/job_queue_status.cpp
```diff
@@ -76,6 +76,7 @@ job_queue_status_type * job_queue_status_alloc() {
   status->status_index[12] = JOB_QUEUE_STATUS_FAILURE; //The driver call to get status has failed, job status remains unchanged
   status->status_index[13] = JOB_QUEUE_FAILED; // Job has failed, no more retries, FINAL STATE
   status->status_index[14] = JOB_QUEUE_DO_KILL_NODE_FAILURE; // Job has failed, node should be blacklisted
+  status->status_index[15] = JOB_QUEUE_UNKNOWN; // Job has failed, node should be blacklisted
```
markusdregi: Change the comment?

ManInFez: Never, it was definitely intentional..... for reasons...

markusdregi (approved):
Looks good to me :+1: Well done!

We should add some tests verifying the robustness of the queue system. Could you make an issue on the matter?

andreabrambilla:
At the beginning of `job_status.hpp` file there is a fancy ascii drawing describing the possible status changes for a job. It would be nice to add the new status to the drawing if possible (or at least have the new status mentioned somewhere nearby so that the documentation stays up-to-date)

PR #641, opened 2019-08-08
Exit signal condition

markusdregi:
**Issue**
Misusage of `WEXITSTATUS`. From documentation:
```
If WIFEXITED(status) is true, return the integer parameter to the exit(2) system call. Otherwise, the return value is meaningless.
```

**Approach**
Use `WIFEXITED` as a condition to apply `WEXITSTATUS`

### DIFF
--- a/python/tests/res/job_queue/test_jobmanager.py
+++ b/python/tests/res/job_queue/test_jobmanager.py
@@ -2,8 +2,10 @@ import json
 import os
 import os.path
 import stat
+import subprocess
 import time
 import datetime
+import unittest
 from unittest import TestCase
 
 from ecl.util.test import TestAreaContext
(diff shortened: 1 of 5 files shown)


mortalisk:
I propose that we simply drop the call to WEXITSTATUS all together.

As it stands now, a reader of the log is likely to miss the distinction between whether this was a signal or a return code.

If that is not acceptable we probably should distinguish in the error logging between the situation where the program terminated itself, and when it was killed with a signal.

Edit: Apparently, the waitpid function returns a 16 bit number where the high bits are the exit code and the low bits are extra info. So we need to use the WEXITSTATUS function after all.

python/tests/res/job_queue/test_jobmanager.py
```diff
@@ -476,5 +478,79 @@ def test_job_list_to_log_ordering(self):
...
+                run_script_name = "run.py"
+                with open(run_script_name, "w") as f:
+                    f.write("\n".join((
+                        '#!/usr/bin/env python',
+                        'import os',
+                        'from res.job_queue import JobManager',
+                        'jobm = JobManager()',
+                        'for job in jobm:',
+                        '    exit_status, msg = jobm.runJob(job)',
+                        'if exit_status != {}:'.format(exit_signal),
```
markusdregi: Should this one not be indented one line? It probably does not matter as there is only one job in the tests, so checking after the loop has the same effect. Nevertheless...

markusdregi:
I'm trying to further wrap my head around the `os.fork` usage here. As exemplified below, the processes will carry out what is inside it's respective block, and due to the `waitpid` call the child process will afterwards carry out it's execution, before the parent will do the same. Doing this inside a library method is a rather severe side effect I would argue!

```python
import os
import time

def f():
    pid = os.fork()
    if pid == 0:
        print("I'm the child ({})".format(pid))
    else:
        print("I'm the parent ({})".format(pid))
        _, exit_status = os.waitpid(pid, 0)

    return pid

print(f())
```

Which gives:
```
I'm the parent (56401)
I'm the child (0)
0
56401
```

Would we perhaps like to do the exit signal processing of the child immediately and then exit?

```python
import os
import time

def f():
    pid = os.fork()
    if pid == 0:
        print("I'm the child ({})".format(pid))
       # Continue doing my thing, executing stuff
       exit_code = process_exit_code(exit_code)
       exit(exit_code)
    else:
        print("I'm the parent ({})".format(pid))
        _, exit_status = os.waitpid(pid, 0)

    # Now only the parent will ever get here
    return pid

print(f())
```

What do you think @mortalisk?

python/tests/res/job_queue/test_jobmanager.py
mortalisk: Indenting the if statement seems reasonable, yes

mortalisk:
I'm trying to wrap my head around your previous comment, but it's early in the morning. Perhaps you could explain over a cup of coffee?

python/tests/res/job_queue/test_jobmanager.py
```diff
@@ -476,5 +478,79 @@ def test_job_list_to_log_ordering(self):
...
+                    }
+                ]
+                create_jobs_json(joblist)
 
+                jobm = JobManager()
+                for job in jobm:
+                    exit_status, msg = jobm.runJob(job)
+                    self.assertEqual(exit_status, exit_signal)
+
+    def test_exit_signals_pid(self):
```
mortalisk, line 513: What is this test testing?

markusdregi: My previous lack of understanding for fork. I believe it can be removed.

markusdregi:
@mortalisk I’m not in the office today, but I’ll happily answer questions regarding this. The main point is that a function that forks, without exiting one of them before returning will actually have the function return twice, once in each fork and hence it will touch disk twice also. Try running and playing around with the snippets I posted 😉

mortalisk:
I think I understand what you meant. But we are using execv which does not return, it replaces the entire process with a new process.

Anyway, we should maybe just use subprcess like @jondequinor suggested in #646

mortalisk:
Closed, since we are touching this in #653, and will probably be invalid after that

PR #645, opened 2019-08-13
Create compatibility jobs (uppercase) on install

andreabrambilla:
**Issue**
Compatibiliy and current version of jobs can get out of sync if in two separate files
Also, files with same name and different case make a mess on case-insensitive OSs

**Approach**
Uppercase version of jobs is created on install by copying the lowercase version. This is done only if the OS is detected to be case-sensitive


markusdregi:
To name them differently we have the forward model jobs (e.g. `ECLIPSE100`) and the simulation jobs (e.g. `eclipse100`). All ERT users are using the forward model jobs and all Everest users are using simulation jobs (more or less). We want to move towards the simulation job format, but for now the forward model jobs are by far the most used in the company.

> Compatibiliy and current version of jobs can get out of sync if in two separate files

To be honest they were supposed to be different. While the forward model jobs files would contain a description of the arguments, together with default values etc, the simulation job files are quite stripped, only pointing to an executable. Hence, I'm not sure whether this is a problem that needs to be addressed.

> files with same name and different case make a mess on case-insensitive OSs

Good point. How about separating the two description files into two different folders to avoid the case insensitive problem altogether? For instance, by moving the lower case versions from `share/ert/forward-models/res/` to `share/ert/forward-models/res/simulation_jobs`.

andreabrambilla:
Thanks for looking into this. I'll revert the changes to the files. Instead of the 2 separate folders, I was thinking of having different file names, which are them switched to the uppercase lowercase version when installing. I'll update the PR

markusdregi:
> which are them switched to the uppercase lowercase version when installing

But if they are then installed to the same location we end up with the same problems don't we?

andreabrambilla:
My idea was to install both version only if the file system is detected as being case-sensitive (that's the hacky `if` in this PR). If the file system is not case sensitive, only the lowercase version is installed (which is supposed to be the "new" version)

share/ert/forward-models/res/eclipse100
```diff
@@ -1 +1,5 @@
 EXECUTABLE script/ecl100
+DEFAULT <VERSION> version
```
markusdregi: In my opinion these jobs should not use any of the keywords besides the `EXECUTABLE`.

andreabrambilla: you are absolutely right, i screwed up while reverting

markusdregi (approved):
Thank you for your patience :)

andreabrambilla:
Thank you for checking this thoroughly ;)

PR #647, opened 2019-08-14
Memory diagnostics

sondreso:
**Issue**
https://github.com/equinor/ert/issues/389

**Approach**
Probe memory usage for each forward model component and add it to the status.json file

### DIFF
--- a/python/res/job_queue/forward_model_status.py
+++ b/python/res/job_queue/forward_model_status.py
@@ -42,7 +42,8 @@ class ForwardModelJobStatus(object):
                  status = "Waiting",
                  error=None,
                  std_out_file="",
-                 std_err_file=""):
+                 std_err_file="",
+                 reported_max_memory_usage=0):
 
         self.start_time = start_time
         self.end_time = end_time
@@ -51,6 +52,7 @@ class ForwardModelJobStatus(object):
         self.error = error
         self.std_out_file = std_out_file
         self.std_err_file = std_err_file
+        self.reported_max_memory_usage = reported_max_memory_usage
 
 
     @classmethod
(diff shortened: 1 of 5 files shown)


python/res/job_queue/job_manager.py
```diff
@@ -126,6 +127,7 @@ class JobManager(object):
 
     DEFAULT_UMASK =  0
     sleep_time    =  10  # Time to sleep before exiting the script - to let the disks sync up.
+    MEMORY_POLL_PERIOD = 0.5  # Time between memory polling of job process
```
andreabrambilla: is this in seconds? I assume users are mostly interested in a small selection of jobs (rms, eclipse, etc) which run for at least for a few minutes, but can go up to hours or days

python/res/job_queue/forward_model_status.py
```diff
@@ -51,6 +52,7 @@ def __init__(self, name,
         self.error = error
         self.std_out_file = std_out_file
         self.std_err_file = std_err_file
+        self.max_memory_usage = max_memory_usage
```
markusdregi: Should we add/inject `reported` or similar into the name to stress that this is not a constraint on the memory usage?

python/res/job_queue/job_manager.py
sondreso: Yes, I was wondering what an appropriate value would be for the poll period, adjust it to 5 seconds?

python/res/job_queue/forward_model_status.py
sondreso: :+1:

sondreso:
Anyone have input on how the memory usage should be reported back to the user? As of now, anything added to the status.json file will show up in the simulation overview when you click a specific realization. However, the fact that neither the unit is not displayed nor that the unit used is byte makes for a great user experience. 

image

sondreso:
Another discussion point: 

The accuracy of the current solution is not great. Right now the rss memory information is used, which seems to give the most sensible result of the different metrics available (https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_info).

Anther alternative is to use the uss metric from `memory_full_info` (https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_full_info), however this requires higher user privileges than memory_info() and is considerably slower.

Assuming that the user don't need extremely accurate information and mainly wants to know if jobs are using to much memory, I think the the current solution using the rss metric is sensible.

markusdregi:
> Assuming that the user don't need extremely accurate information and mainly wants to know if jobs are using to much memory, I think the the current solution using the rss metric is sensible.

That should not be a problem. Especially, if I got it correctly and `rss` is potentially overestimating the memory usage?

sondreso:
> Especially, if I got it correctly and `rss` is potentially overestimating the memory usage?

Yes, some memory usage of shared libraries might be included, along with other inaccuracies

PR #649, opened 2019-08-14
Replace mkdir with exit in failing jobmanager tst

### LINKED ISSUE #648: test_jobmanager test_run_multiple_fail is failing if run by root
Without investigating too much, it is probably due to `root` having all permissions always, whilst the test assumes setting a folder to read-only will prevent `root` from writing to it.

Either way, testing exit status codes could easily be done using the bash builtin `exit`.

Bash is usually the default shell on most distros, including osx (though soon to be replaced by zsh).

jondequinor:
**Issue**
Resolves #648

**Approach**
As the title indicates

### DIFF
--- a/python/tests/res/job_queue/test_jobmanager.py
+++ b/python/tests/res/job_queue/test_jobmanager.py
@@ -1,3 +1,4 @@
+import getpass
 import json
 import os
 import os.path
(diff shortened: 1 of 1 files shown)


andreabrambilla (approved):
:+1:

PR #650, opened 2019-08-14
Update status if target_file missing

### LINKED ISSUE #315: Wrong information in status.json when using target files
Jobs are labeled as a success in `status.json` if they do not produce their target file. Which obviously is wrong.

To reproduce, make replace the content of `POLY_EVAL` with:
```
EXECUTABLE poly_eval.py
TARGET_FILE my_target_file
```

and run `poly.ert`. Now, `POLY_EVAL` should produce a file called `my_target_file` to be considered correct. _ERT_ does in the end conclude that it failed, but the `status.json` and hence the progress window displays misleading information. In addition, the forward model continues after a job has failed...

jondequinor:
**Issue**
Resolves https://github.com/equinor/ert/issues/315

**Approach**
Do a small refactor such that `job_manager.exit` does not kill the current process, but allows `job_dispatch` to make decisions for what to do when a job fails.

### DIFF
--- a/python/res/job_queue/job_manager.py
+++ b/python/res/job_queue/job_manager.py
@@ -293,23 +293,32 @@ class JobManager(object):
             now = time.localtime()
             f.write("%-32s: %02d:%02d:%02d .... " % (job["name"], now.tm_hour, now.tm_min, now.tm_sec))
 
-    def markJobComplete(self, job, exit_status, error_msg):
-        now = time.localtime()
-        extra_fields = {"finished": True,
-                        "exit_status": exit_status,
-                        "status": "completeStatus"}
+    def jobComplete(self, job, success, exit_status, error_msg=None):
+        job_status = job["status"]
+
+        if success:
+            job_status.status = "Success"
+            job_status.error = None
+        else:
+            std_err_out = self.extract_stderr_stdout(job)
+            std_err_out.update({"status": "exit", "finished": True, "error_msg": error_msg, "exit_status": exit_status, "error": True})
+            self.postMessage(job=job, extra_fields=std_err_out)  # Posts to new logstash
+
+            job_status.status = "Failure"
+            job_status.error = error_msg
+
+            self.dump_EXIT_file(job, error_msg)
+
         with open(self.STATUS_file, "a") as f:
             if exit_status == 0:
                 status = ""
             else:
-                status = " EXIT: %d/%s" % (exit_status, error_msg)
-                extra_fields.update({"error_msg": error_msg})
+                status = " EXIT: {}/{}".format(exit_status, error_msg)
+                now = time.localtime()
 
-            f.write("%02d:%02d:%02d  %s\n" %
-                    (now.tm_hour, now.tm_min, now.tm_sec, status))
+            f.write("{:02d}:{:02d}:{:02d}  {}\n".format(now.tm_hour, now.tm_min, now.tm_sec, status))
 
-        status = job["status"]
-        status.end_time = dt.now()
+        job_status.end_time = dt.now()
 
     def createOKFile(self):
         now = time.localtime()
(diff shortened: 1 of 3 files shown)


python/res/job_queue/job_manager.py
```diff
@@ -293,8 +293,7 @@ def startStatus(self, job):
             now = time.localtime()
             f.write("%-32s: %02d:%02d:%02d .... " % (job["name"], now.tm_hour, now.tm_min, now.tm_sec))
 
-
-    def completeStatus(self, exit_status, error_msg, job=None):
+    def markJobComplete(self, job, exit_status, error_msg):
```
markusdregi: Should we make `error_msg` default to the empty string? And then possibly raise an exception if the `exit_status` is not `0`?

python/res/job_queue/job_manager.py
```diff
@@ -306,8 +305,11 @@ def completeStatus(self, exit_status, error_msg, job=None):
                 status = " EXIT: %d/%s" % (exit_status, error_msg)
                 extra_fields.update({"error_msg": error_msg})
 
-            f.write("%02d:%02d:%02d  %s\n" % (now.tm_hour, now.tm_min, now.tm_sec, status))
+            f.write("%02d:%02d:%02d  %s\n" %
```
markusdregi: Feel free to make this new-style formatting while your at it....

markusdregi:
Are you certain you have made it resolve the right issue?

python/res/job_queue/job_manager.py
markusdregi: Or even default to `None` perhaps?

python/res/job_queue/job_manager.py
```diff
@@ -467,15 +469,19 @@ def extract_stderr_stdout(self, job):
                     extra_fields.update({"stdout": stdout})
         return extra_fields
 
-    def exit(self, job, exit_status, error_msg):
-        self.dump_EXIT_file(job, error_msg)
+    def markJobFailure(self, job, exit_status, error_msg):
```
markusdregi: Would it make sense to merge `markJobFailure` and `markJobSuccess` and instead call it `markJob` and pass in a boolean `success` or have it be conditional on the exit status?

markusdregi: Or even put this as internal logic and have `markJobComplete` call it? If one is always expected to call `a()` and then `b()` I would say that the API is a bit off...

python/res/job_queue/job_manager.py
jondequinor: Sat default `error_msg` to `None`, but non-zero `exit_status` is perfectly fine for completed jobs.

python/res/job_queue/job_manager.py
jondequinor: Could not agree more!

python/res/job_queue/job_manager.py
```diff
@@ -293,21 +293,32 @@ def startStatus(self, job):
...
-                        "exit_status": exit_status,
-                        "status": "completeStatus"}
         with open(self.STATUS_file, "a") as f:
             if exit_status == 0:
                 status = ""
             else:
-                status = " EXIT: %d/%s" % (exit_status, error_msg)
-                extra_fields.update({"error_msg": error_msg})
+                status = " EXIT: {}/{}".format(exit_status, error_msg)
+                now = time.localtime()
```
andreabrambilla, line 317: Is this indented correctly? I guess it should be 4 spaces to the left

python/tests/res/job_queue/test_job_dispatch.py
```diff
@@ -81,16 +84,67 @@ def testTerminateJobs(self):
...
+            p.terminate()
+
+            with open("status.json", "r") as s:
+                status_json = s.read()
+                self.assertNotIn("Success", status_json)
+
+                # It is sufficient to test that Running is the status of the
+                # job. This means it will eventually fail after 60 seconds when
+                # the target_file can not be found.
+                self.assertIn("Running", status_json)
```
andreabrambilla, line 150: It is not entirely clear to me what we are testing for in this test

markusdregi:
Do we expect to merge this given #653 ?

jondequinor:
No.

PR #653, opened 2019-08-16
Refactoring job dispath

### LINKED ISSUE #316: Jobs killed due to time limit exceeded (`MAX_RUNTIME`) are categorized as complete
In the simulator, the callbacks give `complete` status for jobs that are killed.

Should either be `killed` or `failed`.

### LINKED ISSUE #646: Replace os.fork usage in job_manager with subprocess
Rationale:
https://www.python.org/dev/peps/pep-0324/#rationale

`subprocess` supports arbitrary code execution between `fork` and `exec`, and asynchronous code execution using `Popen`. There should be no need for UNIX-only `os.fork`.

### LINKED ISSUE #651: No status.json if max_running_minutes is set for a job
If max_running_minutes is set, no status.json file will be created. This results in the GUI behaving slightly odd, however simulations are completed. 

This is because the code path following this if test 
https://github.com/equinor/libres/blob/43527f8d66580b94b27e28399cf554e91c41075b/bin/job_dispatch.py#L277-L299

differs greatly from the normal path, even though the functionality is similar. Suggest integrating the max runtime functionality into the normal code path to make sure other functionalities are the same for both code paths. This is more sensible to do after #646.

### LINKED ISSUE #658: Exit signal condition
Original pr was #641
Remember to check that this is fixed after finishing pr #653 and consider keeping the tests from #641

### LINKED ISSUE #661: job max_running likely broken due to TOC/TOU-like issues
There is a Time-Of-Check Time-Of-Use bug in the `check_license` function for a job.

An `os.stat` is performed where the number of hard links are counted. Based on this count, a hard link is created or not. If not, the check sleeps.

Given that multiple processes run multiple jobs thus multiple concurrent `check_licenses`, this could lead to more licensed jobs running than `max_running` allows.

A non-portable solution could require minimal additional work using [fcntl](https://docs.python.org/3/library/fcntl.html).

A portable solution could be solved by implementing a local licensing server.

mortalisk:
**Overview:**
- Removed code paths that essentially did the same things (e.g. [`kill_job_and_EXIT`](https://github.com/equinor/libres/blob/5c88920febfb97a18c6a2d4b0e51bfd2de8f0241/bin/job_dispatch.py#L102)/[`JobManager.exit`](https://github.com/equinor/libres/blob/5c88920febfb97a18c6a2d4b0e51bfd2de8f0241/python/res/job_queue/job_manager.py#L470), [`JobManager.execJob`](https://github.com/equinor/libres/blob/5c88920febfb97a18c6a2d4b0e51bfd2de8f0241/python/res/job_queue/job_manager.py#L345)/[`exec_job`](https://github.com/equinor/libres/blob/5c88920febfb97a18c6a2d4b0e51bfd2de8f0241/bin/job_dispatch.py#L78))
- Created a `JobRunner` that has the responsibility of running the individual `Job`s.
- Created a `Job` that implements running a job.
- Created a `Reporter` which is subclassed by `Networking`,
  `File` and `Interactive`.
- Created a class of `Message`s that allows both `JobRunner` and `Job` to communicate with the reporters.
- Simplified testing and added tests for introduced classes.
- Removed job spawning using `os.fork` and `os.execvp`, replaced it with `Popen`.

Resolves #658  
Resolves #661 
Resolves #646 
Resolves #651 
Resolves https://github.com/equinor/libres/issues/316


python/res/job_queue/job.py
```diff
@@ -8,7 +8,7 @@
 #  (at your option) any later version.
 #
 #  ERT is distributed in the hope that it will be useful, but WITHOUT ANY
-#  WARRANTY; without even the implied warranty of MERCHANTABILITY or
+#  WARRANTY; without even the implied warranty of MEargListRCHANTABILITY or
```
andreabrambilla: what?

python/job_runner/reporting/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+The reporting package provides classes for reporting the results of forward
+model jobs.
+"""
+from .file import File
+from .interactive import Interactive
+from .network import Network
```
ManInFez: no new line on file

python/job_runner/reporting/file.py
```diff
@@ -0,0 +1,184 @@
...
+from job_runner.reporting.message import Exited, Finish, Init, Running, Start
+
+
+class File(object):
+    LOG_file = "JOB_LOG"
+    ERROR_file = "ERROR"
+    EXIT_file = "EXIT"
+    STATUS_file = "STATUS"
+    OK_file = "OK"
+    STATUS_FILE = "status.json"
```
ManInFez: couldnt jobs.json be defined here, instead of injected as magic strings everywhere?

ManInFez: Also all the is "_file" except status which is "_FILE"

python/job_runner/reporting/file.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def __init__(self):
+        self.status_dict = {}
+        self.node = socket.gethostname()
+
+    def report(self, msg, sync_disc_timeout=10):
```
ManInFez, line 23: File reporter has different signature(... , msg, .. ) Interactive and network ( ... , status, ... )

python/job_runner/reporting/file.py
ManInFez: I would maybe call STATUS_FILE STATUS_json, and perhaps fetch the values from their C counterpart in job_queue/forward_model.cpp

python/job_runner/runner.py
```diff
@@ -0,0 +1,87 @@
...
+                "{} does not exist. Available jobs: {}"
+                .format(unused, [j.name() for j in self.jobs])
+            )
+            yield init_message
+            raise StopIteration()
+        else:
+            yield init_message
+
+        all_jobs_successful = True
+        for job in job_queue:
```
ManInFez, line 63: Job runner should terminate if one job is not successful

ManInFez: There should be a test for this ;)

jondequinor: `job_dispatch` kills itself with `SIGKILL` if some job fails, but this should be documented in `JobRunner`, since that's the place you are likely to look.

We should also test this specifically in the `job_dispatch` testing.

python/job_runner/reporting/network.py
```diff
@@ -0,0 +1,115 @@
...
+
+            # Disabling proxies
+            proxies = {
+                "http": None,
+                "https": None,
+            }
+            requests.post(url, timeout=3,
+                          headers={"Content-Type": "application/json"},
+                          data=data, proxies=proxies)
+        except:  # noqa
```
ManInFez, line 94: Perhaps report somewhere if post is failing? And shouldn't we now what exception is being thrown?

python/job_runner/runner.py
ManInFez: Taken care of by job_dispatch.py

jondequinor: Yes, let's have `job_dispatch` kill on unsuccessful `Finish`, and have file reporter report everything it needs to on `Exited`.

python/job_runner/reporting/network.py
jondequinor: I agree, but let's do this as its own task: https://github.com/equinor/libres/issues/676

python/job_runner/runner.py
```diff
@@ -0,0 +1,87 @@
...
+from job_runner.job import Job
+from job_runner.reporting.message import Init, Finish
+
+
+class JobRunner(object):
+
+    def __init__(self, jobs_file="jobs.json"):
+        try:
+            with open(jobs_file, "r") as json_file:
+                self.jobs_data = json.load(json_file)
```
ManInFez: self.jobs_data does not have to be object attribute as it is not used anywhere else outside this scope

python/job_runner/runner.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self._data_root = self.jobs_data.get("DATA_ROOT")
+        if self._data_root:
+            os.environ["DATA_ROOT"] = self._data_root
+
+        self.simulation_id = self.jobs_data.get("run_id")
+        self.ert_pid = self.jobs_data.get("ert_pid")
+        self.global_environment = self.jobs_data.get("global_environment")
+        self.global_update_path = self.jobs_data.get("global_update_path")
+        self.job_data_list = self.jobs_data["jobList"]
```
ManInFez: job_data_list is also not used outside this scope

python/job_runner/runner.py
jondequinor: It is used in the `File` reporter: https://github.com/mortalisk/libres/blob/refactor_job_dispatch/python/job_runner/reporting/file.py#L134

ManInFez: Then there is a typo: jobs_data  != job_data

ManInFez: Correction this is a runner object, not a job object...

python/job_runner/reporting/file.py
jondequinor: > couldnt jobs.json be defined here, instead of injected as magic strings everywhere?

It could, but it is really a `JobRunner` thing.



> … perhaps fetch the values from their C counterpart in job_queue/forward_model.cpp

`#defines` are garbled compile time, so they're not in the DLL (at least not accessible using the original name).

And as Markus discuss in https://github.com/equinor/ert/issues/434#issuecomment-524188452 the real fix to this might be to drop the file system as means for communication between the queue and job_dispatch altogether?

ManInFez:
test this please

jondequinor:
test this please

python/job_runner/cli.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def main(args):
+    # If run_path is defined, enter into that directory
+    if len(args) > 1:
+        run_path = args[1]
+
+        if not os.path.exists(run_path):
+            sys.stderr.write("No such directory: %s\n" % run_path)
+            sys.exit(1)
```
markusdregi, line 18: Could not these two lines be replaced by:

```python
sys.exit("No such directory: {}".format(run_path))
```

markusdregi (approved):
Looks very good! :+1: I've written some issues to further extend this work. But I think all of those things should be addressed in new and smaller PR's...

Only thing is whether we should merge the `stderr` message and the system-exit call as commented below.

PR #663, opened 2019-08-26
Imp deprication

### LINKED ISSUE #640: imp package is deprecated for python>=3.4
Since python version 3.4 the imp package is pending deprecation in favor of importlib. Related to (https://github.com/equinor/ert/issues/407). Only gives warnings, so not strictly necessary to ensure python 3.6 support

sondreso:
**Issue**
Resolves #640

**Approach**
Dynamically import the relevant module based on the current python version. I am normally not very happy about imports that does not happen at the top of the file, however; in this case I think it is an ok solution.

### DIFF
--- a/python/res/job_queue/ert_script.py
+++ b/python/res/job_queue/ert_script.py
@@ -1,5 +1,4 @@
 import inspect
-import imp
 import sys
 import traceback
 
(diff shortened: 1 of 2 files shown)


python/res/job_queue/ert_script.py
```diff
@@ -112,9 +111,21 @@ def outputStackTrace(self, error=None):
     def loadScriptFromFile(path):
         """ @rtype: type ErtScript """
         try:
-            m = imp.load_source("ErtScriptModule_%d" % ErtScript.__module_count, path)
+            module_name = "ErtScriptModule_%d" % ErtScript.__module_count
+            if sys.version_info[0] == 3 and sys.version_info[1] >= 5:
```
markusdregi: I think the more standard way of writing this is `sys.version_info[:2] >= (3, 5)`. Feel that is slightly easier to read... Also, `elif` but no `else` is bad!

How about
```python
py_version = sys.version_info[:2]
if py_version == (2, 7):
    # Python 2.7 stuff
elif (3, 0) <= py_version < (3, 5):
   # Python 3 pre 3.5 stuff
elif (3, 5) <= py_version:
   # Python 3.5 and later stuff
else:
    raise Exception("Invalid Python verson {}:".format(py_version))
```

markusdregi (commented):
This looks good :+1: I had some suggestions for restructuring the conditioning. Besides that, this is ready to be merged

python/res/job_queue/ert_script.py
sondreso: Agreed!

markusdregi (approved):
Rebase and ship it :ship:

PR #666, opened 2019-08-26
Removed `put_time` so that we can build on gcc 4.8

andreabrambilla:
**Issue**
In order to use `put_time` we need gcc version >= 5 (or maybe even 6, unsure about this), but that is not necessarily available on some of the machines we normally work on (eg the azure cluster)

**Approach**
Replace `put_time` with a local implementation that has a similar behavior


markusdregi (approved):
Looks good to me

PR #674, opened 2019-09-04
Specify libecl version dependency in README

jondequinor:
**Issue**
Colloquially it is known that `libres` needs to be built against a specific version of `libecl`.

**Approach**
Add it to README so as to have it in writing.


sondreso (approved):

ManInFez (approved):
:1st_place_medal:

PR #689, opened 2019-09-26
Output deprecation warnings on simulation wf jobs

### LINKED ISSUE #684: Deprecate all workflow jobs running an ensemble
The tooling to running several ensembles using workflows will be replaced by batch scripts using the CLI. All workflow jobs running an ensemble should be deprecated

ManInFez:
resolves #684

### DIFF
--- a/lib/enkf/enkf_main_jobs.cpp
+++ b/lib/enkf/enkf_main_jobs.cpp
@@ -76,6 +76,7 @@ void * enkf_main_dummy_JOB(void * self , const stringlist_type * args ) {
 
 
 void * enkf_main_ensemble_run_JOB( void * self , const stringlist_type * args ) {
+  printf("WARNING: The ENSEMBLE_RUN job is deprecated and will soon be removed!\n");
   enkf_main_type   * enkf_main = enkf_main_safe_cast( self );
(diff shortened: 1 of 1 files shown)


lib/enkf/enkf_main_jobs.cpp
```diff
@@ -76,6 +76,7 @@ void * enkf_main_dummy_JOB(void * self , const stringlist_type * args ) {
 
 
 void * enkf_main_ensemble_run_JOB( void * self , const stringlist_type * args ) {
+  printf("WARNING: The ENSEMBLE_RUN job is deprecated and will soon be removed!\n");
```
oyvindeide, line 79: Looks good, but perhaps we should put a date on when it will be removed? Guess standard is 1 year, but maybe end of this year?

oyvindeide (approved):

lib/enkf/enkf_main_jobs.cpp
markusdregi: Dates are always scary, because then you suddenly have a warning saying this will be removed in 2012 😂 Also, I don’t expect these messages to hit the users. So we’ll have to make announcements in addition 👍

PR #695, opened 2019-10-02
Include EnkfSimulationRunner in EnkfMain

oyvindeide:
This includes EnkfSimulationRunner in EnkfMain to make sure that ErtScript gets access to EnkfMain and not realEnkfMain as it currently does.

### DIFF
--- a/python/res/enkf/enkf_main.py
+++ b/python/res/enkf/enkf_main.py
@@ -76,7 +76,6 @@ class EnKFMain(BaseCClass):
             parent=real_enkf_main,
             is_reference=True)
 
-        self.__simulation_runner = EnkfSimulationRunner(self)
         self.__fs_manager = EnkfFsManager(self)
         self.__es_update = ESUpdate(self)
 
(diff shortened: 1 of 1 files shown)


python/res/enkf/enkf_main.py
```diff
@@ -99,6 +98,11 @@ def umount(self):
         if self.__fs_manager is not None:
             self.__fs_manager.umount()
 
+    def runWorkflows(self , runtime):
```
markusdregi, line 101: This is not really a fix is it? It just reintroduces the cyclic dependency..

oyvindeide: Right, there is probably a better solution to this, could be to do: 
```

    def runWorkflows(self , runtime):
        """:type res.enkf.enum.HookRuntimeEnum"""
        hook_manager = self._enkf_main().getHookManager()
        hook_manager.runWorkflows(runtime  , self._enkf_main())
```
to:
```
    @staticmethod
    def runWorkflows(ert , runtime):
        """:type res.enkf.enum.HookRuntimeEnum"""
        hook_manager = ert._enkf_main().getHookManager()
        hook_manager.runWorkflows(runtime  , ert)
```
Should probably have a discussion on how we can solve this, is required to run most internal python workflows.

oyvindeide: This has been implemented in #744

PR #706, opened 2019-10-03
Fix system exit

mortalisk:
Improves 2 lines of code, as suggested in #653


ManInFez (approved):
Looks good to me :+1:

PR #711, opened 2019-10-07
Show memory diagnostics in GUI

### LINKED ISSUE #389: Memory diagnostic and reporting for forward models
On cluster nodes with limited memory, it is beneficial / necessary for the user to provide the queuing system upfront information on how much memory is required. Average users typically have no idea. It would be very useful if ERT could monitor and report memory usage for jobs as a whole and for each forward model component ( this one is more tricky).

sondreso:
**Issue**
Resolves https://github.com/equinor/ert/issues/389

**Approach**
Add memory information to the job status object.

image

### DIFF
--- a/python/res/job_queue/forward_model_status.py
+++ b/python/res/job_queue/forward_model_status.py
@@ -45,7 +45,9 @@ class ForwardModelJobStatus(object):
                  status="Waiting",
                  error=None,
                  std_out_file="",
-                 std_err_file=""):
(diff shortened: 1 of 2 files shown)


ManInFez:
~~Should we make the field in the detailed view a bit more human readable? show KB, MB, GB, etc..  when applicable?~~

This should be done in ert, not in libres.... my bad

ManInFez (approved):
:+1:

PR #725, opened 2019-10-09
Fix test

oyvindeide:
Many tests were manipulating environment variables without cleanup, here we add cleanup.


python/res/fm/rms/rms_run.py
```diff
@@ -96,7 +96,7 @@ def _run_child(self):
         exec_env_file = "%s_exec_env.json" % self_exe
         exec_env = os.environ.copy()
         if os.path.isfile(exec_env_file):
-            exec_env.update(json.load(open(exec_env_file)))
+            self.set_up_env(exec_env, json.load(open(exec_env_file)))
```
ManInFez: set_up_env is a static method and should not be called with self

ManInFez (approved):
Good job!! Rebase and squash into one :)

PR #726, opened 2019-10-09
Deprecate and refactor certain config elements

jondequinor:
**Issue**
Resolves https://github.com/equinor/ert/433


markusdregi (approved):
Looks good 👍 Very nice to get rid of these old keywords when moving forward!

PR #727, opened 2019-10-09
Run tests in tmpdir

oyvindeide:
Several tests were running copying files to the source tree, this moves it into tmpdir.

### DIFF
--- a/python/tests/res/enkf/data/test_custom_kw.py
+++ b/python/tests/res/enkf/data/test_custom_kw.py
@@ -1,4 +1,7 @@
 import os
+
+import pytest
+
 from res.enkf.enums import EnkfRunType
 from res.enkf import ErtRunContext
 from res.enkf.config import CustomKWConfig
@@ -11,6 +14,8 @@ from tests import ResTest
 from ecl.util.test.test_area import TestAreaContext
 from ecl.util.util import StringList
 from ecl.util.util import BoolVector
+
+from tests.utils import tmpdir
 
 class CustomKWTest(ResTest):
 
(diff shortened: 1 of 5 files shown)


markusdregi:
Should we use the existing functionality instead? https://github.com/equinor/libres/blob/b2801739be29a652e733f2ae239ea0ec25198683/python/tests/utils/__init__.py#L16

oyvindeide:
> Should we use the existing functionality instead?

The reason I like the pytest version is that it handles tmpdir teardown while keeping around the test-data of a few runs, making debugging easier. And since we are already using pytest we don't have to re-implement existing libraries. Of course it adds more dependency on pytest. Will change if wanted though.

markusdregi:
My main argument is that we should do it in the same manner in all of the repo, unless we have a very good reason to do otherwise.

I don't mind moving to your suggested implementation, but then I would prefer if we reimplement the existing one on top your code. So that `@tmpdir` is still the way to go about it, with `@tmpdir("my/super/data")` still supported, but with the benefits of the `pytest` implementation.

You can do it now in this PR or move the snippet into an issue; that is your call ;)

oyvindeide:
Changed it to use existing functionality, also changed the default for the decorator to have path as an optional parameter as the called function tmp already has it as an optional parameter. Will create an issue to use the builtin pytest tmpdir, as some of the functionality is convenient and helps with teardown and debugging.

python/tests/res/enkf/data/test_custom_kw.py
```diff
@@ -78,7 +83,7 @@ def test_custom_kw_config_data_is_null(self):
                 self.assertFalse( "VALUE_3" in custom_kw_config )
 
 
-
+    @tmpdir()
```
markusdregi, line 86: Does this have any effect? Is not `ErtTestContext` supposed to do this?

markusdregi (approved):
I approve! Please make an issue listing the tests that do now utilize both `tmpdir` and `TestAreaContext` and explain why this was necessary so that we can address it later on.

PR #729, opened 2019-10-10
Remove testAreaContext from test_ecl_run

oyvindeide:
testAreaContext might be creating problems by keeping file handles open, so switching to tmpdir seems reasonable.

### DIFF
--- a/python/tests/res/fm/test_ecl_run.py
+++ b/python/tests/res/fm/test_ecl_run.py
@@ -26,6 +26,7 @@ import re
 from ecl.summary import EclSum
 from ecl.util.test import TestAreaContext
 from tests import ResTest
+from tests.utils import tmpdir
 from res.fm.ecl import *
 from subprocess import Popen, PIPE
 from distutils.spawn import find_executable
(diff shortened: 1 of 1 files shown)


python/tests/res/fm/test_ecl_run.py
```diff
@@ -94,84 +94,80 @@ def init_flow_config(self):
...
-            ta.copy_file( os.path.join(self.SOURCE_ROOT , "test-data/local/eclipse/SPE1.DATA"))
-            ta.copy_file( os.path.join(self.SOURCE_ROOT , "test-data/local/eclipse/SPE1_ERROR.DATA"))
-            os.makedirs("ecl_run")
-            shutil.move("SPE1.DATA", "ecl_run")
-            shutil.move("SPE1_ERROR.DATA" , "ecl_run")
-            flow_config = FlowConfig()
-            sim = flow_config.sim()
-            flow_run = EclRun("ecl_run/SPE1.DATA", sim)
-            flow_run.runEclipse( )
+        self.init_flow_config()
```
andreabrambilla, line 160: I suppose this test should get the `use_tmpdir` fixture as well, right?

andreabrambilla:
once this works, can we also remove the `from ecl.util.test import TestAreaContext` on line 27, or do we still need it in some cases?

python/tests/res/fm/test_ecl_run.py
oyvindeide: Yes, which is in #727, though we might consider rewriting the whole tmpdir implementation somewhat based on feedback in that PR.

andreabrambilla: I am not sure I follow, in #727 `test_ecl_run.py` is not touched

oyvindeide:
> once this works, can we also remove the `from ecl.util.test import TestAreaContext` on line 27, or do we still need it in some cases?

Good catch, yes, have updated

python/tests/res/fm/test_ecl_run.py
oyvindeide: You're right, I misread your comment, I've removed TestAreaContext from test_flow

andreabrambilla (approved):

Issue #701, opened 2019-10-03
Enable black for `job_runner`


markusdregi:
The `job_runner` module is entirely new. Make it black compatible and add blacking of that module as part of our CI/test-suite.

PR #730, opened 2019-10-11
Remove fork

### LINKED ISSUE #712: Flaky tests
Several tests are flaky, in particular this relates to `test_ecl_run.py`. Part of the reason could be that the environment changes between tests, for example is site-config changed in several places without being reset.

sondreso:
**Issue**
Resolves #712 

**Approach**
Remove forking from ecl_run

### DIFF
--- a/python/res/fm/ecl/ecl_run.py
+++ b/python/res/fm/ecl/ecl_run.py
@@ -4,6 +4,8 @@ import time
 import datetime
 import socket
 from collections import namedtuple
+import subprocess
+from contextlib import contextmanager
 
 try:
     from ecl.summary import EclSum
@@ -34,6 +36,12 @@ def make_LSB_machine_list( LSB_HOSTS ):
     return LSB_HOSTS.split()
 
 
+@contextmanager
+def pushd(run_path):
+    starting_directory = os.getcwd()
+    os.chdir(run_path)
+    yield
+    os.chdir(starting_directory)
 
 
 class EclRun(object):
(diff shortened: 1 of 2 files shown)


python/res/fm/ecl/ecl_run.py
```diff
@@ -164,37 +165,34 @@ def initMPI(self):
...
-
-        os.close(fd_stdout)
-        os.close(fd_stderr)
-
-
-
-    def execSerialEclipse(self):
-        os.execv( self.sim.executable , [ self.sim.executable , self.base_name ])
+        with open(stdout_file, "w") as stdout, open(stderr_file, "w") as stderr:
+            process = subprocess.Popen([self.sim.executable, self.base_name], env=self.sim.env, stdout=stdout, stderr=stderr)
```
markusdregi: Please split this line. Something like:
```python
process = subprocess.Popen(
    [self.sim.executable, self.base_name],
    env=self.sim.env,
    stdout=stdout,
    stderr=stderr,
    )

Comment (markusdregi) [reply python/res/fm/ecl/ecl_run.py]: > Note This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough
> output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use 
> Popen.communicate() when using pipes to avoid that.

File: python/res/fm/ecl/ecl_run.py [hunk 2]
```diff
@@ -164,37 +165,34 @@ def initMPI(self):
...
+        stderr_file = "%s.stderr" % self.base_name
+        with open(stdout_file, "w") as stdout, open(stderr_file, "w") as stderr:
+            process = subprocess.Popen([self.sim.mpirun,
+                                   "-machinefile",
+                                   self.machine_file,
+                                   "-np",
+                                   "%s" % self.num_cpu,
+                                   self.sim.executable,
+                                   self.base_name],
+                                  env=self.sim.env, stdout=stdout, stderr=stderr)
```
markusdregi: I suggest you unfold the entire list of arguments...

python/res/fm/ecl/ecl_run.py
```diff
@@ -203,38 +201,32 @@ def execEclipse(self):
...
-                self.assertECLEND( )
-                if self.num_cpu > 1:
-                    self.summary_block( )
-
-                with open(OK_file, "w") as f:
-                    f.write("ECLIPSE simulation OK")
-            else:
-                with open(OK_file, "w") as f:
-                    f.write("ECLIPSE simulation complete - NOT checked for errors.")
+            with open(OK_file, "w") as f:
```
markusdregi: Change the order of these two cases to allow for an early exit...

python/res/fm/rms/rms_run.py
```diff
@@ -96,7 +96,7 @@ def _run_child(self):
         exec_env_file = "%s_exec_env.json" % self_exe
         exec_env = os.environ.copy()
         if os.path.isfile(exec_env_file):
-            exec_env.update(json.load(open(exec_env_file)))
+            RMSRun.set_up_env(exec_env, json.load(open(exec_env_file)))
```
markusdregi: I suggest you encompass this within a `with`-statement opening the file.

python/res/fm/rms/rms_run.py
```diff
@@ -153,22 +153,22 @@ def exec_rms(self, exec_env):
...
-            os.execve( self.config.executable , args, env )
+            os.execve(self.config.executable, args, exec_env)
         else:
-            os.execv(self.config.executable, args )
-
-
+            os.execv(self.config.executable, args)
+
+
+    @staticmethod
```
markusdregi: This has nothing to do with the class per see, I suggest you move the implementation out of the class and put it above or below in this file... Also, remember to consider whether functions are part of the public API. If not, prefixing by `_` is not a bad idea ;)

python/res/fm/rms/rms_run.py
```diff
@@ -153,22 +153,22 @@ def exec_rms(self, exec_env):
...
+    def set_up_env(exec_env, env_extension):
+        """
+        This method takes in two dicts of environmental variables and sets up and
+        execute environment. If the values are empty the key will be popped.
```
markusdregi: Please explain the difference between `exec_env` and `env_extension`. There is definitively someone in the future that will wonder what this does

python/res/fm/rms/rms_run.py
```diff
@@ -153,22 +153,22 @@ def exec_rms(self, exec_env):
...
+        """
+        for key, value in env_extension.items():
+            if value is None or len(str(value).strip()) == 0:
+                if key in exec_env:
+                    exec_env.pop(key)
```
markusdregi: Should we copy `exec_env` and use return an edited version of that copy instead to avoid direct manipulation of an argument?

python/res/fm/ecl/ecl_run.py
```diff
@@ -164,37 +165,34 @@ def initMPI(self):
...
+                                   "%s" % self.num_cpu,
+                                   self.sim.executable,
+                                   self.base_name],
+                                  env=self.sim.env, stdout=stdout, stderr=stderr)
+            return process.wait()
 
 
     def execEclipse(self):
-        os.chdir( self.run_path )
+        cur_dir = os.path.realpath(os.curdir)
```
markusdregi: I think ` os.getcwd()` is the right function to use here, not `os.curdir`

python/res/fm/ecl/ecl_run.py
```diff
@@ -164,37 +165,34 @@ def initMPI(self):
...
 
-        self.initEnv( )
-        self.initRedirect( )
+        os.chdir(self.run_path)
```
markusdregi: Could you please implement this as a context manager

python/tests/conftest.py
```diff
@@ -14,6 +14,14 @@ def source_root():
     raise RuntimeError('Cannot find the source folder')
 
 
+@pytest.fixture()
+def use_tmpdir(tmpdir):
```
markusdregi: This functionality was just merged in _libres_, please use this on instead: https://github.com/equinor/libres/blob/eb8a93b0d6b2b770720d5c273b2647b781f6c655/python/tests/utils/__init__.py#L16

markusdregi:
Overall this looks good :+1: Made a couple of comments that should be addressed...

python/res/fm/ecl/ecl_run.py
```diff
@@ -164,77 +171,77 @@ def initMPI(self):
...
+                stdout=stdout,
+                stderr=stderr)
+            process.communicate()
+            return process.wait()
 
-        os.close(fd_stdout)
-        os.close(fd_stderr)
 
+    def execEclipse(self):
+        with temp_dir_change():
```
markusdregi: Suggestion: How about if we passed the runpath as an argument. Something along the lines of:

```python
with pushd(self.run_path):
    # Do some stuff
```

python/res/fm/ecl/ecl_run.py
```diff
@@ -164,77 +171,77 @@ def initMPI(self):
...
 
+            if not os.access( self.data_file , os.R_OK ):
+                raise OSError("Can not read data file:%s" % self.data_file )
 
-    def execParallellEclipse(self):
-        self.initMPI( )
-        os.execv( self.sim.mpirun , [self.sim.mpirun , "-machinefile" , self.machine_file , "-np" , "%s" % self.num_cpu , self.sim.executable , self.base_name ] )
-
+            if self.num_cpu == 1:
+                return_code = self.execSerialEclipse()
```
markusdregi: With the context manager you can go back to returning directly here if you prefer ;)

python/tests/res/fm/test_ecl_run.py
```diff
@@ -216,13 +216,8 @@ def test_run_api(self):
...
             shutil.move("SPE1.DATA" , "ecl_run")
             ecl_config = Ecl100Config( )
-            run(ecl_config, ["ecl_run/SPE1.DATA", "--version=2014.2"])
-
-            self.assertTrue( os.path.isfile("ecl_run/SPE1.DATA"))
-            self.assertTrue( os.path.isfile("ecl_run/SPE1.DATA"))
             self.assertTrue( os.path.isfile("ecl_run/SPE1.DATA"))
-
-
+            run(ecl_config, ["ecl_run/SPE1.DATA", "--version=2014.2"])
```
markusdregi, line 220: Should we keep one of the asserts after the run as well?

python/res/fm/ecl/ecl_run.py
```diff
@@ -164,77 +172,73 @@ def initMPI(self):
...
-        os.execv( self.sim.executable , [ self.sim.executable , self.base_name ])
+        with open(stdout_file, "w") as stdout, open(stderr_file, "w") as stderr:
+            process = subprocess.Popen(
+                [self.sim.executable, self.base_name],
+                env=self.sim.env,
+                stdout=stdout,
+                stderr=stderr
+            )
+            process.communicate()
+            return process.wait()
```
markusdregi: According to the documentation `Popen.communicate` waits for the process to finish. Hence, I think the intended usage here is to return `process.returncode`.

python/res/fm/ecl/ecl_run.py
```diff
@@ -164,77 +172,73 @@ def initMPI(self):
...
+                [self.sim.mpirun,
+                 "-machinefile",
+                 self.machine_file,
+                 "-np",
+                 "%s" % self.num_cpu,
+                 self.sim.executable,
+                 self.base_name],
+                env=self.sim.env,
+                stdout=stdout,
+                stderr=stderr)
```
markusdregi: This is getting really picky; so besides making the line breaking of the `Popen` calls on the same format (one is `stderr)` and the other is `stderr\n)`) feel free to ignore my comments below.

Personally I like
```python
process = subprocess.Popen(
    ......,
    stdout=stdout,
    stderr=stderr,
)
```
because there is nothing special about the last arguments. I find it easier to read. In addition it is more friendly with respect to git history; because you can inject another argument and get a single plus line.

In addition, I'm not a big fan of non-standard indents used for code alignment, as they quickly go out of sync. My personal preference, as well as _black_'s preference (the tool we use on the repositories we have activated automatic code formatting) is the following if you are to split a list over multiple lines:
```python
process = subprocess.Popen(
    [
        self.sim.mpirun,
        "-machinefile",
        self.machine_file,
        "-np",
        "%s" % self.num_cpu,
        self.sim.executable,
        self.base_name,
    ],
    env=self.sim.env,
    stdout=stdout,
    stderr=stderr,
)
```
As stated above, do with these comments as you like ;)

sondreso: Completely agree, I just wish we could start using black so that we could enforce this everywhere. It's easy to forget the formatting when I can't just format the entire file.

markusdregi: I agree. This one would be a start: https://github.com/equinor/libres/issues/701 ;)

markusdregi (approved):
Well done! :tada:

PR #732, opened 2019-10-14
Fix potential garbage in model config

andreabrambilla:
Realloc can reallocate memory, if the 2 pointers passed to `util_realloc_string_copy are the same`, we can either copy garbage because the source string is deallocated, or break one of the preconditions of `strcpy`, therefore leading to undefined behavior

I have run a few test, and in practice it is unlikely that this issue hit any of our users. I noticed it only because (seemingly) some explicit precondition checks are enabled when compiling libres with clang


ManInFez (approved):
LGTM

PR #734, opened 2019-10-14
Made check for modified environment less strict

andreabrambilla:
**Issue**
When a test manipulates env variables, the final order of the list of env variables may be different from the initial one, even though the content is unchanged

**Approach**
Compare sets instead of lists


markusdregi (approved):
I would prefer `sorted`, as otherwise we are explicitly ruling out checking for duplicate elements. In general I find `sorted` easier to verify in these scenarios when reading code. However, I agree that it does not matter in this case, so merge if you want prefer it like this 👍

andreabrambilla:
I thought about the issue with duplicates, but the lists we have here are constructed from `os.environ`, which is (some kind of) dictionary, and therefore cannot include duplicates

Issue #737, opened 2019-10-16
Unstable tests


oyvindeide:
Several tests appear to influence each other, some, not all, of these are marked as `pytest.mark.unstable`.

oyvindeide:
The tests are now split into three separate chunks, which run independently, 
`pytest -m "equinor_test and not unstable"`, `pytest -m "not equinor_test"` and `pytest -m "unstable"`. To replicate the problem run `pytest` with all tests (including Equinor specific tests). The tests don't fail gracefully, they fail with `Aborted (core dumped)`, taking down `pytest` with them. Several things have been tried, including stopping test files being produced in the source tree and preventing the manipulating of env. variables during tests. Depending on which tests are being run the problem will manifest itself in different places, and it should be noted that none of the tests marked as unstable are failing. The list of unstable tests is probably not exhaustive.

oyvindeide:
For more context, a lot of tests are keeping file handles open, this could be related to the problem or a coincidence.

sondreso:
We sometimes experienced the mutex error described in https://github.com/equinor/libres/issues/781. Might be related somehow?

PR #738, opened 2019-10-16
Mark unstable tests and run seperately

oyvindeide:
Several tests appear to influence each other, this marks them as unstable and runs them separately. Have made an issue in: #737

### DIFF
--- a/testjenkins.sh
+++ b/testjenkins.sh
@@ -152,7 +152,15 @@ run_pytest_normal () {
 run_pytest_equinor () {
 	run enable_environment
 	pushd $LIBRES_ROOT/python
-	python -m pytest -s -m "equinor_test"
+	python -m pytest -s -m "equinor_test and not unstable"
+	popd
+}
+
+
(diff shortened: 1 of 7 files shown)


testjenkins.sh
```diff
@@ -152,7 +152,15 @@ run_pytest_normal () {
 run_pytest_equinor () {
 	run enable_environment
 	pushd $LIBRES_ROOT/python
-	python -m pytest -s -m "equinor_test"
+	python -m pytest -s -m "equinor_test and not unstable"
```
sondreso, line 155: Should `" ... and not unstable"` also be added to `run_pytest_normal`?

sondreso (approved):
Except for the minor comment this looks good :+1:

testjenkins.sh
oyvindeide: I did leave it out on purpose, as long as it doesn't provide problems that could give us an early warning if a new test is causing problems.

PR #739, opened 2019-10-17
Add scaling job entry point

oyvindeide:
This adds the entry point for the observations scaling job

### DIFF
--- a/python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
+++ b/python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
@@ -3,6 +3,8 @@ import os
 import shutil
 import random
 import configsuite
+import yaml
+
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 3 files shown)


share/ert/workflows/jobs/internal/scripts/enkf_scaling_job.py
```diff
@@ -0,0 +1,29 @@
...
+from res.enkf import ErtScript
+from res.enkf.jobs.scaling import scaling_job
+
+
+class EnkfScalingJob(ErtScript):
+
+    def run(self, job_config_file):
+
+        user_config = load_yaml(job_config_file)
+
```
xjules: do we need the empty lines?

python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
```diff
@@ -186,6 +188,33 @@ def test_old_enkf_scaling_job(setup_ert):
...
+    obs_vector = obs["POLY_OBS"]
+
+    assert_obs_vector(obs_vector, 1.0)
+
+    job_config = {"CALCULATE_KEYS": {"keys": ["POLY_OBS"]}}
+
+    with open("job_config.yml", "w") as fout:
+        yaml.dump(job_config, fout)
+
+    job = ert.getWorkflowList().getJob("CORRELATE_OBSERVATIONS")
```
xjules, line 205: More suggestions shall come:
OUTLIER_OBS_SCALING
PCA_OBS_SCALING

share/ert/workflows/jobs/internal/scripts/enkf_scaling_job.py
oyvindeide: No we don't, I've removed them, you're right, easier to read that way.

python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
oyvindeide: Good suggestions, would like it to be as simple and descriptive as possible. The question is, if we expand the usage in the future, should we make a new job, or a more complex config? A more complex config where the user can for example specify type of normalization, scaling factor calculation, etc.

xjules (approved):
Well done!

PR #741, opened 2019-10-17
Update python3 warning

oyvindeide:
Modify python 3 warning and make cmake fail if python wrappers are
enabled and python version is too old. Almost identical to: https://github.com/equinor/libecl/commit/f0400944af3e17e8f62faa7819bea21fd54caa9a with one important difference, this commit will throw an error instead of printing a warning if the python version is not compatible. The rationale being that if python wrappers are intended to be enabled but are not, that will make `libres` unusable.


ManInFez:
python version is always checked even if python is not enabled. This results in cmake error with python < 2.7 without python enabled.

oyvindeide:
> python version is always checked even if python is not enabled. This results in cmake error with python < 2.7 without python enabled.

Should not happen with ENABLE_PYTHON = OFF? BUILD_PYTHON is deprecated

ManInFez (approved):
tested on python version "pil-og-bue" (2.6.9) and is well behaved

PR #744, opened 2019-10-18
Make runWorkflows static

ManInFez:
In order to feed ert into the workflows in "sane" matter we use runWorkflow as a static method

### DIFF
--- a/python/res/enkf/enkf_simulation_runner.py
+++ b/python/res/enkf/enkf_simulation_runner.py
@@ -77,11 +77,11 @@ class EnkfSimulationRunner(BaseCClass):
         """ @rtype: int """
         return self.runSimpleStep(job_queue, run_context)
 
-
-    def runWorkflows(self , runtime):
+    @staticmethod
+    def runWorkflows( runtime, ert):
         """:type res.enkf.enum.HookRuntimeEnum"""
-        hook_manager = self._enkf_main().getHookManager()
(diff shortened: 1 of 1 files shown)


python/res/enkf/enkf_simulation_runner.py
```diff
@@ -77,11 +77,11 @@ def runEnsembleExperiment(self, job_queue, run_context):
...
 
-
-    def runWorkflows(self , runtime):
+    @staticmethod
+    def runWorkflows( runtime, ert):
         """:type res.enkf.enum.HookRuntimeEnum"""
-        hook_manager = self._enkf_main().getHookManager()
-        hook_manager.runWorkflows(runtime  , self._enkf_main())
+        hook_manager = ert.getHookManager()
+        hook_manager.runWorkflows(runtime  , ert)
```
andreabrambilla, line 84: Just discussed this with @oyvindeide 
This should work, and we think we should push it further. Let's get rid of this module entirely. Almost all of the methods here can be made free functions (by passing in the `ert` instance) , and we can move them to `enkf_main.py`
The only method that cannot be made free is `createRunPath`, but that method is already available in the EnkfMain class (with a slightly different uppercase/lowercase combination)

andreabrambilla (approved):
ok, we can postpone further refactoring to later on, go ahead :+1:

PR #716, opened 2019-10-08
Fix for collections ABC import DeprecationWarning

wouterjdb:
**Issue**
Resolves the DeprecationWarning that `collections` throws in stat.py

`DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working`

**Approach**
Did as suggested by the warning, changed `from collections import Sequence` to `from collections.abc import Sequence`.

PR #746, opened 2019-10-18
Fix for collections ABC import DeprecationWarning

oyvindeide:
From: 
https://github.com/equinor/libres/pull/716

### DIFF
--- a/python/res/util/stat.py
+++ b/python/res/util/stat.py
@@ -13,7 +13,10 @@
 #
 #  See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
 #  for more details.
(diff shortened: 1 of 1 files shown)


python/res/util/stat.py
```diff
@@ -13,7 +13,10 @@
 #
 #  See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
 #  for more details.
-from collections import Sequence
+try:
+    from collections.abc import Sequence
+except ImportError:
+    from collections import Sequence
```
mortalisk, line 19: Looks good

mortalisk (approved):

PR #748, opened 2019-10-21
Fix cmake Werror flag

ManInFez:
Testing travis with " instead of '  for setting Werror flag


andreabrambilla (approved):

andreabrambilla:
See also https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66943

PR #749, opened 2019-10-21
Move the nan filtering to after index filtering

oyvindeide:
When nans were filtered out before the index list was applied the index would not match the input data.


jondequinor (approved):
LGTM

PR #751, opened 2019-10-22
Replace sys.exit(1) with raising error

oyvindeide:
Sys.exit(1) in an internal job in ERT causes the whole process
to freeze.

### DIFF
--- a/python/res/enkf/jobs/scaling/scaling_job.py
+++ b/python/res/enkf/jobs/scaling/scaling_job.py
@@ -25,11 +25,9 @@ def scaling_job(ert, user_config_dict):
     config = setup_configuration(config_dict, job_config.build_schema())
 
     if not valid_configuration(config):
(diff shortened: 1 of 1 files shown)


jondequinor (changes requested):
`NameError` is reserved for actual naming issues. Perhaps `ValueError` is more appropriate?

From https://docs.python.org/3/library/exceptions.html#ValueError:

> Raised when an operation or function receives an argument that has the right type but an inappropriate value \[…\].

oyvindeide:
> `NameError` is reserved for actual naming issues. Perhaps `ValueError` is more appropriate?
> 
> From https://docs.python.org/3/library/exceptions.html#ValueError:
> 
> > Raised when an operation or function receives an argument that has the right type but an inappropriate value […].

Good point, I've updated accordingly.

jondequinor (approved):

PR #688, opened 2019-09-26
Add testkomodo.sh

mortalisk:
Also fix c tests library names to enable running against komodo.

Partial solution to equinor/ert#387

Reviewer:

- The build script in the Jenkins job [libres-komodo-run](https://ci.equinor.com/be/view/SCOUT/job/libres-komodo-run/) first sources komodo, checks out corresponding git tag, and then runs the testkomodo.sh script.
- The Jenkins build is currently running this code on my fork mortalisk/libres on master branch. After merge this must be switched to use equinor/libres in the jenkins build.
- There will still be some python tests failing, but c tests should pass for bleeding at least.
- New issues should be made for each failing test after merge.

PR #755, opened 2019-10-22
Don't pipe Eclipse output to LOG

### LINKED ISSUE #498: Pipe eclipse stdout to stdout
The interesting log from Eclipse is not the one currently written to stdout by the forward model. This is instead written to `LOG`, but possibly also to stdout. Check if we can pass our stdout pipe to the eclipse script. If so, it would be nice to get this output in stdout so that it is available in the monitoring in the GUI.

Check if possible to do the same for ERROR.

oysteoh:
Resolves https://github.com/equinor/ert/issues/498

### DIFF
--- a/python/res/fm/ecl/ecl_run.py
+++ b/python/res/fm/ecl/ecl_run.py
@@ -173,43 +173,33 @@ class EclRun(object):
 
 
     def execSerialEclipse(self):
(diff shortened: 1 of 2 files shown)


sondreso (approved):
LGTM :+1:

sondreso:
Wait for https://github.com/equinor/libres/pull/688 to merge this

PR #708, opened 2019-10-03
Retry callback into job queue / fix simulation_context

xjules:
- fix sumbit test cound and update status
- test queue retries
- test for queue manager and fix to job queue test
- fix for simulation_context and simulation_batch and the tests which use python job now
- integrates and updates #757 

**Issue**
#686 

**Approach**
It includes changing the job status and making sure number of submits is properly set.

PR #757, opened 2019-10-22
WIP: Use queue manager to kill jobs

ManInFez:
**Issue**
Resolves #my_issue

**Approach**

### DIFF
--- a/python/res/job_queue/job_queue_manager.py
+++ b/python/res/job_queue/job_queue_manager.py
@@ -155,6 +155,15 @@ class JobQueueManager(BaseCClass):
                 started_job_threads.append(job.run(job_queue.driver))
                 job = job_queue.fetch_next_waiting()
             time.sleep(1)
+            if job_queue.stopped_by_user:
+                for job in job_queue.job_list:
+                    job.stop(job_queue.driver)
 
         for thread in started_job_threads:
             thread.join()
+
+        #Cleanup all not-started jobs
+        for job in job_queue.job_list:
+            if job.status == JobStatusType.JOB_QUEUE_WAITING:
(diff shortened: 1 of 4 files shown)


python/res/job_queue/job_queue_manager.py
```diff
@@ -155,6 +155,15 @@ def execute_queue(self):
...
+                for job in job_queue.job_list:
+                    job.stop(job_queue.driver)
 
         for thread in started_job_threads:
             thread.join()
+
+        #Cleanup all not-started jobs
+        for job in job_queue.job_list:
+            if job.status == JobStatusType.JOB_QUEUE_WAITING:
+                job._set_status(JobStatusType.JOB_QUEUE_IS_KILLED)
```
andreabrambilla, line 168: This looks a bit like a hack. I would guess that `job.stop` handles the status change for a job that is running. I don't see why the jobs that are waiting should be treated differently.
If this is really necessary, the comment should be more exhaustive

markusdregi:
This PR has neither an issue nor a description. Please add some more information....

xjules:
> This PR has neither an issue nor a description. Please add some more information....

This PR is now included in #708, can go away after #708 is merged.

python/res/job_queue/job_queue_manager.py
xjules: > This looks a bit like a hack. I would guess that `job.stop` handles the status change for a job that is running. I don't see why the jobs that are waiting should be treated differently.
> If this is really necessary, the comment should be more exhaustive

This's to get the correct status on non started jobs (after being stopped by the user. Agree, will include a comment there.

PR #763, opened 2019-10-24
Removes validation checking that a case has data [Do not merge]

oyvindeide:
This is a temporary solution to make the job run through ERT, as
it reports that cases have no data, even when they do.

### DIFF
--- a/python/res/enkf/jobs/scaling/scaling_job.py
+++ b/python/res/enkf/jobs/scaling/scaling_job.py
@@ -197,7 +197,6 @@ def valid_job(observations, user_config, ensamble_size, storage):
     error_messages.extend(obs_keys_errors)
 
     if obs_keys_present:
-        error_messages.extend(has_data(observations, calculation_keys, ensamble_size, storage))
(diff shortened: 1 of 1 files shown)


markusdregi:
Could you please elaborate on the case when this goes wrong? Are you sure that you have data for all members of the ensemble? In particular this one seems a bit optimistic?

```python
def has_data(observations, keys, ensamble_size, storage):
    """
    Checks that all keys have data and returns a list of error messages
    """
    ....
    active_mask = BoolVector(True, ensamble_size)
    ....
```

oyvindeide:
You're probably right about that, and should make sure this allows for failed realizations. The main problem occurs when running cases through `ERT`, where this function always returns false, even with all realisations succeeding. This was just a stop gap measure to get it working in `ERT` while I figure out why it fails, as it works when running an identical case just through `libres`.

oyvindeide:
You were right about the optimistic checking of realizations. In `ERT` all realizations are not marked as success despite none failing. Closing this with failing realization checking in: https://github.com/equinor/libres/pull/766

PR #766, opened 2019-10-28
Check for missing realizations

oyvindeide:
Checks which realizations are active before validating to account for failed realizations

### DIFF
--- a/python/res/enkf/jobs/scaling/scaling_job.py
+++ b/python/res/enkf/jobs/scaling/scaling_job.py
@@ -11,7 +11,7 @@ from res.enkf.jobs.scaling.scaled_matrix import DataMatrix
 from res.enkf.jobs.scaling.measured_data import MeasuredData
 from res.enkf import LocalObsdata, ActiveList
 from ecl.util.util import BoolVector
-
+from res.enkf import RealizationStateEnum
 
 def scaling_job(ert, user_config_dict):
     """
@@ -219,7 +219,12 @@ def has_data(observations, keys, ensamble_size, storage):
     Checks that all keys have data and returns a list of error messages
     """
     error_msg = "Key: {} has no data"
-    active_mask = BoolVector(True, ensamble_size)
+    active_realizations = storage.realizationList(RealizationStateEnum.STATE_HAS_DATA)
+
+    if len(active_realizations) == 0:
+        return [error_msg.format(key) for key in keys]
+
+    active_mask = BoolVector.createFromList(ensamble_size, active_realizations)
     return [error_msg.format(key) for key in keys if not observations[key].hasData(active_mask, storage)]
 
 
--- a/python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
+++ b/python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
@@ -708,3 +708,44 @@ def test_main_entry_point_block_data_calc():
 
     for index, node in enumerate(obs_vector):
             assert node.getStdScaling(index) == 2.0
+
+
+@pytest.mark.usefixtures("setup_tmpdir")
+def test_validate_failed_realizations():
+    """
+    Config has several failed realisations
+    """
+    test_data_dir = os.path.join(_TEST_DATA_DIR, "local", "custom_kw")
+    shutil.copytree(test_data_dir, "test_data")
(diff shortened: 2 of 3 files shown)


python/res/enkf/export/summary_collector.py
```diff
@@ -9,9 +9,9 @@
 class SummaryCollector(object):
 
     @staticmethod
-    def createActiveList(ert, fs):
+    def createActiveList(ensamble_size, fs):
```
markusdregi: Unless you have a very good reason for doing so, we should not change the current API...

python/res/enkf/jobs/scaling/scaling_job.py
```diff
@@ -219,7 +220,8 @@ def has_data(observations, keys, ensamble_size, storage):
     Checks that all keys have data and returns a list of error messages
     """
     error_msg = "Key: {} has no data"
-    active_mask = BoolVector(True, ensamble_size)
+    active_realizations = SummaryCollector.createActiveList(ensamble_size, storage)
```
markusdregi: It seems a bit arbitrary to get this functionality from the `SummaryCollector`. How about using `storage.realizationList(RealizationStateEnum.STATE_HAS_DATA)` instead?

oyvindeide: That makes more sense :+1:

python/res/enkf/export/summary_collector.py
oyvindeide: There is a reason, if it's good enough is another matter. It's symptomatic for a lot of these functions to take in EnkFMain and use a long chain of commands on it, making it very hard to create new functionality without pulling EnkFMain down to the lowest level, when in this case what was actually needed was an int. Changing that as the functionality was used seemed a rational way to start stabbing at the problem, but might not be worth a breaking change.

markusdregi: I see your reasoning, but let us save this for later... We should first make an evaluation of what we want the API to look like before we start making micro improvements and at the same time breaking our API.

python/res/enkf/jobs/scaling/scaling_job.py
```diff
@@ -219,7 +219,8 @@ def has_data(observations, keys, ensamble_size, storage):
     Checks that all keys have data and returns a list of error messages
     """
     error_msg = "Key: {} has no data"
-    active_mask = BoolVector(True, ensamble_size)
+    active_realizations = storage.realizationList(RealizationStateEnum.STATE_HAS_DATA)
+    active_mask = BoolVector.createFromList(ensamble_size, active_realizations)
```
markusdregi, line 227: Could I have a test that fails on the previous setup and now succeeds? :)

python/res/enkf/export/summary_collector.py
oyvindeide: :+1:

python/res/enkf/jobs/scaling/scaling_job.py
oyvindeide: :+1:

python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
```diff
@@ -708,3 +708,23 @@ def test_main_entry_point_block_data_calc():
...
+    shutil.copytree(test_data_dir, "test_data")
+    os.chdir(os.path.join("test_data"))
+
+    res_config = ResConfig("mini_fail_config")
+    ert = EnKFMain(res_config)
+
+    fs = ert.getEnkfFsManager().getCurrentFileSystem()
+    observations = ert.getObservations()
+
+    assert scaling_job.has_data(observations, ["GEN_PERLIN_1"], ert.getEnsembleSize(), ert.getEnkfFsManager().getCurrentFileSystem()) == []
```
markusdregi: When GitHub activates the horizontal scrolling bar is a good hint that you should perhaps break your line ;)

oyvindeide: Yes, still cleaning that up a little bit, also realized that this allows an ensemble with no data to pass, which it should not, so adding a test for that and a fix.

markusdregi (approved):
Looks good to me :+1: 

Made some issues with further work for this module...

PR #774, opened 2019-10-29
Add verbosity to scaling job

oyvindeide:
Adds output to scaling job, also does some modification to the python logger.

### DIFF
--- a/python/res/enkf/jobs/scaling/scaled_matrix.py
+++ b/python/res/enkf/jobs/scaling/scaled_matrix.py
@@ -51,6 +51,7 @@ class DataMatrix(object):
             len(events.keys), data_matrix.shape[1], nr_components
         )
 
+        print("Scaling factor calculated from {}".format(events.keys))
         return scaling_factor
 
     @staticmethod
@@ -88,4 +89,9 @@ class DataMatrix(object):
             pc is the number of primary components from PCA analysis
                 below a user threshold
         """
+        print(
+            "Calculation scaling factor, nr of primary components: {:d}, number of observations: {:d}".format(
+                nr_components, nr_keys * nr_observations
+            )
+        )
         return np.sqrt(nr_keys * nr_observations / float(nr_components))
--- a/python/res/enkf/jobs/scaling/scaling_job.py
+++ b/python/res/enkf/jobs/scaling/scaling_job.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 
-import sys
 import configsuite
 
 from copy import deepcopy
@@ -44,9 +43,7 @@ def _observation_scaling(ert, config):
     Collects data, performs scaling and applies scaling, assumes validated input.
     """
     obs = ert.getObservations()
-
     measured_data = MeasuredData(ert, config.CALCULATE_KEYS)
-
     matrix = DataMatrix(measured_data.data)
     matrix.std_normalization(config.CALCULATE_KEYS.keys, inplace=True)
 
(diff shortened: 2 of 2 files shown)


oyvindeide:
Could consider setting up the logger slightly differently, currently it just writes to stderr. Ideally the c-logger and the python logger would have duplicate streams so we get the logging in the same place.

python/res/__init__.py
```diff
@@ -30,8 +30,21 @@
...
+logFormatter = logging.Formatter(
+    '%(levelname)s: %(asctime)s %(name)s %(message)s',
+    "%m/%d - %H:%M:%S"
+)
+
+rootLogger = logging.getLogger()
+
+consoleHandler = logging.StreamHandler()
+consoleHandler.setFormatter(logFormatter)
+consoleHandler.setLevel(logging.DEBUG)
```
markusdregi: Are you sure about this one? The `C`-code does the exact opposite:

https://github.com/equinor/libres/blob/59bc1a14b9a18f841395672d7a3e5d5455a5243d/lib/res_util/log.cpp#L100

python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
```diff
@@ -95,6 +95,8 @@ def setup_ert(tmp_path_factory):
                 {ConfigKeys.NAME: "poly_eval", ConfigKeys.PATH: "POLY_EVAL"}
             ],
             ConfigKeys.SIMULATION_JOB: [{ConfigKeys.NAME: "poly_eval"}],
+            ConfigKeys.HOOK_WORKFLOW_KEY: ["obs_scaling", "POST_SIMULATION"],
```
markusdregi: Could you explain why the tests are changed?

oyvindeide: This is an integration test with ErtScript which wasn't tested before. But you're right, doesn't belong in this PR, will create another PR with this, still think it's valuable to have as this wasn't tested at all before.

python/res/__init__.py
oyvindeide: Yes, should be higher than `debug`, would like to set it to `info` for the purpose of this PR, at least temporarily, before this change it was `warning`. Another option is to change to print statements or writing a file, but don't feel like that's a good solution. Ideally I think we would integrate the `python` and `C` logging using the same stream and same levels.

python/tests/res/enkf/jobs/scaling/test_enkf_scaling_job.py
markusdregi: Of definitively! Just was a bit curious as by the title of the PR I would not expect these kind of changes to the tests.

jondequinor:
Why are we not using `res.util.log.Log` here?

oyvindeide:
> Why are we not using `res.util.log.Log` here?

The reason I haven't used it is because it's a wrapper around the C-logging function, and don't think that's the functionality we would like to use in the future, feel it's better to use the built-in Python logging. But we should definitely standardize, currently we have `reslog`, `log` and `logging`.

jondequinor:
The C-logger loads log level, output file, etc from configuration. I think this should be supported in for the logger introduced here as well.

Is `res.util.log` to be deprecated, then I think it should be marked as such.

oyvindeide:
> The C-logger loads log level, output file, etc from configuration. I think this should be supported in for the logger introduced here as well.
> 
> Is `res.util.log` to be deprecated, then I think it should be marked as such.

Agreed, think I'll split out the logging set up from this PR and just do the actual logging.

oyvindeide:
On a more practical note, would it be preferable to have another way to communicate with the user in this job than through `log`? Currently we have three logs, `log`, `reslog` and `logging` from python. Why we have two different implementations in `C` I'm unsure of, any ideas? They share the same default value (warning), and increasing the default to `INFO` (which would be the output from this job), floods `stdout` with output, as only `log` has a fixed threshold for stdout, not `reslog`. To avoid this the user has to set the logging level in the config, or It is of course possible to add another default value for `reslog`. Preferences?

oyvindeide:
Could do something like this: https://github.com/equinor/libres/pull/783
Makes all the loggers log to the same file, which seems to work well, python is in append, so doesn't seem to be any conflicts. Allows setting different level for stderr and file.

markusdregi (approved):
:+1:

PR #777, opened 2019-10-30
Enable black formatting and testing on job_runner

oyvindeide:


.travis.yml
```diff
@@ -60,6 +60,7 @@ before_install:
 
 install:
   - pip install -r requirements.txt
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then pip install black; fi
```
markusdregi: This should be handled in the `setup.py`'s `test_requires` instead...

python/tests/global/test_formatting.py
```diff
@@ -0,0 +1,29 @@
...
+    from click.testing import CliRunner
+
+    root = str(Path(__file__).parent.parent.parent)
+
+    runner = CliRunner()
+    resp = runner.invoke(
+        black.main,
+        [
+            "--check",
+            os.path.join(root, "job_runner"),
```
markusdregi: I would prefer we instead used the `--exclude` flag and explicitly removed all current files not adhering to black formatting. Then new files would immediately be forced to follow the formatting rules.

oyvindeide: That is a better solution, though it looks like the exclude flag is actually a regex, so had to make a workaround.

.travis.yml
oyvindeide: I agree that is a solution for the future, but we don't have a `setup.py` in `libres` though.

python/tests/global/test_formatting.py
oyvindeide: On second thoughts the workaround is causing problems of it's own, so a very explicit regex might be a better solution.

python/tests/global/test_formatting.py
```diff
@@ -0,0 +1,35 @@
...
+    return fnames
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6), reason="requires Python3")
+def test_code_style():
+    from click.testing import CliRunner
+    import black
+    from pathlib import Path
+
+    root = str(Path(__file__).parent.parent.parent)
```
jondequinor: Could you use `tests.conftest.source_root` here? This bars refactoring of this file.

python/tests/global/test_formatting.py
```diff
@@ -0,0 +1,35 @@
...
+
+    excluded_files = get_excluded_files(root)
+    all_py_files = get_py_files(root)
+    files_to_test = set(all_py_files) - set(excluded_files)
+
+    runner = CliRunner()
+
+    resp = runner.invoke(black.main, ["--check"] + list(files_to_test))
+    assert (
```
jondequinor: Maybe just import `unittest`, make a testcase and use `self.assertEqual(resp.exit_code, 0, "msg…")`?

oyvindeide: Because this test is already depending on `pytest`, especially with the `@pytest.mark.skipif(sys.version_info < (3, 6), reason="requires Python3")` above it, I feel it is more consistent to use `assert`, which is the `pytest` way, instead of `unittest`.

python/tests/global/test_formatting.py
oyvindeide: Sure, do you agree with something like this: `root = os.path.join(source_root(), "python")`?

jondequinor: Yes, totally. This means we can move `test_formatting.py` out of `global` without having to edit the contents of the file.

oyvindeide: Agreed, moved it out of `global` now, also allowed me to get rid of the relative import as importing from `global` is a no-go.

.travis.yml
oyvindeide: Fixed in https://github.com/equinor/libres/pull/784 which introduces `test_requirements.txt`

jondequinor (approved):
LGTM

oyvindeide:
retest this please

PR #782, opened 2019-11-05
ecl_run: "Tee" Eclipse output into .LOG and stdout

pinkwah:
**Issue**
Resolves the second half of equinor/ert#532

**Approach**
Use POSIX IO functions to read from the subprocess' stdout and write it into stdout and file as a bytestream.

The `os.read` function blocks if there isn't any data to read but the pipe hasn't been closed, which is the behaviour we want so we don't 100% the CPU. [`man 3 read`]

There is a function in `io.BufferedReader` (ie. superclass of `sys.stdout` and `open(...)`) called `read1` that is the blessed way to do this (and also supports Windows I think), but it doesn't exist in `2.7`, so I have chosen to use `os` directly.

### DIFF
--- a/python/res/fm/ecl/ecl_run.py
+++ b/python/res/fm/ecl/ecl_run.py
@@ -18,10 +18,10 @@ from .ecl_config import EclConfig
 from res.util.subprocess import await_process_tee
 
 
-EclipseResult = namedtuple("EclipseResult" , "errors bugs")
-body_sub_pattern = "(\s^\s@.+$)*"
-date_sub_pattern = "\s+AT TIME\s+(?P<Days>\d+\.\d+)\s+DAYS\s+\((?P<Date>(.+)):\s*$"
-error_pattern    = "^\s@--  ERROR%s$%s"   % (date_sub_pattern, body_sub_pattern)
+EclipseResult = namedtuple("EclipseResult", "errors bugs")
+body_sub_pattern = r"(\s^\s@.+$)*"
+date_sub_pattern = r"\s+AT TIME\s+(?P<Days>\d+\.\d+)\s+DAYS\s+\((?P<Date>(.+)):\s*$"
+error_pattern = r"^\s@--  ERROR{}${}".format(date_sub_pattern, body_sub_pattern)
 
 def make_LSB_MCPU_machine_list( LSB_MCPU_HOSTS ):
     host_numcpu_list = LSB_MCPU_HOSTS.split()
@@ -268,8 +268,8 @@ class EclRun(object):
 
 
     def readECLEND(self):
-        error_regexp = re.compile("^\s*Errors\s+(\d+)\s*$")
-        bug_regexp = re.compile("^\s*Bugs\s+(\d+)\s*$")
+        error_regexp = re.compile(r"^\s*Errors\s+(\d+)\s*$")
+        bug_regexp = re.compile(r"^\s*Bugs\s+(\d+)\s*$")
 
         report_file = os.path.join( self.run_path , "{}.ECLEND".format(self.base_name))
         if not os.path.isfile(report_file):
--- a/python/res/fm/ecl/ecl_run.py
+++ b/python/res/fm/ecl/ecl_run.py
@@ -174,58 +174,35 @@ class EclRun(object):
             for host in machine_list:
                 fileH.write("%s\n" % host)
 
-
-    def execSerialEclipse(self):
+    def execEclipse(self):
         log_name = "{}.LOG".format(self.base_name)
 
-        with open(log_name, 'wb') as log_file:
-            process = subprocess.Popen(
-                [
-                    self.sim.executable,
-                    self.base_name
-                ],
-                env=self.sim.env,
-                stdout=subprocess.PIPE,
-            )
-            return await_process_tee(process, sys.stdout, log_file)
+        with pushd(self.run_path), open(log_name, "wb") as log_file:
+            if not os.path.exists(self.data_file):
+                raise IOError("Can not find data_file:{}".format(self.data_file))
+            if not os.access(self.data_file, os.R_OK):
+                raise OSError("Can not read data file:{}".format(self.data_file))
 
-
-    def execParallellEclipse(self):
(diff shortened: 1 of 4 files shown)


ertomatic:

markusdregi:
Jenkins: Add to whitelist

python/tests/res/fm/test_ecl_run.py
```diff
@@ -188,6 +188,7 @@ def test_run(self):
         ecl_run = EclRun("SPE1.DATA", sim)
         ecl_run.runEclipse( )
 
+        self.assertTrue( os.path.isfile( os.path.join( ecl_run.runPath() , "%s.LOG" % ecl_run.baseName())))
```
markusdregi: Could you extend the tests by checking that there is content in the log files?

python/res/fm/ecl/ecl_run.py
```diff
@@ -173,33 +175,39 @@ def initMPI(self):
...
-        process = subprocess.Popen(
-            [
-                self.sim.executable,
-                self.base_name
-            ],
-            env=self.sim.env
-        )
-        process.communicate()
-        return process.returncode
+        log_name = "%s.LOG" % self.base_name
```
markusdregi: I would prefer new-style string formatting ;)

markusdregi (approved):
I had two minor comments. Besides those this looks very good :+1:

python/res/fm/ecl/ecl_run.py
```diff
@@ -338,3 +346,18 @@ def checkCase(cls , refcase , simcase):
 Simulation %s : %s
 """ % ( refcase , ref.getEndTime() , simcase , sim.getEndTime())
             raise ValueError( msg )
+
+    @staticmethod
+    def _await_process(process, out_file):
```
jondequinor: Should `_await_process` be renamed to also describe its main side effect, i.e. writing to `out_file`

And should you not rather pass an arbitrary amount of `*dest_buffers` and just loop over these as you write?

So `_tee_process_stdout(process, *dest_buffers):`? That would make that little method unit testable. So for instance, in your unit tests you could just pass a bunch of buffers, and in your `execParallellEclipse` you'd pass the file, and `sys.stdout`.

And do you need to read/write an arbitrary `1024` bytes? Usually, these standard streams can be expected to delimit output with newlines.

pinkwah: How about `_await_process_tee`? That would describe the function's main features:
- Block (wait) until process finishes.
- Tee output into stdout and file.

I agree that consumers of the function might expect it to take a list of files, since that's what the `tee` program does. However, at the moment we don't need this functionality, so it's probably unwise to overcomplicate the function.

jondequinor: Agreed!

dotfloat:
retest this please

dotfloat:
Jenkins retest this please

dotfloat:
Jenkins retest this please

pinkwah:
Jenkins retest this please

pinkwah:
Jenkins is happy now that #804 is merged. Can @jondequinor or @markusdregi double-check that everything is good before merging?

I added the following line since the last review:
https://github.com/equinor/libres/blob/a319fa221dbbe182d237c783618776d2b23c2135/python/res/fm/ecl/ecl_run.py#L366

According to https://docs.python.org/3/library/subprocess.html#subprocess.Popen , `Popen` doesn't close the pipes automatically unless `close_fds=True`. Instead I'm closing it after the process has ended.

python/tests/res/fm/test_ecl_run.py
```diff
@@ -326,3 +333,19 @@ def test_error_parse(self):
         self.assertEqual( error_list[0], error0)
         self.assertEqual( error_list[1], error1)
 
+
+    @tmpdir()
+    def test_await_process_tee(self):
+        with open('a', 'wb') as a_fh, open('b', 'wb') as b_fh:
+            process = Popen(['/bin/cat', '/bin/cat'], stdout=PIPE)
```
jondequinor: Are you _catting_ `/bin/cat`? :roller_coaster:

jondequinor (approved):
LGTM

python/tests/res/fm/test_ecl_run.py
```diff
@@ -326,3 +333,19 @@ def test_error_parse(self):
         self.assertEqual( error_list[0], error0)
         self.assertEqual( error_list[1], error1)
 
+
+    @tmpdir()
+    def test_await_process_tee(self):
```
markusdregi: I'm not a fan of testing private functions. Could we replace this by a test that mocks `sys.stdout`?

python/res/fm/ecl/ecl_run.py
```diff
@@ -173,33 +175,39 @@ def initMPI(self):
...
+                    self.base_name
+                ],
+                env=self.sim.env,
+                stdout=subprocess.PIPE,
+            )
+            return self._await_process_tee(process, sys.stdout, log_file)
 
 
     def execParallellEclipse(self):
+        log_name = "{}.LOG".format(self.base_name)
```
markusdregi: Should we unify this logic to ensure that parallell and serial is consistent?

python/tests/res/fm/test_ecl_run.py
pinkwah: I'll move the `_await_process_tee` into a new `res.util.subprocess` module as a public function.

python/tests/res/fm/test_ecl_run.py
pinkwah: :sunglasses:

python/res/fm/ecl/ecl_run.py
pinkwah: Unified into `execEclipse`: ef3106eeb494eb120d7dfb86d5e3ca23813cf15e

pinkwah:
Made the changes. Also changed all regex strings into raw strings because Python was rightfully complaining about `\s` not being a valid escape sequence.

I'll squash the commits in a sane way before merging. *

markusdregi (approved):
🚢 it

PR #790, opened 2019-11-08
Add new status for layer to monitor threads and not job process

### LINKED ISSUE #787: Make separate status for job monitoring thread and job process thread
Job queue manager is currently running its monitoring based on the status of the underlying job process. Job queue node also monitors based on this status. The job queue manager should keep track of the status of the node monitor threads. 

builds on #708

ManInFez:
**Issue**
Resolves #787 

Adds a new status layer, as the job_manager should keep track for the monitor threads 

Each job now keeps track of its own monitor thread in order or not having two threads looking at the a same job

### DIFF
--- a/python/res/job_queue/job_queue_manager.py
+++ b/python/res/job_queue/job_queue_manager.py
@@ -19,9 +19,8 @@ Module implementing a queue for managing external jobs.
 """
 from cwrap import BaseCClass
 from res import ResPrototype
-from res.job_queue import Job, JobStatusType, ThreadStatusType
+from res.job_queue import Job, JobStatusType, ThreadStatus
 import time
-import threading
 
 class JobQueueManager(BaseCClass):
     TYPE_NAME = "job_queue_manager"
@@ -128,27 +127,35 @@ class JobQueueManager(BaseCClass):
         else:
             return self.queue.get_max_running()
 
-    def execute_queue(self):
-        job_queue = self.queue
-        
-        while job_queue.is_active():
-            job = job_queue.fetch_next_waiting()
-            while not job_queue.stopped and job is not None and job_queue.count_running() < self.max_running():
-                job.run(job_queue.driver, max_submit=job_queue.max_submit)
-                job = job_queue.fetch_next_waiting()
+    def _available_capacity(self):
+        return not self.queue.stopped and self.queue.count_running() < self.max_running()
+
+    def _launch_jobs(self):
+        while self.queue.is_active() and not self.queue.stopped:
+            #Start waiting jobs
+            while self._available_capacity():
+                job = self.queue.fetch_next_waiting()
+                if job is None:
+                    break
+                job.run(self.queue.driver, max_submit=self.queue.max_submit)
             time.sleep(1)
-            if job_queue.stopped:
-                for job in job_queue.job_list:
-                    job.stop()
 
-        for job in job_queue.job_list:
-            job.wait_for()
-            if job.thread_status == ThreadStatusType.THREAD_FAILED:
-                raise AssertionError("Unexpected job status type after running job: {}".format(job.status))
+    def _stop_jobs(self):
+        for job in self.queue.job_list:
+            job.stop()
+        while self.queue.is_active():
+            time.sleep(1)
 
-        #clean-up to get the correct status on non started jobs after being stopped by user
-        if job_queue.stopped:
-            for job in job_queue.job_list:
-                if job.status != JobStatusType.JOB_QUEUE_DONE:
-                    job._set_status(JobStatusType.JOB_QUEUE_FAILED)
+    def _assert_complete(self):
+        for job in self.queue.job_list:
+            if job.thread_status != ThreadStatus.DONE:
+                msg = "Unexpected job status type after running job: {} with thread status: {}"
+                raise AssertionError(msg.format(job.status, job.thread_status))
 
+    def execute_queue(self):
+        self._launch_jobs()
+
+        if self.queue.stopped:
+            self._stop_jobs()
+
+        self._assert_complete()
--- a/python/res/job_queue/job_queue_manager.py
+++ b/python/res/job_queue/job_queue_manager.py
@@ -19,8 +19,9 @@ Module implementing a queue for managing external jobs.
 """
 from cwrap import BaseCClass
 from res import ResPrototype
-from res.job_queue import Job, JobStatusType
+from res.job_queue import Job, JobStatusType, ThreadStatusType
 import time
+import threading
 
 class JobQueueManager(BaseCClass):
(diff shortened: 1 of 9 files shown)


python/res/job_queue/queue.py
```diff
@@ -260,20 +260,16 @@ def getJobStatus(self, job_number):
         int_status = self._get_job_status(job_number)
         return JobStatusType( int_status )
 
-    def is_running(self):
+    def is_active(self):
```
xjules, line 263: have you tested whether `is_alive` was enough, but basically it's fine

python/tests/res/job_queue/test_job_queue.py
```diff
@@ -77,70 +76,67 @@ def test_kill_jobs(self):
...
+                lambda: self.assertFalse(job_queue.is_active())
             )
             
             for job in job_queue.job_list:
                 assert job.status == JobStatusType.JOB_QUEUE_IS_KILLED
             
-            for t in threads:
-                t.join()
+            for job in job_queue.job_list:
+                job.thread.join()
```
xjules: is this necessary now?

xjules (changes requested):
Just a small comment on whether we need join after not `queue.is_active()`

python/res/job_queue/job_queue_node.py
```diff
@@ -39,7 +39,10 @@ def __init__(self,job_script, job_name, run_path, num_cpu,
         argc = 1
         argv = StringList()
         argv.append(run_path)
-        self.started = False
+
+        self.thread_status = ThreadStatusType.THREAD_READY
```
markusdregi: This should be exposed as a read-only property.

python/res/job_queue/job_queue_node.py
```diff
@@ -39,7 +39,10 @@ def __init__(self,job_script, job_name, run_path, num_cpu,
...
+        self.thread = None
```
markusdregi: Could we make the `thread` an implementation detail by extending the `JobQueueNode` interface with a `join` function?

python/res/job_queue/job_queue_node.py
```diff
@@ -90,21 +93,32 @@ def _job_monitor(self, driver, max_submit):
...
 
+        self.thread_status = ThreadStatusType.THREAD_DONE
+
     def run(self, driver, max_submit=2):
-        self.started = True
-        x = Thread(target=self._job_monitor, args=(driver, max_submit))
-        x.start()
-        return x
+        # Prevent multiple threads working on the same object
+        if self.thread is not None and self.thread.is_alive():
```
markusdregi: Is the later condition necessary?

python/res/job_queue/job_queue_node.py
```diff
@@ -90,21 +93,32 @@ def _job_monitor(self, driver, max_submit):
...
+            self.thread.join()
+
+        self.thread_status = ThreadStatusType.THREAD_RUNNING
+
+        try:
+            self.thread = Thread(target=self._job_monitor, args=(driver, max_submit))
+            self.thread.start()
+        except: # Only except for Thread exceptions...
```
markusdregi: Please address comment

python/res/job_queue/job_queue_node.py
ManInFez: unless the thread has been started, you should not wait for it... could, in theory, lock the whole thing.

markusdregi: I agree :+1:

python/res/job_queue/job_queue_node.py
ManInFez: :+1:

python/res/job_queue/job_queue_node.py
ManInFez: :+1:

python/tests/res/job_queue/test_job_queue.py
ManInFez: it is at least a safe guard to ensure all spawned threads are finished

python/res/job_queue/queue.py
ManInFez: This will not be enough, as it will not include jobs ready to be started.

python/res/job_queue/job_queue_node.py
```diff
@@ -83,32 +90,55 @@ def _job_monitor(self, driver, max_submit):
         while self.is_running():
             time.sleep(1)
             self.update_status(driver)
-            if self._stopped:
+            if self._thread_status == ThreadStatusType.THREAD_STOPPING:
```
xjules: why not just THREAD.STOP ?

python/res/job_queue/job_queue_node.py
```diff
@@ -83,32 +90,55 @@ def _job_monitor(self, driver, max_submit):
...
+    def wait_for(self):
+        if self._thread is not None and self._thread.is_alive():
+            self._thread.join()
 
     def update_status(self, driver):
         if self.status != JobStatusType.JOB_QUEUE_WAITING:
             self._update_status(driver)
+
+    def _set_thread_status(self, new_status):
+        self._mutex.acquire()
```
xjules: is this necessary? Shouldn't all assignments be atomic?

python/res/job_queue/queue.py
xjules: ok

xjules (approved):
Good job - looks robust enough now!

python/res/job_queue/job_queue_manager.py
```diff
@@ -122,27 +123,29 @@ def __repr__(self):
         return 'JobQueueManager(waiting=%d, running=%d, success=%d, failed=%d, %s)' % (nw,nr,ns,nf,ir)
 
     def max_running(self):
-        return self.queue.get_max_running()
+        return max(1, self.queue.get_max_running())
```
markusdregi: Please elaborate.

ManInFez: When running a single test run, a local copy of QueueConfig is made, and used to create the JobQueue. The local copy does not include the queue options set, which basically is MAX_RUNNING. Thus max running is set to 0 in the driver. 

The old system has an awkward workaround, when max running was 0:

```
  int max_running = job_queue_get_max_running(queue);
  int num_submit_new = util_int_min(max_submit,  max_running - total_active);

  // If max_running == 0 that should be interpreted as no limit; i.e. the queue
  // layer will attempt to send an unlimited number of jobs to the driver - the
  // driver can reject the jobs.
  if (max_running == 0)
    num_submit_new = util_int_min(max_submit,
                                  job_queue_status_get_count(queue->status,
                                                             JOB_QUEUE_WAITING));
```

I think we can reiterate over this when we define the API for the queue system, but for now I think 1 is better than unlimited.

PR #791, opened 2019-11-08
Fix attribute error in rms_run.py os.stat result

joakim-hove:
There is a minor syntax error when the rms job is checking the stat output of a target file.

The current rms testing is done with a very basic mock script which does not capture this behaviour at all, it was therefor quite nontrivial to make a proper test for this fix. Since the current code accesses attributes from standard module `os.stat()` incorrectly it should be quite trivial though; and I hope this can be merged.

### DIFF
--- a/python/res/fm/rms/rms_run.py
+++ b/python/res/fm/rms/rms_run.py
@@ -47,8 +47,7 @@ class RMSRun(object):
                 self.target_file = os.path.join(os.getcwd(), target_file)
 
             if os.path.isfile(self.target_file):
-                st = os.stat(self.target_file)
-                self.target_file_mtime = st.mtime
+                self.target_file_mtime = os.path.getmtime(self.target_file)
             else:
                 self.target_file_mtime = None
 
@@ -129,8 +128,7 @@ class RMSRun(object):
         if self.target_file_mtime is None:
             return
 
-        st = os.stat(self.target_file)
-        if st.mtime == self.target_file_mtime:
+        if os.path.getmtime(self.target_file) == self.target_file_mtime:
             raise Exception("The target file:{} is unmodified - interpreted as failure".format(self.target_file))
 
 
--- a/python/tests/res/fm/rms
+++ b/python/tests/res/fm/rms
@@ -3,14 +3,37 @@
 import json
 import sys
 import os
+import time
 
 config = {"exit_status" : 0}
 if os.path.isfile("action.json"):
     config = json.load(open("action.json"))
 
+
+def write_target_file(target_file):
+    with open(target_file, "w") as f:
+        f.write("this is the target file {}\n".format(time.time()))
+
+
 if "target_file" in config:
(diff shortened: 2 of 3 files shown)


markusdregi:
I agree with the code changes. I believe the following test should do the trick:

```python
    def test_run_class_with_existing_target_file(self):
        with TestAreaContext("test_run_existing_target"):
            with open("rms_config.yml", "w") as f:
                f.write("executable:  {}/bin/rms".format(os.getcwd()))

            os.mkdir("run_path")
            os.mkdir("bin")
            os.mkdir("project")
            shutil.copy(os.path.join(self.SOURCE_ROOT, "python/tests/res/fm/rms"), "bin")
            self.monkeypatch.setenv("RMS_SITE_CONFIG", "rms_config.yml")

            target_file = os.path.join(os.getcwd(), "rms_target_file")
            action = {
                "exit_status": 0,
                "target_file": target_file,
            }
            with open("run_path/action.json", "w") as f:
                f.write(json.dumps(action))

            with open(target_file, "w") as f:
                f.write("This is a dummy target file")

            r = RMSRun(0, "project", "workflow", run_path="run_path", target_file=target_file)
            r.run()
```

joakim-hove:
Thank you for the suggested test - I have updated the PR with your suggested test. Would be cool if 7dc50a2 was also cherry picked to the 2.6 release?

ertomatic:
No problem 👍 

We will definitively get it into the 2.6 release. Thanks for the fix :)

ertomatic (approved):
LGTM.

Please squash your commits before merging.

markusdregi:
No problem 👍

We will definitively get it into the 2.6 release. Thanks for the fix :)

joakim-hove:
> We will definitively get it into the 2.6 release. Thanks for the fix :)

Great - thank you.

> Please squash your commits before merging.

I will happily squash of course - I did keep it in two commits to simplify cherry-picking the actual fix - your call?

markusdregi:
I think I prefer a squash ;)

joakim-hove:
> I think I prefer a squash ;)

Squash it is; observe that one additional change (a `time.sleep(1)` has been introduced in the mock rms binary).

markusdregi:
Observe that it is only the macOS runs that are failing... Maybe that could be a hint?

joakim-hove:
> Observe that it is only the macOS runs that are failing... Maybe that could be a hint?

Ehhh - no not really; I am a bit at a loss here :-(

markusdregi:
> > Observe that it is only the macOS runs that are failing... Maybe that could be a hint?
> 
> Ehhh - no not really; I am a bit at a loss here :-(

That appears to be the situation does it not? At least when I open the travis builds it is only the macOS builds that fails?

markusdregi:
I had an idea about this being related to that `macOS`'s `tmp`-folder really just being a symlink. That has occurred as an issue before. Unfortunately, it did not seem to be the case this time...

PR #798, opened 2019-11-15
Restructure pca matrix

### LINKED ISSUE #759: Allow different data types as input to PCA-scaling
Different data types will have different number of data points.

### LINKED ISSUE #796: Restructure input for PCA calculation in scaling job
Input to PCA-scaling should be restructured such that all notion of time and space is flattened, the input matrix should be structured as individual points on the x-axis and realizations on the y-axis.

oyvindeide:
**Issue**
Resolves #796 
Resolves #759

### DIFF
--- a/python/res/enkf/jobs/scaling/measured_data.py
+++ b/python/res/enkf/jobs/scaling/measured_data.py
@@ -9,8 +9,10 @@ from res.enkf.plot_data import PlotBlockDataLoader
 
 
 class MeasuredData(object):
-    def __init__(self, ert, events):
-        self.data = self._get_data(ert, events.keys)
+    def __init__(self, ert, keys):
+        self.data = self._get_data(ert, keys)
+
+    def remove_nan_and_filter(self, events):
         self.filter_on_column_index(events.index)
         self.remove_nan(events.keys)
         self.filter_out_outliers(events)
@@ -27,7 +29,7 @@ class MeasuredData(object):
         self.data = self._filter_on_column_index(index_list)
 
     def filter_out_outliers(self, events):
-        self.data = self._filter_out_outliers(events)
+        self.data = self._filter_out_outliers(events.keys, events.std_cutoff, events.alpha)
 
     def _get_data(self, ert, observation_keys):
         """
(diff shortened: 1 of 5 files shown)


python/res/enkf/jobs/scaling/measured_data.py
```diff
@@ -9,58 +9,79 @@
...
         Adds simulated and observed data and returns a dataframe where ensamble members will
-        have a data key, observed data will be prefixed with OBS_ and observated standard
-        deviation will be prefixed with STD_.
+        have a data key, observed data will be named OBS and observated standard
+        deviation will be named STD.
         """
         measured_data = pd.DataFrame()
         case_name = ert.getEnkfFsManager().getCurrentFileSystem().getCaseName()
-        for key in observation_keys:
+        for key, index_list in zip(observation_keys, index_lists):
```
jondequinor, line 48: Maybe comment on why zip is used here?

python/res/enkf/jobs/scaling/measured_data.py
```diff
@@ -202,30 +223,32 @@ def _combine_filters(filters):
...
-        return data.set_index([[observation_key] * len(data)])
+        return data.set_index(data.index.values)
 
     @staticmethod
-    def _add_summary_observations(ert, data_key, observation_key, case_name):
+    def _add_summary_observations(ert, data_key, case_name):
         data = SummaryObservationCollector.loadObservationData(
             ert, case_name, [data_key]
         ).transpose()
         data = data.set_index(
```
jondequinor, line 236: I think an explanation would help here. Why are you doing this?

jondequinor (approved):
Could you elaborate on your approach here? I am having trouble connecting the linked issues to the changeset.

Besides the things I've asked for comments on, it looks OK!

python/res/enkf/jobs/scaling/measured_data.py
oyvindeide: The change introduced here will allow the user to input an index list for each key, which wasn't possible before as a common was given for all keys. The calling function should make sure that an index list (could be None) exists for each key.

python/res/enkf/jobs/scaling/measured_data.py
oyvindeide: Will add it to the code :+1:

python/res/enkf/jobs/scaling/measured_data.py
oyvindeide: :+1:

PR #799, opened 2019-11-18
Remove schedule as history 795

### LINKED ISSUE #754: Remove applications
Remove all applications in the applications folder and the corresponding c-code only used by these applications. Also move the build options in cmake-files.

### LINKED ISSUE #795: Remove SCHEDULE as HISTORY_SOURCE
It should not be possible to set `HISTORY_SOURCE` as `SCHEDULE` due to known problems. We should print that this feature is not available and quit.

jondequinor:
**Issue**
Resolves #795 
Resolves #754  

**Approach**
- Removed the non-config code that enables `SCHEDULE` as `HISTORY_SOURCE`
- Added tests that ensure that this fails in a good way
- Left the relevant config code intact, since it will just fail poorly if removed (and `SCHEDULE` is provided as `HISTORY_SOURCE`).


jondequinor:
jenkins test this please

ManInFez:
test this please

python/tests/res/enkf/test_model_config.py
```diff
@@ -163,3 +163,15 @@ def test_model_config_dict_constructor(self):
...
+
+    def test_schedule_file_as_history_is_disallowed(self):
+        case_directory = self.createTestPath("local/configuration_tests")
+        with TestAreaContext("test_constructor") as work_area:
+            work_area.copy_directory(case_directory)
+            with self.assertRaises(ValueError) as cm:
+                ResConfig(
+                    user_config_file="configuration_tests/sched_file_as_history_source.ert"
+                )
+            self.assertIn(
```
ManInFez: shouldnt this be on the same indentation level?

python/res/enkf/model_config.py
```diff
@@ -76,6 +76,10 @@ def __init__(self, data_root, joblist, last_history_restart, refcase,
             raise ValueError(
                 "Error: Unable to create ModelConfig with multiple config objects")
 
+        hist_src_enum = self._get_history_src_enum(config_dict, config_content)
+        if hist_src_enum == HistorySourceEnum.SCHEDULE:
+            raise ValueError("HISTORY_SOURCE as SCHEDULE is not supported")
```
ManInFez: use ConfigKes.HISTORY_SOURCE and History.get_source_string in exception string?

python/tests/res/enkf/test_model_config.py
```diff
@@ -163,3 +163,15 @@ def test_model_config_dict_constructor(self):
...
+                "HISTORY_SOURCE as SCHEDULE is not supported", str(cm.exception)
```
ManInFez: use ConfigKes.HISTORY_SOURCE and History.get_source_string in exception string?

ManInFez:
If you are removing all the "get enum from string / get string from enum" for "SCHEDULE", we should also remove the actual enum. Otherwise we should use the functions for getting the strings for the exception values  :)

markusdregi (commented):
I agree with Åsmund's comments above :)

python/tests/res/enkf/test_model_config.py
jondequinor: No, then any exception raised by `self.assertIn` is eaten by the `with-block`.

jondequinor:
jenkins test this please

ManInFez (approved):
LGTM  :+1:

PR #801, opened 2019-11-18
Use a semaphore to restrict the internalization of data

### LINKED ISSUE #567: Throttle data loading
Examinations in https://github.com/equinor/ert/issues/508 shows that we should only internalize from some forward models at the time. We should support 1000 forward models with 500^3 data points.

ManInFez:
resolves equinor/ert#567

### DIFF
--- a/python/res/job_queue/job_queue_node.py
+++ b/python/res/job_queue/job_queue_node.py
@@ -83,9 +83,9 @@ class JobQueueNode(BaseCClass):
                 self.status == JobStatusType.JOB_QUEUE_RUNNING  or
                 self.status == JobStatusType.JOB_QUEUE_UNKNOWN) # dont stop monitoring if LSF commands are unavailable
     
-    def _job_monitor(self, driver, max_submit):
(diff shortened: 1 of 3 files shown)


ManInFez:
Should probably discuss the number of concurrent threads internalizing data?

python/res/job_queue/job_queue_node.py
```diff
@@ -112,7 +113,7 @@ def _job_monitor(self, driver, max_submit):
 
             self._set_thread_status(ThreadStatus.DONE)
 
-    def run(self, driver, max_submit=2):
+    def run(self, driver, pool_sema, max_submit=2):
```
sondreso, line 116: Could we have the same order of the arguments for this function and _job_monitor?

sondreso (approved):
Other than my comment above and the discussion about how much we should throttle this looks good!

python/res/job_queue/job_queue_node.py
ManInFez: :+1:

markusdregi:
:+1:

PR #804, opened 2019-11-20
Fix reference counting error in enkf_fs

### LINKED ISSUE #737: Unstable tests

mortalisk:
Resolves #737 

enkf_fs_get_ref in enkf_fs.cpp increaments the reference when it probably shouldn't. This caused the counter to be incremented when python created a reference to it, but python would not call the decrement when the python object was decremented, since it was supposedly a reference (that means some sort of weak reference in cwrap language).

The fix currently avoid this by making python use the strong reference functionality in stead, so that it will both increment and decrement the counter correctly.

### DIFF
--- a/python/res/enkf/enkf_fs_manager.py
+++ b/python/res/enkf/enkf_fs_manager.py
@@ -33,7 +33,6 @@ class FileSystemRotator(object):
     def dropOldestFileSystem(self):
         if len(self._fs_list) > 0:
             case_name = self._fs_list[0]
-            fs = self._fs_map[case_name]
             del self._fs_list[0]
             del self._fs_map[case_name]
 
@@ -46,7 +45,7 @@ class FileSystemRotator(object):
 
     def __get_fs(self, name):
         fs = self._fs_map[name]
-        return fs.weakref( )
+        return fs.copy( )
 
     def __getitem__(self, case):
         """ @rtype: EnkfFs """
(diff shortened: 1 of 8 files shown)


python/res/enkf/enkf_fs_manager.py
```diff
@@ -33,7 +33,7 @@ def addFileSystem(self, file_system, full_name):
     def dropOldestFileSystem(self):
         if len(self._fs_list) > 0:
             case_name = self._fs_list[0]
-            fs = self._fs_map[case_name]
+            #fs = self._fs_map[case_name]
```
sondreso: Delete or uncomment

pinkwah:
This PR isn't complete: We should make sure that it isn't possible to use the "weak" EnkfFs in the future. Remove `weakref` from `EnkfFs` and throw an exception in `EnkFs.__init__` if `self.isReference() is True`

The bad `enkf_fs_ref` is used here also:
https://github.com/equinor/libres/blob/149b7f6aea1fa563c1832e86f8356101d1043b95/python/res/enkf/ert_run_context.py#L41-L42

Edit:  after investigation, it seems these two actually are weak references, so ignore this

python/tests/res/enkf/test_enkf_fs_manager1.py
```diff
@@ -38,10 +38,7 @@ def test_create(self):
 
             fs2 = fsm.getFileSystem("newFS")
             self.assertEqual(2, fsm.getFileSystemCount())
-            self.assertEqual(1, fs2.refCount())
-
-            with self.assertRaises(AssertionError):
-                fs2.umount( )
+            #self.assertEqual(1, fs2.refCount())
```
pinkwah: Still have to do something about these asserts.

mortalisk: I deleted those. I think it is sufficiently tested in test_enkf_fs_manager2.py

markusdregi (approved):
You can merge this to the tunes of _Simply the best_ :musical_note:

jondequinor (approved):
LGTM

PR #806, opened 2019-11-20
Add ThreadPoolExecutor to workflowrunner

### LINKED ISSUE #802: Prevent leaking threads in workflow runner
workflow_runner needs someway of managing threads, perhaps with context manager. In order for our test environment to be sane, we need all spawned threads to be manageable.  Wrokflow_runner.run spawns a thread as a demon an exits. Spawning a thread as a demon should not be the only safeguard for it terminate in due time.

xjules:
**Issue**
Resolves #802 

**Approach**
Add ThreadPoolExecutor as an attribute to workflowrunner. 
Exception is provided workflowrunner.exception()

### DIFF
--- a/python/res/job_queue/workflow_runner.py
+++ b/python/res/job_queue/workflow_runner.py
@@ -1,7 +1,6 @@
-from threading import Thread
 from res.job_queue import Workflow
 from res.util.substitution_list import SubstitutionList
-
+from concurrent import futures
 
 class WorkflowRunner(object):
     def __init__(self, workflow, ert=None, context=None):
@@ -20,12 +19,22 @@ class WorkflowRunner(object):
 
         self.__context = context
         self.__workflow_result = None
+        self._workflow_executor = futures.ThreadPoolExecutor(max_workers=1)
+        self._workflow_job = None
+
+    def __enter__(self):
+        self.run()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        self.wait()
 
     def run(self):
-        workflow_thread = Thread(name="ert_gui_workflow_thread")
-        workflow_thread.setDaemon(True)
-        workflow_thread.run = self.__runWorkflow
-        workflow_thread.start()
+        if self.isRunning():
+            raise AssertionError('An instance of workflow is already running!')
+        else:
+            self._workflow_job = self._workflow_executor.submit(self.__runWorkflow)
+
 
     def __runWorkflow(self):
         self.__workflow_result = self.__workflow.run(self.__ert, context=self.__context)
@@ -41,9 +50,16 @@ class WorkflowRunner(object):
     def cancel(self):
         if self.isRunning():
             self.__workflow.cancel()
+        self.wait()
+
+    def exception(self):
+        if self._workflow_job is not None:
+            return self._workflow_job._exception
+        return None
 
     def wait(self):
-        self.__workflow.wait()
+        # This returns a tuple (done, pending), since we run only one job we don't need to use it
+        _, _ = futures.wait([self._workflow_job], timeout=None, return_when=futures.FIRST_EXCEPTION)
 
     def workflowResult(self):
         """ @rtype: bool or None """
--- a/python/tests/res/job_queue/test_workflow_runner.py
+++ b/python/tests/res/job_queue/test_workflow_runner.py
@@ -5,7 +5,11 @@ from tests import ResTest
 from ecl.util.test import TestAreaContext
 from res.util.substitution_list import SubstitutionList
 from .workflow_common import WorkflowCommon
-
+import sys
+if sys.version_info >= (3, 3):
+    from unittest.mock import patch
+else:
+    from mock import patch
 
 class WorkflowRunnerTest(ResTest):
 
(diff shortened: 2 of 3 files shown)


xjules:
test this please!

python/res/job_queue/workflow_runner.py
```diff
@@ -2,6 +2,20 @@
...
+    def __init__(self, target, name):
+        super(WorkflowRunnerThread, self).__init__(target=target, name=name)
+        self.setDaemon(True)
+        self._exception = None
+        # in python2.7 it's called __target therefore we need to inject this too
+        self._target = target
+
+    def run(self):
+        try:
+            if self._target:
```
markusdregi: Should this not be done by `super` instead? I.e. `super(WorkflowRunnerThread, self).run(self)`?

python/res/job_queue/workflow_runner.py
```diff
@@ -2,6 +2,20 @@
 from res.job_queue import Workflow
 from res.util.substitution_list import SubstitutionList
 
+class WorkflowRunnerThread(Thread):
+    def __init__(self, target, name):
+        super(WorkflowRunnerThread, self).__init__(target=target, name=name)
+        self.setDaemon(True)
+        self._exception = None
+        # in python2.7 it's called __target therefore we need to inject this too
```
markusdregi: I don't understand this comment.

markusdregi:
I still don't see how this PR is resolving the given issue. It solves the problem in the tests, but that is more by changing the test, then adopting the code at large (which I find a bit scary).

I'll have try to have a closer look at the usage of the WorkflowRunner and come back to you with something more concrete ;)

python/res/job_queue/workflow_runner.py
```diff
@@ -41,9 +54,18 @@ def isCancelled(self):
...
+
+    @property
+    def runner_thread(self):
+        return self.__workflow_thread
 
     def wait(self):
-        self.__workflow.wait()
+        if self.runner_thread:
+            self.runner_thread.join()
+        print('DBG wait finnished with following exception', self.runner_thread._exception)
```
oysteoh: what happens if **not** self.runner_thread ? from what i understand your print will fail when calling _exception on it?

python/res/job_queue/workflow_runner.py
xjules: The difference of Python2.7 vs Python3.3+ is `thread.__target` vs `thread._target`; therefore we set an attribute `thread._target` (which is tackled by __init__ for Python3.3+)

python/res/job_queue/workflow_runner.py
xjules: Good point, should be inside the if statement

python/res/job_queue/workflow_runner.py
xjules: Let's discuss this - need to know an opinion on this one

ManInFez (approved):
This looks good to me. It fixes the potential leaking of threads during testing. We should look into how Ert is using this and protect the execution in a similar manner.

markusdregi (changes requested):
I would like to have another discussion regarding this PR before we merge it when Julius is back in the office :)

python/res/job_queue/workflow_runner.py
```diff
@@ -20,12 +20,20 @@ def __init__(self, workflow, ert=None, context=None):
...
+
+    def __exit__(self, type, value, traceback):
+        self.wait()
 
     def run(self):
-        workflow_thread = Thread(name="ert_gui_workflow_thread")
-        workflow_thread.setDaemon(True)
-        workflow_thread.run = self.__runWorkflow
-        workflow_thread.start()
+        #should we check if it's already running?
```
markusdregi: If that is intended usage then please go ahead ;)

python/res/job_queue/workflow_runner.py
```diff
@@ -20,12 +20,20 @@ def __init__(self, workflow, ert=None, context=None):
 
         self.__context = context
         self.__workflow_result = None
+        self.__workflow_executor = ThreadPoolExecutor(max_workers=1)
```
markusdregi: Double under scored prefix is reserved for Python. Please use single underscore instead 🙏

markusdregi (commented):
Beautiful 👏 Can you evaluate the usage in _ERT_ and make an issue for changing the usage of `WorkflowRunner` in _ERT_ if that should be done?

python/res/job_queue/workflow_runner.py
```diff
@@ -1,7 +1,7 @@
-from threading import Thread
 from res.job_queue import Workflow
 from res.util.substitution_list import SubstitutionList
-
+from concurrent.futures import ThreadPoolExecutor
+from concurrent import futures
```
markusdregi, line 3: Do we need both of these imports?

markusdregi:
Do we need to update the requirements file and the Komodo dependencies of libres?

ertomatic:

python/res/job_queue/workflow_runner.py
xjules: No, will be gone ... soon :)

xjules:
> Do we need to update the requirements file and the Komodo dependencies of libres?

No, we shouldn't.

python/res/job_queue/workflow_runner.py
xjules: Ok, shall I change the existing ones too?

python/res/job_queue/workflow_runner.py
xjules: ok

python/res/job_queue/workflow_runner.py
xjules: A new issue spawned for existing ones #861

markusdregi:
> > Do we need to update the requirements file and the Komodo dependencies of libres?
> 
> No, we shouldn't.

Are you sure? I don't think `futures` comes with Python 2.7...

xjules:
> > > Do we need to update the requirements file and the Komodo dependencies of libres?
> > 
> > 
> > No, we shouldn't.
> 
> Are you sure? I don't think `futures` comes with Python 2.7...

~~That agrees, but it's included in libecl ... but can make it explicit in libres too.~~ That relates to `future` no `futures`
'futures` included in libres too now. The issue https://github.com/agronholm/pythonfutures/issues/29 only relates to spawning new processes, where in our case we spawn only new threads within the same process, so we are good to go!

python/res/job_queue/workflow_runner.py
markusdregi: You should remove the comment and make a decision as to whether you want to check this or not ;) I would lean towards the later

xjules: Comment removed!

python/res/job_queue/workflow_runner.py
```diff
@@ -20,12 +19,18 @@ def __init__(self, workflow, ert=None, context=None):
...
+
+    def __exit__(self, type, value, traceback):
+        self.wait()
 
     def run(self):
-        workflow_thread = Thread(name="ert_gui_workflow_thread")
-        workflow_thread.setDaemon(True)
-        workflow_thread.run = self.__runWorkflow
-        workflow_thread.start()
+        self._workflow_job = self._workflow_executor.submit(self.__runWorkflow)
```
markusdregi: Are we not leaking a thread now if we call run twice and then cancel arbitrary many times? If so, I would say that we should raise an Exception if `run` is called twice and it is clear misuse of the API.

xjules: The thread should not leak as it is the part of ThreadPool, i.e., just the job submitted twice - of-course it shouldn't be done for different reasons, so yes we can raise an exception here.

python/res/job_queue/workflow_runner.py
```diff
@@ -20,12 +19,22 @@ def __init__(self, workflow, ert=None, context=None):
...
+    def __exit__(self, type, value, traceback):
+        self.wait()
 
     def run(self):
-        workflow_thread = Thread(name="ert_gui_workflow_thread")
-        workflow_thread.setDaemon(True)
-        workflow_thread.run = self.__runWorkflow
-        workflow_thread.start()
+        if self.isRunning():
+            raise Exception('An instance of workflow is already running!')
```
markusdregi: Never raw exceptions, perhaps an `AssertionError`?

xjules: ok. Was browsing around in libres and  found raw Exception still ....

markusdregi (approved):
Good job 👍 

Can you create a new tag on master listing that a new dependency is added after this is merged and add this to the dependency list for master and the new tag in komodo-releases?

PR #809, opened 2019-11-22
Remove unused code

### LINKED ISSUE #803: Remove dead code
Removing `applications` from `libres`[\[1\]](url) probably increased the code that is _dead_\*. We should remove dead code, as it is a severe hindrance of comprehension and refactoring of the code.

\* Not called internally by a `.so` nor by any Python code, nor any _used_ binaries.

jondequinor:
**Issue**
Resolves #803 

**Approach**
Using [`callcatcher`](https://github.com/caolanm/callcatcher) I compiled `libres` and analyzed the code. I found the following symbols to be unused outside of testing (i.e. they are not used anywhere else, not even python code):
https://gist.github.com/jondequinor/3090934452e598de9925425916351543

I then found the following symbols to be completely unused, even in testing:
https://gist.github.com/jondequinor/6ac142d1a1bcaa84b5926f388795bd13

I then summarily eliminated that code.


ManInFez (approved):
Well done!

PR #814, opened 2019-11-25
add index bounds check in gen_obs_update_std_scale

mortalisk:
Add util_abort if index is out of bounds in gen_obs_update_std_scale


oyvindeide (approved):
LGTM

PR #816, opened 2019-11-25
Run rng_init on when running experiment

ManInFez:
Resolves equinor/scout#43

### DIFF
--- a/python/res/enkf/enkf_simulation_runner.py
+++ b/python/res/enkf/enkf_simulation_runner.py
@@ -34,7 +34,6 @@ class EnkfSimulationRunner(BaseCClass):
     def runSimpleStep(self, job_queue, run_context):
         """ @rtype: int """
         #### run simplestep ####
-        #self._enkf_main().initRun(run_context)
 
         if run_context.get_step():
             ecl_config = self._enkf_main().ecl_config.assert_restart()
--- a/python/res/enkf/enkf_simulation_runner.py
+++ b/python/res/enkf/enkf_simulation_runner.py
@@ -34,7 +34,7 @@ class EnkfSimulationRunner(BaseCClass):
     def runSimpleStep(self, job_queue, run_context):
         """ @rtype: int """
         #### run simplestep ####
-        self._enkf_main().initRun(run_context)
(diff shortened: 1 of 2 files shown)


joakim-hove:
```C++
if (enkf_main->rng_manager)
    rng_manager_free(enkf_main->rng_manager);
```
and similar for the other argument.

python/res/enkf/enkf_simulation_runner.py
```diff
@@ -32,7 +32,7 @@ def _enkf_main(self):
     def runSimpleStep(self, job_queue, run_context):
         """ @rtype: int """
         #### run simplestep ####
-        self._enkf_main().initRun(run_context)
+        #self._enkf_main().initRun(run_context)
```
markusdregi: ????

ManInFez: whops.. pushed the removal to the wrong branch

markusdregi:
Can you please explain the idea behind the fix? In particular, I don't understand why you are removing the init call from the Python code...

Also, this is a code change without a test. Always scary! I get that this might be a bit difficult to test as of now, but could you please investigate the feasibility of testing and at the very least make an issue?

ManInFez:
The init call is already performed in create_runpath in the C-code. This is not an ideal fix, the execution flow should be taken care of properly. As far as I see it
- init_run must be called before create_runpath
- create_runpath must be called before "run pre-simulation workflow" 
- recreating the rng_manager after create_runpath will make things crash as some rng pointers are kept somewhere. 
- "run pre-simulation workflow" must happen before run simple step
 
Maybe we can discuss it tomorrow a bit?

markusdregi (changes requested):
This currently breaks the workflows running the update step.

In addition, this is motivated as a feature in ERT/libres of reproducible parameters even when rerunning. We should make a test verifying that we get this behavior in the correct repo.

ertomatic:

sondreso:
This PR is closed due to inactivity. Feel free to rebase and adapt to the changes that have been introduced (mainly pip installable libres and black formatting), and reopen the PR :smile:

PR #819, opened 2019-11-27
Intruduce max_runtime and runtime in job_queue_node on python layer

### LINKED ISSUE #576: MAX_RUNTIME not working?
Users have reported `MAX_RUNTIME` to not work properly over `lsf`. I suggest to modify the poly case so that the job sleeps for a long time, add `MAX_RUNTIME 2` and see whether we can reproduce.

ManInFez:
resolves https://github.com/equinor/ert/issues/576

### DIFF
--- a/python/res/enkf/enkf_simulation_runner.py
+++ b/python/res/enkf/enkf_simulation_runner.py
@@ -105,6 +105,8 @@ class EnkfSimulationRunner(BaseCClass):
 
     def start_queue(self, run_context, job_queue):
         max_runtime = self._enkf_main().analysisConfig().get_max_runtime()
+        if max_runtime == 0:
+            max_runtime = None
 
         # submit jobs
         for i in range(len(run_context)):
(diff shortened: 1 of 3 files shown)


wouterjdb:
FYI: Also the `STOP_LONG_RUNNING` keyword didn't actually trigger the kill command. Not sure if that is solved with this or that it should be part of this PR at all.

markusdregi:
> FYI: Also the STOP_LONG_RUNNING keyword didn't actually trigger the kill command. Not sure if that is solved with this or that it should be part of this PR at all.

That will be a new issue ;)

markusdregi (approved):
:rocket:

wouterjdb:
FYI: I build a new environment with the latest ERT master and now realizations are nicely being killed again using the `MAX_RUNTIME` keyword. Thanks for the quick response! Moving on to the next iteration as we speak.... without manually killing realizations, so nice.... 🎉

markusdregi:
@wouterjdb Thanks for testing and reporting 👍

PR #821, opened 2019-11-28
Fix missing parent env

### LINKED ISSUE #573: Run OPM as a forward model
There was some issues with binary overloading in PATH in the current stable version, but external users have also reported problems with running flow in fresher builds. Setup a simple setup like SPE1 (see https://github.com/OPM/opm-tests/tree/master/spe1) and run it via ERT.

mortalisk:
Resolves https://github.com/equinor/ert/issues/573

The parent environment was getting overwritten in stead of extended. This should fix that.

### DIFF
--- a/python/res/fm/ecl/ecl_run.py
+++ b/python/res/fm/ecl/ecl_run.py
@@ -128,11 +128,10 @@ class EclRun(object):
 
 
 
-    def initEnv(self):
-        for var,value in self.sim.env.items():
-            os.environ[var] = value
-
-
+    def _build_simulator_environment(self):
+        my_env = os.environ.copy()
+        my_env.update(self.sim.env.items())
+        return my_env
 
     def initMPI(self):
 
@@ -199,7 +198,7 @@ class EclRun(object):
 
             process = subprocess.Popen(
                 command,
-                env=self.sim.env,
+                env=self._build_simulator_environment(),
                 stdout=subprocess.PIPE,
             )
             return await_process_tee(process, sys.stdout, log_file)
--- a/python/tests/res/fm/test_ecl_run.py
+++ b/python/tests/res/fm/test_ecl_run.py
@@ -177,6 +177,165 @@ class EclRunTest(ResTest):
         with self.assertRaises(Exception):
             run(flow_config, ["SPE1.DATA", "--version=no/such/version"])
 
+    @tmpdir()
+    def test_running_flow_given_env_config_can_still_read_parent_env(self):
+        version = "1111.11"
+
+        # create a script that prints env vars ENV1 and ENV2 to a file
+        with open("flow", "w") as f:
+            f.write("#!/bin/bash\n")
+            f.write("echo $ENV1 > out.txt\n")
+            f.write("echo $ENV2 >> out.txt\n")
+        executable = os.path.join(os.getcwd(), "flow")
+        os.chmod(executable, 0o777)
+
+        # create a flow_config.yml with environment extension ENV2
+        conf = {
+            "default_version": version,
+            "versions": {
+                version: {
+                    "scalar": {
+                        "executable": executable,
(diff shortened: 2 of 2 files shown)


python/res/fm/ecl/ecl_run.py
```diff
@@ -129,10 +129,11 @@ def numCpu(self):
 
 
     def initEnv(self):
+        my_env = os.environ.copy()
         for var,value in self.sim.env.items():
```
markusdregi: `dict.update` does exactly the same as this loop ;)

python/res/fm/ecl/ecl_run.py
```diff
@@ -129,10 +129,11 @@ def numCpu(self):
 
 
     def initEnv(self):
```
markusdregi: Should this function get a new name and become private?

mortalisk: I missed this comment. Do you have a suggestion for a new name? Do you mean prefix with _? Just for clarification, what does this "private" actually signify? That it is unsafe to call? That it has side effects (which this one does not)? That we hope no one calls it? Or that we have checked that no one is calling it?

markusdregi: It is mere social convention. If you are interaction with a `_` prefixed function or field in Python that is not a local variable or from `self` you are breaking the social contract between Python programmers and all your thoughts about semver, correctness guaranteed by an API provider etc is out the window.

markusdregi: How about `_build_simulator_environment(self)`?

mortalisk: Ok, so by extension, all functions not prefixed with _ come with a guarantee that we won't change the behavior? If so, we shouldn't rename or make it private or change what it does at all maybe?

sondreso:
Is this PR ready for review? I see the initEnv discussion is not resolved?

python/res/fm/ecl/ecl_run.py
mortalisk: After discussion with @sondreso I conclude that renaming it as you suggest makes sense. _ means that I do not intend anyone outside this module to call it. Lack of _ means that it is intended to be called by other modules in libres. (for some definition of module). No guarantees are attached to either name convention, unless otherwise specified. Lack of _ probably should not mean that the symbol is intended for consumers of libres in general, only internally in the library.

wouterjdb:
How close are you with this PR?

markusdregi:
Can you add a third test that demonstrates that the config takes precedence over the environment? More or less a copy of `test_running_flow_given_env_config_can_still_read_parent_env` where `ENV2` also is set in the environment, but the config value is still the one detected at the end?

markusdregi (approved):
Besides that one lacking test, that the implementation should pass currently, this looks perfect 👍 Well done 👏 Merge at will when the last test is added.

python/tests/res/fm/test_ecl_run.py
```diff
@@ -177,6 +177,165 @@ def test_flow(self):
...
+        sim = flow_config.sim()
+        flow_run = EclRun("DUMMY.DATA", sim)
+        flow_run.runEclipse()
+
+        # assert that the script was able to read both the variables correctly
+        with open("out.txt") as f:
+            lines = f.readlines()
+
+        self.assertEqual(len(lines), 2)
+        self.assertEqual(lines[0].strip(), "OVERWRITTEN1")
```
markusdregi, line 337: Just what I had in mind ;)

mortalisk: Good. Then I'll merge when checks pass

mortalisk:
Jenkins retest this please

wouterjdb:
Tested and confirmed working for me. 👍

markusdregi:
@wouterjdb Thanks for reporting and testing :)

PR #822, opened 2019-11-28
Fix job status regression

### LINKED ISSUE #823: Regression job status
Fetching the status of a job from the `BatchContext` after a job has failed will give an exception instead of returning that the job failed. This is a severe regression in the functionality of the API.

markusdregi:
Resolves: #823

I've added a test that fails on the specified regression. It runs a case with 10 realisations where all the odd realisation members fail and the even succeed. In addition, they all sleep `ensemble_index/2` seconds to ensure some time running before they terminate. This is for sure not a unittest, but it tests a crucial part of our API that was apparently untested.

I've tried checking all places utilising the `ErtRunContext` in both _libres_ and _ERT_ and have not been able to find any places depending on a failed realisation returning `None` instead of the `RunArg`, instead they check whether the index is active, as they should.

I'll try to mock the `JobQueue` to speed up the test speed as it currently runs for 22 seconds on my local computer.


markusdregi:
Update: I was hoping to mock the `JobQueue` and hence never initiate the real queue system. I suspect this would yield a significant decrease in test time. However, due to the switches between Python and C++ I've not managed to propagate the mock all the way down in the call stack to where it would be useful.

I tried reducing the sleep time and the ensemble size to see if that would give anything, but the time reductions was very small...

python/tests/res/simulator/test_batch_sim.py
```diff
@@ -409,6 +411,58 @@ def test_stop_sim(self):
             runpath = 'storage/batch_sim/runpath/%s/realisation-0' % case_name
             self.assertTrue(os.path.exists(runpath))
 
+    @tmpdir()
+    def test_ctx_status_failing_jobs(self):
+        def assertQueueStatus(status, idx, final_state_only=False):
```
ManInFez: is this function actually doing any assertions?

markusdregi: Good catch. It ended up in between implementations, one where the checking was done outside and one where it was done inside the function. I'll fix it!

ManInFez:
It is being used in the progress updating in BaseRunModel in ert https://github.com/equinor/ert/blob/d9597c033bc5de8ff551908a4481fb579bc33bd6/ert_shared/models/base_run_model.py#L216

markusdregi:
> It is being used in the progress updating in BaseRunModel in ert https://github.com/equinor/ert/blob/d9597c033bc5de8ff551908a4481fb579bc33bd6/ert_shared/models/base_run_model.py#L216

How did I miss that one? Thanks!

Will make a corresponding PR in ERT then ;)

python/res/simulator/simulation_context.py
```diff
@@ -30,6 +30,11 @@ def __init__(self, ert, sim_fs, mask, itr, case_data):
         self._ert.getEnkfSimulationRunner().createRunPath(self._run_context)
         self._sim_thread = self._run_simulations_simple_step()
 
+        # Wait until the queue is active before we finish the creation
+        # to ensure sane job status while running
+        while self.isRunning() and not self._queue_manager.isRunning():
```
ManInFez, line 35: is there a case where the queue_manager is running and the thread is not alive?

markusdregi: Most likely not. But there might be a case when neither the thread nor the queue_manager is running and then we should stop trying...

python/tests/res/simulator/test_batch_sim.py
```diff
@@ -409,6 +411,60 @@ def test_stop_sim(self):
             runpath = 'storage/batch_sim/runpath/%s/realisation-0' % case_name
             self.assertTrue(os.path.exists(runpath))
 
+    def assertContextStatusOddFailures(self, ctx, final_state_only=False):
```
ManInFez: Rename ctx to batch_ctx or equivalent

ManInFez (approved):
:+1: LGTM!

PR #834, opened 2019-12-06
Add new logger module to use for reporting to kibana

### LINKED ISSUE #585: Post to kibana when executing workflow-job which executes an ensemble
We are deprecating the workflow-jobs which executes and ensemble. These jobs has been used for running a series of simulations, which now is possible with using bash-script and the CLI. 

We should report  back to kibana who is actually using this, in order to show them the new path, and lead them into enlightenment.

ManInFez:
Use new module in network-reporting in job_runner

Report use deprecated WF jobs to kibana

resolves: https://github.com/equinor/ert/issues/585

### DIFF
--- a/python/ert_logger/__init__.py
+++ b/python/ert_logger/__init__.py
@@ -1,0 +1,29 @@
+import os, pwd, json, requests
+from copy import deepcopy
+
+LOG_URL = "http://devnull.statoil.no:4444"
+USER = pwd.getpwuid(os.getuid()).pw_name
+APPLICATION_NAME = "ERT"
+BASE_MESSAGE = {
+    "user": USER,
+    "application": APPLICATION_NAME,
+    "komodo_release": os.getenv("KOMODO_RELEASE", "--------"),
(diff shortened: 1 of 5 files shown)


sondreso (approved):
LGTM

python/ert_logger/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+        proxies = {"http": None, "https": None}
+        requests.post(
+            LOG_URL,
+            timeout=3,
+            headers={"Content-Type": "application/json"},
+            data=data,
+            proxies=proxies,
+        )
+    except:  # noqa
+        pass
```
mortalisk, line 29: do we want to pass here?

jondequinor: Not really, but hard to know, but if you catch anything, you need to make sure you've caught them all. Unless—this code will break `job_dispatch`.

jondequinor (approved):
LGTM

PR #836, opened 2019-12-09
Added support for activating previously inactive observations

geirev:
**Issue**
Resolves the issue of introducing newly activated observations during the iteration steps in ies_enkf.  Before we did not include any other observations than those that were active in the first iteration.
Now all active observations are used in each iteration. 

I also fixed a bug in the definition of nrobs in ies_enkf_linalg_extract_active which would crash ert if the situation above would occur. This is a rare situation for reservoir applications but it became actual for Camille's wind application.

**Approach**
Instead of storing only the initial measurement perturbations from iteration 0, in data->E I now allocate a full data->E with rows for all the measurements, and measurement perturbations are copied into their respective rows the first time they are input to the analysis computation. Thus, we now always use all active observations input to IES for in each iteration.  This modification significantly simplified the code.

A test is modified to check if the new algorithm runs smoothly.

### DIFF
--- a/lib/analysis/modules/ies_enkf.c
+++ b/lib/analysis/modules/ies_enkf.c
@@ -55,13 +55,7 @@ extern "C" {
 /* ies_enkf function headers */
 /***************************************************************************************************************/
 void ies_enkf_linalg_extract_active(const ies_enkf_data_type * data,
-                                    const matrix_type * D0,
-                                    const matrix_type * Yin,
-                                    const matrix_type * Rin,
                                     matrix_type * E,
-                                    matrix_type * D,
-                                    matrix_type * Y,
-                                    matrix_type * R,
                                     FILE * log_fp,
                                     bool dbg);
 
@@ -199,9 +193,9 @@ void ies_enkf_updateA( void * module_data,
    ies_enkf_data_type * data = ies_enkf_data_safe_cast( module_data );
    const ies_enkf_config_type * ies_config = ies_enkf_data_get_config( data );
 
+
    int nrobs_msk     =ies_enkf_data_get_obs_mask_size(data); // Total number of observations
    int nrobs_inp     =matrix_get_rows( Yin );                // Number of active observations input in current iteration
-   int nrobs         =nrobs_inp;                             // Number of selected active observations
 
 
    int ens_size_msk  = ies_enkf_data_get_ens_mask_size(data);   // Total number of realizations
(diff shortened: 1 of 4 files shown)


ertomatic:

markusdregi:
Jenkins: Test this please

markusdregi:
Hi Geir!

Thank you for the contribution. I heard rumors of this one coming from Camille ;)

We'll find some time soon to review the changes.

lib/analysis/modules/ies_enkf.c
```diff
@@ -278,26 +275,25 @@ void ies_enkf_updateA( void * module_data,
...
+   matrix_type * Y   = matrix_alloc_copy( Yin );
+   matrix_type * E   = matrix_alloc( nrobs_inp, ens_size );
+   matrix_type * D   = matrix_alloc_copy( Din );
+   matrix_type * R   = matrix_alloc_copy( Rin );
 
 /* Subtract new measurement perturbations              D=D-E    */
-   matrix_inplace_sub(D0,Ein);
+   matrix_inplace_sub(D,Ein);
 
 /* Extract active observations and realizations for D, E, Y, and R    */
```
xjules, line 286: The comment needs an update here

lib/analysis/modules/tests/ies_iteration.cpp
```diff
@@ -265,6 +265,7 @@ matrix_type * swap_matrix(matrix_type * old_matrix, matrix_type * new_matrix) {
 
 void test_deactivate(const char * testdata_file) {
```
xjules: I would suggest a more descriptive test function name, e.g. something like `test_deactivate_observations_and_realizations.` Also in `es_testdata.cpp` there is another `test_deactivate` function. Also I sort of thought that this test is for reactivating too. But maybe it does it and I just don't see it :)

lib/analysis/modules/ies_enkf.c
```diff
@@ -199,9 +193,9 @@ void ies_enkf_updateA( void * module_data,
    ies_enkf_data_type * data = ies_enkf_data_safe_cast( module_data );
    const ies_enkf_config_type * ies_config = ies_enkf_data_get_config( data );
 
+
    int nrobs_msk     =ies_enkf_data_get_obs_mask_size(data); // Total number of observations
```
xjules, line 197: In which way `nrobs_msk` is used? since `nrobs_inp` is used as the number of active observations.

lib/analysis/modules/ies_enkf.c
```diff
@@ -435,79 +430,39 @@ void ies_enkf_updateA( void * module_data,
 *  only the active realizations and observations.
 */
 void ies_enkf_linalg_extract_active(const ies_enkf_data_type * data,
```
xjules, line 432: Function needs a comment update

lib/analysis/modules/ies_enkf_data.c
```diff
@@ -167,16 +167,66 @@ void ies_enkf_data_fclose_log(ies_enkf_data_type * data) {
...
+          }
+          bool_vector_iset(data->obs_mask0,iobs,true);
+       }
+       if ( bool_vector_iget(data->obs_mask,iobs) ){
+          m++;
+       }
+    }
+
+    if (dbg) {
+      int m_nrobs      =util_int_min(obs_size_msk-1, 50);
```
xjules: fix for indentation and whitespaces

xjules (changes requested):
Very nice work that handles observations (re) activation. Just had small comments about naming convention, formatting and testing. Mainly whether we test reactivation too.

geirev:
Thanks. I am on vacation i Costa Rica this week, but will fix next week.

lib/analysis/modules/ies_enkf.c
geirev: The comment is correct. I just removed D0 since it, in the new code is identical to D.

lib/analysis/modules/ies_enkf.c
geirev: The comment is now corrected.

lib/analysis/modules/ies_enkf_data.c
geirev: Done, I think.

lib/analysis/modules/tests/ies_iteration.cpp
geirev: I have changed the function name as suggested.

I am testing reactivation as well. It is a bit tricky since there is no reactivation function. What I do is to start with identical copies testdata and testdata2.  In the first iteration, I remove an observation in testdata2 and compute the update. In the subsequent iterations, I use testdata which includes the initially removed datapoint. I have commented further on this in the test function.

lib/analysis/modules/ies_enkf.c
geirev: nrobs_msk is only used for diagnostics while printing the number of active observations versus the total number of observations.

geirev:
I think I have addressed the comments, rebased to master, squashed commits into one major, and force pushed.

geirev:
Any progress on this one?

xjules:
Hei Geir, thanks for addressing my comments already looking into it.

xjules (approved):
We had a closer look at the observation re/de-activation API and decided to leave it for now. So can be merged as is. Nice piece of work!

oyvindeide:
Jenkins: test this please

PR #837, opened 2019-12-11
Add komodo version file to runpath if komodo is used

sondreso:
**Issue**
Resolves partially https://github.com/equinor/scout/issues/100 

**Approach**
Add `komodo.version` file in runpath if `KOMODO_RELEASE` environment variable is set.


ertomatic:

sondreso:
Jenkins test this please

markusdregi:
I would prefer if we don't communicate this via files and the best would be if we can do it without touching libres. How about using the `KOMODO_VERSION` environment variable in the `komodo_job_dispatch` script to set the komodo version that should be used?

PR #838, opened 2019-12-11
tmpdir fix

### LINKED ISSUE #813: Tests creating files in the source folder
Several tests are creating files in the source folder, despite using `TestAreaContext`, which should put the test in a `tmpdir`, indicating that `TestAreaContext` has a bug. Decorating these functions with 
```
from tests.utils import tmpdir
@tmpdir()
```

puts the files in a `tmpdir` but doesn't fix the underlying problem. Tests in question:
- [x] test_ecl_config_creation from test_enkf_library
- [x] test_create from test_enkf_fs_manager1
- [x] test_rotate from test_enkf_fs_manager_2
- [x] test_with_gen_kw from test_enkf_runpath
- [x] test_without_gen_kw test_enkf_runpath
- [x] test_create2 from test_enkf_fs
- [x] test_transfer_var from test_enkf_transfer_env
- [x] test_simulation_model from test_enkf_sim_model
- [x] test_scale_obs from test_enkf_obs
- [x] testObs from test_enkf_obs
- [x] test_obs_block_scale_std from test_enkf_obs
- [x] test_load_results_manually from test_enkf_load_result_manually
- [x] test_config from test_enkf
- [x] test_run_context from test_enkf
- [x] test_run_context_from_external_folder from test_enkf

related to #742

xjules:
**Issue**
Resolves #813 

**Approach**
The cause is described in details in #813. Using decorator `@tmpdir` provides currently the most convenient solution as the test is wrapped inside a function, which automatically collects all variables when exiting the function.

### DIFF
--- a/python/tests/res/enkf/test_enkf_runpath.py
+++ b/python/tests/res/enkf/test_enkf_runpath.py
@@ -18,6 +18,7 @@ import pytest
 
 from ecl.util.test import TestAreaContext
 from tests import ResTest
+from tests.utils import tmpdir
 from ecl.util.util import BoolVector
 
 from res.enkf import (EnsembleConfig, AnalysisConfig, ModelConfig, SiteConfig,
@@ -39,6 +40,7 @@ class EnKFRunpathTest(ResTest):
     def setUp(self):
         pass
 
+    @tmpdir()
     def test_with_gen_kw(self):
         case_directory = self.createTestPath('local/snake_oil_no_data/')
         with TestAreaContext('test_enkf_runpath', store_area=True) as work_area:
@@ -59,6 +61,7 @@ class EnKFRunpathTest(ResTest):
             rp.load()
             self.assertEqual(len(rp), 1)
 
+    @tmpdir()
     def test_without_gen_kw(self):
         case_directory = self.createTestPath('local/snake_oil_no_data/')
         with TestAreaContext('test_enkf_runpath', store_area=False) as work_area:
--- a/python/tests/res/enkf/export/test_custom_kw_collector.py
+++ b/python/tests/res/enkf/export/test_custom_kw_collector.py
@@ -1,4 +1,5 @@
 from tests import ResTest
+from tests.utils import tmpdir
 from res.test import ErtTestContext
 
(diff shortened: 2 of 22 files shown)


xjules:
test this please!

ertomatic:

python/tests/res/enkf/test_enkf_runpath.py
```diff
@@ -58,7 +60,11 @@ def test_with_gen_kw(self):
             self.assertEqual(len(rp), 0)
             rp.load()
             self.assertEqual(len(rp), 1)
+            run_context = None
+            fs = None
+            main = None
```
sondreso: Is this needed even with the tempdir fix? And if it is, should we maybe use `del run_context` rather than assigning `None`, just to be a little more explicit about what is happening? (https://andreasbergstrom.com/posts/python-del-vs-assign-none/) I think this part also may justify a comment, so we don't accidentally remove it in the future.

python/tests/res/enkf/test_enkf_runpath.py
```diff
@@ -74,3 +80,6 @@ def test_without_gen_kw(self):
             self.assertFileDoesNotExist('snake_oil_no_data/storage/snake_oil_no_gen_kw/runpath/realisation-0/iter-0/parameters.txt')
             self.assertEqual(len(os.listdir('snake_oil_no_data/storage/snake_oil_no_gen_kw/runpath')), 1)
             self.assertEqual(len(os.listdir('snake_oil_no_data/storage/snake_oil_no_gen_kw/runpath/realisation-0')), 1)
+            fs = None
+            run_context = None
+            main = None
```
sondreso: Same as above :arrow_up:

sondreso (approved):
Other than the small comment above this looks good :rocket:

python/tests/res/enkf/test_enkf_runpath.py
xjules: you are right, this `None` is no longer necessary  since `@tmpdir` is in place - so will fix it. As a side note, assigning it to `None` calls via cwrap del automatically.

PR #844, opened 2019-12-13
Update template render job description

lars-petter-hauge:
**Issue**
While working on https://github.com/equinor/ert/issues/426, I found that the job description file should be updated to include an `ARGLIST` in order for the `FORWARD_MODEL` keyword to function. It is possible to run the job as is with the following command:

`SIMULATION_JOB TEMPLATE_RENDER -o outpath -i my_input.json -t tmpl.jinja`, but it isn't possible to forward the arguments for the `FORWARD_MODEL` command.

Update the description slightly as well, and added explicit shortkey arguments.

I have one more question though. There is an explicit check for `parameters.json` (or `enkf.EnkfDefaults.DEFAULT_GEN_KW_EXPORT_NAME` to be exact) for this job to run. So in practice you _can_ add another input file, but it is required with a `parameters.json` as well. Should we have the requirement be optional?

### DIFF
--- a/share/ert/forward-models/old_style/TEMPLATE_RENDER
+++ b/share/ert/forward-models/old_style/TEMPLATE_RENDER
@@ -1,1 +1,2 @@
 EXECUTABLE  ../templating/script/template_render
+ARGLIST -i <INPUT_FILES> -o <OUTPUT_FILE> -t <TEMPLATE_FILE>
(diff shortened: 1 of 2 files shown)


lars-petter-hauge:
It might be that the `ARGLIST` should only be added to the `old_style` jobs, given that the "new style" is supposed to work with `SIMULATION_JOB` instead.

oysteoh (approved):

lars-petter-hauge:
Alright, I'll rebase and the fixup and instead of changing the behaviour in this PR we could consider doing a new PR that handels the `parameters.json` as an optional file. I'll update the docs in ERT accordingly to current behaviour.

ertomatic:

PR #862, opened 2020-01-07
CMake: Use current git tag for libres version

pinkwah:
The behaviour is such that if HEAD is on a tag, then that tag is used exactly. Eg, if HEAD has tag "2.7.1", then the library version in libres will be 2.7.1. However, if HEAD is not on a tag, then git will format the version as f"{last used tag}-{number of commits since last tag}-{current commit's short
sha}". This means that only RES_VERSION_MICRO contains additional git information, and must be stored as a string (which it already is).

Will make a similar PR for libecl if this is merged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,9 +14,9 @@ enable_testing()
 
 #-----------------------------------------------------------------
 
-set( RES_VERSION_MAJOR 2 )   # Remember to update release notes whenever
-set( RES_VERSION_MINOR 7 )   # you change the ERT_VERSION_MINOR or MAJOR
-set( RES_VERSION_MICRO git ) # with "new in Ert Version X.X.X"!
+set( RES_VERSION_MAJOR 0 )
+set( RES_VERSION_MINOR 0 )
+set( RES_VERSION_MICRO "not-available" )
 
 # If the micro version is not integer, that should be interpreted as a
 # development version leading towards version MAJOR.MINOR.0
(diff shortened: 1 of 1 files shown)


ertomatic:

dotfloat:
Will make a similar PR for libecl once this is merged.

oyvindeide:
Jenkins test this please

CMakeLists.txt
```diff
@@ -39,6 +39,19 @@ if(GIT_FOUND)
...
+     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+     OUTPUT_VARIABLE GIT_TAG
+     OUTPUT_STRIP_TRAILING_WHITESPACE
+   )
+
+   string( REGEX REPLACE "^([^.]+)\\.([^.]+)\\.(.+)$" "\\1" RES_VERSION_MAJOR "${GIT_TAG}")
+   string( REGEX REPLACE "^([^.]+)\\.([^.]+)\\.(.+)$" "\\2" RES_VERSION_MINOR "${GIT_TAG}")
+   string( REGEX REPLACE "^([^.]+)\\.([^.]+)\\.(.+)$" "\\3" RES_VERSION_MICRO "${GIT_TAG}")
+
+   message( STATUS "libres version: ${GIT_TAG}" )
```
ManInFez: I think it is also a good time to get rid of the hard-coded version above, and set to N/A or similar if git is not available.

ManInFez:
jenkins test this please!

ManInFez (approved):
nice! :+1:

PR #865, opened 2020-01-09
Allow HOOK_WORKFLOW in site-config

sondreso:
Needed for https://github.com/equinor/ert-statoil/pull/170 and https://github.com/equinor/equilibrium/pull/8


lib/enkf/site_config.cpp
```diff
@@ -86,6 +86,12 @@
    xxx_site counterpart are stored.
  */
 
+
+#define RUN_MODE_PRE_SIMULATION_NAME  "PRE_SIMULATION"
```
jondequinor: Would it be possible to move both these and `hook_manager.cpp`'s `defines` to a shared header? I would expect to find these definitions in `config_keys.hpp`.

jondequinor (changes requested):

lib/enkf/site_config.cpp
sondreso: Agreed :+1:

markusdregi (commented):
The code looks good 👍 However, new functionality with an unchanged test base is not good. Could you write a Python test that makes a site config with a hooked workflow, creates a ResConfig and then validates that the workflow is hooked?

python/tests/res/enkf/test_site_config.py
```diff
@@ -90,4 +91,38 @@ def test_constructors(self):
...
+                """INTERNAL False
+EXECUTABLE echo
+MIN_ARG 1
+"""
+            )
+
+        with open("ECHO_WORKFLOW", "w") as fh:
+            fh.write("ECHO_WORKFLOW_JOB hello")
+
+        os.environ["ERT_SITE_CONFIG"] = site_config_filename
```
markusdregi: I suggest that we use the `monkeypatch` approach @oyvindeide introduced some time back instead: https://github.com/equinor/libres/blob/b2801739be29a652e733f2ae239ea0ec25198683/python/tests/res/fm/test_rms_config.py

python/tests/res/enkf/test_site_config.py
```diff
@@ -90,4 +91,38 @@ def test_constructors(self):
...
+
+        res_config = ResConfig(user_config_file=test_config_filename)
+        self.assertTrue(len(res_config.hook_manager) > 0)
```
markusdregi: An assert on the name of the workflow and an exact length of the workflow would be appreciated :)

python/tests/res/enkf/test_site_config.py
sondreso: :+1:

python/tests/res/enkf/test_site_config.py
sondreso: :+1:

jondequinor (approved):
LGTM

markusdregi (approved):
🚢

PR #866, opened 2020-01-10
runtime starts when job is reported as running by driver

### LINKED ISSUE #621: MAX_RUNTIME doesn't always work
I've observed multiple times now that 1 (or more I think?) simulations are not killed whereas `MAX_RUNTIME` dictates that they should be killed after 300 seconds (in my case).

Example: I'm running a thousand realizations. After a couple of successful iterations only 939 realizations are left. In iteration five, 1 simulation is never killed and keeps on running for ever and ever - stopping progress completely. (This is on LSF - I can force going to the next iteration by running `bkill #####`)

image

jondequinor:
**Issue**
Resolves https://github.com/equinor/ert/issues/621

**Approach**
Start clock when `is_running`

### DIFF
--- a/python/res/job_queue/job_queue_node.py
+++ b/python/res/job_queue/job_queue_node.py
@@ -110,11 +110,13 @@ class JobQueueNode(BaseCClass):
         )
 
     def _job_monitor(self, driver, pool_sema, max_submit):
-        self._start_time = time.time()
+        
         self.submit(driver)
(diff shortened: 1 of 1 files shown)


ManInFez:
jenkins test this please!

ManInFez:
jenkins test this please!

ManInFez (approved):
approve by proxy!

PR #871, opened 2020-01-10
ExternalErtScript: Explicit bytes -> str conversion

pinkwah:
We use the [`codecs`](https://docs.python.org/3/library/codecs.html) library instead of `str(self._stdoutdata, 'UTF-8')` because the latter will throw a UnicodeDecodeError if the output is not UTF-8. Instead, we try our best with the given output and replace all invalid codepoints with � (U+FFFD Replacement Character). This is controlled by the 'replace' argument, with another reasonable option being 'ignore'.

### DIFF
--- a/python/res/job_queue/external_ert_script.py
+++ b/python/res/job_queue/external_ert_script.py
@@ -1,3 +1,4 @@
+import codecs
 import sys
 from subprocess import Popen, PIPE
 from res.job_queue import ErtScript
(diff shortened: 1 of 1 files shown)


sondreso (approved):
Nice!

joakim-hove:
I need this commit - is it in release? Would it be possible to get it backported to the 2.6 branch?

sondreso:
This commit is in the (very) newly released 2.8.a2 release. While this is a alpha release, it contains mostly bugfixes compared to the current 2.6 release. The commit will probably not be backported to 2.6.

joakim-hove:
Ok -thank you. 2.7 was quickly done with?

sondreso:
Yes, we are attempting to have monthly releases (and use semantic versioning) :slightly_smiling_face:

PR #875, opened 2020-01-15
test for reading multi_seeds files in rms job

### LINKED ISSUE #843: Add test for multi_seed_file
There is no test coverage on multi_seed_file in `libres/python/tests/res/fm/test_rms_run.py`, should be added.

xjules:
**Issue**
Resolves #843 

**Approach**
Adding test function that creates `random.seeds` file and runs RMSRun.

### DIFF
--- a/python/tests/res/fm/test_rms_run.py
+++ b/python/tests/res/fm/test_rms_run.py
@@ -27,6 +27,41 @@ from _pytest.monkeypatch import MonkeyPatch
 from res.fm.rms import RMSRun
 import res.fm.rms
 
+from tests.utils import tmpdir
+import pytest
+from tests.conftest import source_root
+
+@tmpdir()
+@pytest.mark.parametrize(
+    "test_input,expected_result",
+    [
+        (0, 422851785),
+        (1, 723121249),
+        (2, 132312123),
+    ],
+)
+def test_run_class_multi_seed(tmpdir, monkeypatch, test_input, expected_result):
+    with open("rms_config.yml", "w") as f:
+        f.write("executable:  {}/bin/rms".format(os.getcwd()))
+
+    os.mkdir("test_run_multi_seed")
+    os.mkdir("run_path")
+    os.mkdir("bin")
+    os.mkdir("project")
+    shutil.copy(os.path.join(source_root(), "python/tests/res/fm/rms"), "bin")
+    monkeypatch.setenv("RMS_SITE_CONFIG", "rms_config.yml")
+
+    action = {"exit_status": 0}
+    with open("run_path/action.json", "w") as f:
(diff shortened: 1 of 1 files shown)


xjules:
Jenkins: test this please!

python/tests/res/fm/test_rms_run.py
```diff
@@ -46,6 +46,42 @@ def test_create(self):
             r = RMSRun(0, "rms", "workflow")
 
 
+    def test_run_class_multi_seed(self):
+        with TestAreaContext("test_run_multi_seed"):
```
oyvindeide: Can you replace `TestAreaContext` with `@tmpdir`?

oyvindeide (commented):
Feel free to disagree with me, but wouldn't it be nice to use pytest's parameterize here, as you are essentially running three tests in one? 
```suggestion
from tests.conftest import source_root
import pytest
@pytest.mark.parametrize(
    "test_input,expected_result",
    [
        (0, 422851785),
        (1, 723121249),
        (2, 132312123),
    ],
)
def test_run_class_multi_seed(tmpdir, monkeypatch, test_input, expected_result):
    tmpdir.chdir()
    with open("rms_config.yml", "w") as f:
        f.write("executable:  {}/bin/rms".format(os.getcwd()))

    os.mkdir("test_run_multi_seed")
    os.mkdir("run_path")
    os.mkdir("bin")
    os.mkdir("project")
    shutil.copy(os.path.join(source_root(), "python/tests/res/fm/rms"), "bin")
    monkeypatch.setenv("RMS_SITE_CONFIG", "rms_config.yml")

    action = {"exit_status": 0}
    with open("run_path/action.json", "w") as f:
        f.write(json.dumps(action))

    seed_file = ["3", "422851785", "723121249", "132312123"]
    with open("run_path/random.seeds", "w") as f:
        f.write("\n".join(seed_file))

    r = RMSRun(test_input, "project", "workflow", run_path="run_path")
    assert r.seed == expected_result
```
Just getting rid of `TestAreaContext` reduces the run time significantly, despite having to run everything three times.

python/tests/res/fm/test_rms_run.py
xjules: Agree on both points. Running tests via parametrize is semantically more closer to what's actually happening.

xjules: ~~Also, I reckon that I would fix `@tmpdir` all other tests in `test_rms_run.py`~~ Should be part of another PR I reckon

xjules: done

oyvindeide (approved):
Good job!

PR #879, opened 2020-01-15
Fix error in triangular distribution

### LINKED ISSUE #878: Incorrect sampling with triangular distributions
The transformation for triangular distribution is incorrect for asymmetric distributions. This can be demonstrated easily by testing with e.g.
`VAR TRIANGULAR 0.5 1.0 5.0`
and plotting the histogram distribution

sondreso:
**Issue**
Resolves #878

### DIFF
--- a/lib/enkf/tests/trans_func.cpp
+++ b/lib/enkf/tests/trans_func.cpp
@@ -40,6 +40,20 @@ void test_triangular() {
   stringlist_free(args);
 }
 
+void test_triangular_assymetric() {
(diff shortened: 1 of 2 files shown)


jondequinor (changes requested):
There's a test called `test_triangular`, I wonder if we could assert that this works as intended?

sondreso:
That's why it's still in progress :wink:

jondequinor (approved):
LGTM

PR #880, opened 2020-01-16
Backport fix of error in triangular distribution 

sondreso:
Backport of #879


jondequinor:
I think it makes sense to bump `CMakeLists.txt` here—we don't expect anything else from libres in 2.6.6.

jondequinor (approved):
LGTM

PR #896, opened 2020-01-28
Add filtering on module to prevent warnings from other modules

oysteoh:


sondreso (approved):
LGTM! Tested locally :robot:

sondreso:
Note: Related to https://github.com/equinor/libecl/issues/679

python/ert/__init__.py
```diff
@@ -1,8 +1,14 @@
 import warnings
 
-warnings.simplefilter('always', DeprecationWarning) # see #1437
```
oyvindeide, line 3: Should probably do this globally then? There is another one in: libres/python/res/__init__.py

oyvindeide: Never mind, didn't see that you'd already done that

PR #898, opened 2020-02-07
Use standard C++ library for generating random distribution

### LINKED ISSUE #894: "E is not finite" - observation with 0.000 variance not removed
I have experienced now multiple times, with slightly different setups, that after an X number of iterations I end up with an error saying that "E is not finite".

`E(552,0) = -nan`

When the following test is performed:

https://github.com/equinor/libres/blob/4e90285db47057096d958decb1f76a126d6bbbfb/lib/enkf/obs_data.cpp#L460

While diving into the update log I found that there are one (or a few) observations where the mean and variance of the simulated data is 0.000.

`224 : WWPR_BR-P-20_28_02_2001                    1.175 +/-          100.000  Active   |             0.000 +/-           0.000`

**Some wild guesses/speculations:** Could it be that the `0.000` values are used in a division and thus lead to this error? Shouldn't these observations have been removed in the first place - is there maybe a different tolerance on removing than the rounding that is used afterwards in a division?

ManInFez:
__bug__:
When sampling a normal distribution, the following function was called: 

`return mean + std * sqrt(-2.0 * log(R1)) * cos(2.0 * pi * R2);`

where `R1` and `R2` where random numbers generated from our RNG. 

On the very rare case that that `R1` was zero, `log(R1)` would return `-infinite`, resulting in  `matrix_assert_finite` failing.

__fix__:
Use c++ standard version for sampling a normal distribution.

resolves #894

### DIFF
--- a/lib/enkf/enkf_util.cpp
+++ b/lib/enkf/enkf_util.cpp
@@ -21,7 +21,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdbool.h>
-
+#include <random>
 #include <ert/util/util.h>
 #include <ert/util/rng.h>
 #include <ert/ecl/ecl_util.h>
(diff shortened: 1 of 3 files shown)


ManInFez:
Test this please!

sondreso (approved):
LGTM :rocket:

PR #910, opened 2020-02-25
Add sdpsoft to jenkins and use sdpsoft git it ci

ManInFez:
**Issue**
Resolves #my_issue

**Approach**

### DIFF
--- a/ci/jenkins/testkomodo.sh
+++ b/ci/jenkins/testkomodo.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -xe
 PROJECT=libres
+RELEASE_PATH=${KOMODO_ROOT}/${RELEASE_NAME}
+GIT=${SDPSOFT}/bin/git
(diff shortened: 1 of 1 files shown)


ci/jenkins/testkomodo.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -xe
 PROJECT=libres
+RELEASE_PATH=${KOMODO_ROOT}/${RELEASE_NAME}
```
oysteoh, line 4: At line 12 where RELEASE_PATH is used - ${RELEASE_NAME} is appended once more. Is this correct?

oysteoh (approved):

ManInFez:
Test this please!

ci/jenkins/testkomodo.sh
jondequinor: Yes, because that's a file in which you'll find a yaml with versions

jondequinor (approved):

PR #915, opened 2020-03-09
Use workflow future to determine running status

### LINKED ISSUE #913: Possible bug in workflow runner
Multiple users report following error when running workflow
```
Traceback (most recent call last):
  File "/prog/res/komodo/2020.02.01-py27/root/lib/python2.7/site-packages/ert_gui/tools/workflows/run_workflow_widget.py", line 141, in workflowFinishedWithFail
    error = self._workflow_runner.workflowError()
  File "/prog/res/komodo/2020.02.01-py27/root/lib/python2.7/site-packages/res/job_queue/workflow_runner.py", line 78, in workflowError
    for error_line in error:
TypeError: 'NoneType' object is not iterable
```

Conversation found here: https://equinor.slack.com/archives/CDPJULWR4/p1582812077085900

jondequinor:
**Issue**
Resolves #913 

**Approach**
Use future status in `isRunning`

### DIFF
--- a/python/res/job_queue/workflow_runner.py
+++ b/python/res/job_queue/workflow_runner.py
@@ -41,7 +41,13 @@ class WorkflowRunner(object):
 
     def isRunning(self):
         """ @rtype: bool """
-        return self.__workflow.isRunning()
+        if self.__workflow.isRunning():
+            return True
+
+        # Completion of _workflow does not indicate that __workflow_result is
+        # set. Check future status, since __workflow_result follows future
+        # completion.
+        return self._workflow_job is not None and not self._workflow_job.done()
 
     def isCancelled(self):
         """ @rtype: bool """
@@ -79,5 +85,3 @@ class WorkflowRunner(object):
             error_message += error_line + "\n"
 
         return error_message
-
-


markusdregi:
Could you please elaborate on why it helps to check the status of the job as well? :)

jondequinor:
> Could you please elaborate on why it helps to check the status of the job as well? :)

Added comment https://github.com/equinor/libres/blob/include-future-in-running-status/python/res/job_queue/workflow_runner.py#L46

python/res/job_queue/workflow_runner.py
```diff
@@ -41,7 +41,14 @@ def __runWorkflow(self):
 
     def isRunning(self):
         """ @rtype: bool """
-        return self.__workflow.isRunning()
+        workflow_running = self.__workflow.isRunning()
```
markusdregi: Would if be easier to read as:

```python
if self.__workflow.isRunning():
    return True
return self._workflow_job is not None and not self._workflow_job.done()
```

or something similar?

jondequinor: Indeed! There are many ways to skin this beast.

markusdregi (approved):
🚢

PR #925, opened 2020-04-02
Ensure integer division when making cpu list

### LINKED ISSUE #922: Odd number of CPU's fail for Eclipse with Python 3
User reported following error:

```
File "/prog/res/komodo/2020.03.rc2-py36/root/lib/python3.6/site-packages/res/fm/ecl/ecl_run.py", line 157, in initMPI
   LSB_MCPU_machine_list = make_LSB_MCPU_machine_list( LSB_MCPU_HOSTS )
 File "/prog/res/komodo/2020.03.rc2-py36/root/lib/python3.6/site-packages/res/fm/ecl/ecl_run.py", line 29, in make_LSB_MCPU_machine_list
   for index in range(len(host_numcpu_list) / 2):
TypeError: 'float' object cannot be interpreted as an integer
```

sondreso:
**Issue**
Resolves #922 

**Approach**
Use explicit integer division when creating cpu list

### DIFF
--- a/python/res/fm/ecl/ecl_run.py
+++ b/python/res/fm/ecl/ecl_run.py
@@ -26,7 +26,7 @@ error_pattern = r"^\s@--  ERROR{}${}".format(date_sub_pattern, body_sub_pattern)
 def make_LSB_MCPU_machine_list( LSB_MCPU_HOSTS ):
     host_numcpu_list = LSB_MCPU_HOSTS.split()
     host_list = []
-    for index in range(len(host_numcpu_list) / 2):
+    for index in range(len(host_numcpu_list) // 2):
         machine = host_numcpu_list[ 2*index ]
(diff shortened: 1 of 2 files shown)


markusdregi:
Could you write a test verifying the error? Since `make_LSB_MCPU_machine_list` is exposed publicly you can just call it directly instead of doing it via an integration test of `EclRun` ;) Although I think the "correct" approach would be to make the function private... But that we can resolve later!

sondreso:
I was going to write a test, but realized that we already have a test that should catch this. The problem is that the equinor specific tests only run on Python 2... I was looking into running the tests on Python 3 as well, but there is some stuff that needs to change. So I will take the suggested approach above now, and then look into running the tests on Python 3 at a later point.

markusdregi (approved):
LGTM 🚢

PR #929, opened 2020-04-07
PyPI libres

### LINKED ISSUE #888: Make res pip installable

### LINKED ISSUE #905: Remove .libecl_version
When https://github.com/equinor/libecl/pull/691 is merged, libecl is pip-installable. We should then remove the `.libecl_version` file and instead specify a lower bound for libecl version in `setup.py`.

pinkwah:
This is a _very_ opinionated PR that replaces the old CMake build pipeline with one that exclusively uses pip/python. Since only the SCOUT team's project rely on `libres`, this is safe. Still, however, we should bump the major version.

To test today, `pip install equinor-libres`.

If this PR is merged, the following PRs must be merged also:
- __ert:__ https://github.com/equinor/ert/pull/1081
- __everest:__ TODO
- __semeio:__ https://github.com/equinor/semeio/pull/240
- __equilibrium:__ TODO
- __subscript:__ TODO
- __upwind:__ TODO
- __fmu-report:__ TODO

---

This PR build on top of the work done for *libecl*, which can be viewed here: https://github.com/equinor/libecl/pull/691

We remove the direct linking of `libres.so` with `libecl.so` in favour of the run-time `RTLD_GLOBAL` linking of `libecl.so` when `ecl` is imported

---

HM Tutorial tested! Follow the steps described in the Jenkins file to set up HM Tutorial for Python 3. Create a fresh venv and `pip install`:
1. Update `pip`, `setuptools`, `
1. This PR
2. ERT
3. Subscript

And have fun.

---

Resolves #888 
Resolves #905


python/tests/job_runner/test_job_dispatch.py
```diff
@@ -66,36 +68,31 @@ def test_terminate_jobs(self):
...
             )
+        os.chmod("setsid", 0o755)
 
-        current_dir = os.path.realpath(os.curdir)
+        job_dispatch_script = importlib.util.find_spec("job_runner.job_dispatch").origin
         job_dispatch_process = Popen(
             [
-                os.path.realpath("job_dispatch_executer"),
+                os.getcwd() + "/setsid",
+                sys.executable,
```
sondreso, line 93: Why is `sys.executable` added here?

pinkwah: The original creates a program specifically for this test. However, `/usr/bin/setsid` exists on Linux, which `exec`s its arguments and allowed me to delete a few lines of confusing code in that test. But then it turned out that macOS didn't have this utility, so I wrote one in Python. Mostly just shuffling code around tbh.

The reason I'm using `sys.executable` is because I have removed the executable bit from `job_dispatch.py` and removed the shebang, so I have to specify that it should be run with the current Python executable.

Imo, it's a change for the better. Previously it was `#!/usr/bin/env python`, but now we're being explicit about which Python binary we should use.

sondreso: But `job_dispatch.py` is called directly in the wild (for example when submitting to lsf), so we need to shebang, or no? :thinking:

.github/workflows/python-bdist.yml
```diff
@@ -0,0 +1,135 @@
+name: Python Binary Distribution
+
+on: [push]
```
sondreso, line 3: Think we need to add [release](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release) as well

python/tests/job_runner/test_job_dispatch.py
pinkwah: https://github.com/equinor/libres/blob/05e06df2926c77f7d9031e3268178afe360f0127/setup.py#L63-L67

pinkwah: Basically instead of doing it the incorrect hacky way I decided to do it the Pythonic correct way and let `setuptools`/`pip` handle creation of shebangs.

sondreso: Ahh, I didn't see that! Agree this is much nicer!

README.md
```diff
@@ -1,8 +1,22 @@
...
+
+``` sh
+$ pip install git+https://github.com/equinor/libres.git@master
+```
+
+## Building (Legacy)
+
+*libres* is meant to be installed using `setup.py`, rather than `CMakeLists.txt`.
+
+*libres* is a C library with Python bindings. The C library requires [*libecl*](https://github.com/equinor/libecl)
```
markusdregi: I'm not sure if it matters, but our "official statement" the last year or so is that we're a Python library with (large) parts implemented in C.

setup.py
```diff
@@ -0,0 +1,94 @@
...
+        "-DRES_VERSION=" + version,
+        "-DECL_INCLUDE_DIRS=" + get_libecl_include(),
+        # we can safely pass OSX_DEPLOYMENT_TARGET as it's ignored on
+        # everything not OS X. We depend on C++11, which makes our minimum
+        # supported OS X release 10.9
+        "-DCMAKE_OSX_DEPLOYMENT_TARGET=10.9",
+        "-DCMAKE_INSTALL_LIBDIR=python/res/.libs",
+    ],
+    classifiers=[
+        "Development Status :: 1 - Planning",
```
markusdregi: I would assume we can bump this one ever so slightly :shrug:

pinkwah:
https://github.com/equinor/libres/blob/05e06df2926c77f7d9031e3268178afe360f0127/python/res/__init__.py#L68

check if exist? https://github.com/equinor/libres/blob/05e06df2926c77f7d9031e3268178afe360f0127/python/res/__init__.py#L78

publish libecl applications to pypi

pinkwah:
Jenkins test this please

README.md
```diff
@@ -1,110 +1,89 @@
...
-bash% export PYTHONPATH=/local/ert/install/lib/python3.6/site-packages:$PYTHONPATH
+$ pip install -r requirements.txt
+$ python setup.py develop
 ```
 
-Observe that path components `lib64` and `lib/python3.6/site-packages` will
-depend on your Python version and which Linux distribution you are using. The
-example given here is for RedHat based distributions.
+Alternatively, `pip install -e .` will also setup `libres` for development, but
+it will be more difficult to recompile the C library.
```
sondreso, line 30: How do you recompile in the two different scenarios?

pinkwah: I'll add this:

[scikit-build](https://scikit-build.readthedocs.io/en/latest/index.html) is used for compiling the C library. It creates a directory named `_skbuild` which is reused upon future invocations of either `python setup.py develop`, or `python setup.py build_ext`. The latter only rebuilds the C library. In some cases this directory must be removed in order for compilation to succeed.

sondreso (approved):
Amazing work! :rocket:

PR #936, opened 2020-04-29
Stop shell jobs from using data_root

### LINKED ISSUE #934: Abandon path interpretation for shell jobs
Currently shell jobs interpret relative paths relative to the config file location over the location of the runpath. This functionality is to be abandoned and instead users will have to apply `<CONFIG_PATH>` explicitly if they want to operate on files relative to the configuration location.

markusdregi:
This is a major break in all shell jobs in libres. Prior to this commit
all shell jobs would interpret relative paths first relative to the
configuration file and only if that file did not exist, relative to the
runpath. From now on users will have to use <CONFIG_PATH> to pick up
files relative to the configuration file.

**Issue**
Resolves #934

### DIFF
--- a/python/res/fm/shell/shell.py
+++ b/python/res/fm/shell/shell.py
@@ -10,26 +10,6 @@ class Shell(object):
     """
 
     @staticmethod
-    def prefix_path(input_path):
-        # Absolute input path goes straight through.
-        if os.path.isabs( input_path ):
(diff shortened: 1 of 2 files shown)


sondreso:
Looks good, but the tests needs an update :100:

sondreso (approved):
:canoe:

Issue #946, opened 2020-05-19
Do not write to the same tmp file for each test


lars-petter-hauge:
Issue; if two different users try to the run the tests the second will fail with permissions

https://github.com/equinor/libres/blob/525b74295e1771a888738c5b72a1e634ce9229d3/lib/job_queue/tests/job_torque_test.cpp#L52

Should generate a temporary file  with a unique name instead.

PR #947, opened 2020-05-19
Use mkstemp in torque_test

lars-petter-hauge:
**Issue**
#946 

**Approach**
Use a tmp generated name instead.

Quick note from cppref:

> Although the names generated by std::tmpnam are difficult to guess, it is possible that a file with that name is created by another process between the moment std::tmpnam returns and the moment this program attempts to use the returned name to create a file. 

Could be generated some warnings on this, but in terms of the test itself we should be good.

### DIFF
--- a/lib/job_queue/tests/job_torque_test.cpp
+++ b/lib/job_queue/tests/job_torque_test.cpp
@@ -49,7 +49,21 @@ void setoption_setalloptions_optionsset() {
   test_assert_true( torque_driver_set_option( driver , TORQUE_SUBMIT_SLEEP , "0.25"));
   test_assert_int_equal( 250000 , torque_driver_get_submit_sleep(driver));
 
-  test_assert_true( torque_driver_set_option( driver , TORQUE_DEBUG_OUTPUT , "/tmp/torqueue_debug.txt"));
+  char tmp_path[] = "/tmp/torque_debug_XXXXXX";
+  // We do not strictly need the file, we are only interested in a path name
+  // tmpnam is however deprecated in favor of mkstemp, and no good substitute
+  // for tmpnam (with similar functionality) was found.
+  int fd = mkstemp(tmp_path);
+
+  if (fd == -1) {
+    printf("Unable to create dummy log file");
+    exit(1);
+  }
+
+  close(fd);
+  unlink(tmp_path);
+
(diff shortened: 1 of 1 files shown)


lars-petter-hauge:
or not..

```/Users/travis/build/equinor/libres/lib/job_queue/tests/job_torque_test.cpp:52:83: error: 'tmpnam' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of tmpnam(3), it is highly recommended that you use mkstemp(3) instead. [-Werror,-Wdeprecated-declarations]```

swapping out with mkstemp

File: lib/job_queue/tests/job_torque_test.cpp [hunk 1]
```diff
@@ -49,7 +49,10 @@ void setoption_setalloptions_optionsset() {
   test_assert_true( torque_driver_set_option( driver , TORQUE_SUBMIT_SLEEP , "0.25"));
   test_assert_int_equal( 250000 , torque_driver_get_submit_sleep(driver));
 
-  test_assert_true( torque_driver_set_option( driver , TORQUE_DEBUG_OUTPUT , "/tmp/torqueue_debug.txt"));
+  char tmp_path[] = "/tmp/torque_debug_XXXXXX.txt";
+  mkstemp(tmp_path);
```
sondreso: From documentation: 

> RETURN VALUE
> Upon successful completion, mkstemp() shall return an open file descriptor. Otherwise, -1 shall be returned if no suitable file could be created.

Do we want to have an open file descriptor here? We only need the path, and i suspect this can have unintended side effects.

lars-petter-hauge: I agree, which is why I started out with `std::tmpnam` instead.

Would adding a close to it at the end of the test, as it is in its current state, suffice? Or would you rather use a different implementation?

sondreso: It's interesting, because the reason why it has to create and open the file is because it wants to guarantee that you don't create collisions with others trying to create temporary files. I don't know how the torque driver works, but if that is just appending to the file it gets, then you could just close the file immediately after you get it back from `mkstemp`. That way the `mkstemp` would serve as a `touch` and then you can use it as normal in the torque driver.

sondreso: We should also handle the error case when `mkstemp` returns `-1`.

sondreso (approved):

PR #948, opened 2020-05-19
Backup PYTHONPATH when starting RMS

sondreso:
**Issue**
Resolves partially https://github.com/equinor/fmu-dev/issues/17

**Approach**
Backup PYTHONPATH to environment variable

When this is merged I will create a similar patch the runrms in subscript.

### DIFF
--- a/python/res/fm/rms/rms_run.py
+++ b/python/res/fm/rms/rms_run.py
@@ -150,10 +150,12 @@ class RMSRun(object):
         if self.config.threads:
             args += ["-threads", str(self.config.threads)]
 
-        if exec_env:
-            os.execve(self.config.executable, args, exec_env)
-        else:
-            os.execv(self.config.executable, args)
+        if not exec_env:
+            exec_env = os.environ.copy()
+        exec_env["_PRE_RMS_BACKUP"] = "1"
+        if "PYTHONPATH" in os.environ:
+            exec_env["_PRE_RMS_PYTHONPATH"] = os.environ["PYTHONPATH"]
+        os.execve(self.config.executable, args, exec_env)
 
 
     @staticmethod
--- a/python/tests/res/fm/test_rms_run.py
+++ b/python/tests/res/fm/test_rms_run.py
@@ -31,6 +31,47 @@ from tests.utils import tmpdir
 import pytest
 from tests.conftest import source_root
 
+
+@tmpdir()
+@pytest.mark.parametrize(
+    "python_path",
+    [
+        (""),
+        ("/test/python/path/"),
+        (None),
+    ],
+)
+def test_run_backup_python_path(tmpdir, monkeypatch, python_path):
+    with open("rms_config.yml", "w") as f:
+        f.write("executable:  {}/bin/rms".format(os.getcwd()))
+
+    os.mkdir("run_path")
+    os.mkdir("bin")
+    os.mkdir("project")
+    shutil.copy(os.path.join(source_root(), "python/tests/res/fm/rms"), "bin")
+    monkeypatch.setenv("RMS_SITE_CONFIG", "rms_config.yml")
+    if python_path is None:
+        monkeypatch.delenv("PYTHONPATH")
+    else:
+        monkeypatch.setenv("PYTHONPATH", python_path)
+
+    action = {"exit_status": 0}
+    with open("run_path/action.json", "w") as f:
+        f.write(json.dumps(action))
+
+    r = RMSRun(0, "project", "workflow", run_path="run_path")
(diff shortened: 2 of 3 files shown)


python/res/fm/rms/rms_run.py
```diff
@@ -150,10 +151,10 @@ def exec_rms(self, exec_env):
         if self.config.threads:
             args += ["-threads", str(self.config.threads)]
 
-        if exec_env:
-            os.execve(self.config.executable, args, exec_env)
-        else:
-            os.execv(self.config.executable, args)
+        if not exec_env:
+            exec_env = os.environ.copy()
+        exec_env[self.BACKUP_ENVIRONMENT_VARIABLE] = os.environ.get("PYTHONPATH","")
```
hnformentin: Why do we need "" in os.environ.get("PYTHONPATH","")?
I tried os.environ.get("PYTHONPATH","") and os.environ.get("PYTHONPATH") and it seemed the same here.

python/tests/res/fm/test_rms_run.py
```diff
@@ -31,6 +31,34 @@
...
+    os.mkdir("project")
+    shutil.copy(os.path.join(source_root(), "python/tests/res/fm/rms"), "bin")
+    monkeypatch.setenv("RMS_SITE_CONFIG", "rms_config.yml")
+    monkeypatch.setenv("PYTHONPATH", "/test/python/path/")
+
+    action = {"exit_status": 0}
+    with open("run_path/action.json", "w") as f:
+        f.write(json.dumps(action))
+
+
```
hnformentin: Do we need two spaces here?

python/res/fm/rms/rms_run.py
sondreso: `dict.get("SOMETHING")` returns `None` if SOMETHING is not found in the dictionary, while `dict.get("SOMETHING", "")` returns the empty string if it does not find SOMETHING. :slightly_smiling_face:

pinkwah: `PYTHONPATH` might be unset, in which case `os.environ.get("PYTHONPATH")` returns `None` and not `""`. I do question the line here, if `PYTHONPATH` is unset, shouldn't `_PRE_RMS_PYTHONPATH` also be unset?

python/res/fm/rms/rms_run.py
```diff
@@ -25,6 +25,7 @@ class RMSRun(object):
     _multi_seed_file = "random.seeds"
     _max_seed = 2146483648
     _seed_factor = 7907
+    BACKUP_ENVIRONMENT_VARIABLE = "_PRE_RMS_PYTHONPATH"
```
pinkwah: You are setting this constant that you're only using once. This can just be an inline string literal whenever you need to use it. We're won't be changing the name of the backup envvar, so why have an unnecessary level of indirection.

python/res/fm/rms/rms_run.py
sondreso: Yes, I also thought about this, but what should the `run_external` script do when it cannot find the `_PRE_RMS_PYTHONPATH` variable? Continue with the current `PYTHONPATH` or unset it?

markusdregi: Unset it?

markusdregi: I guess you're trying to avoid the scenario where one cannot separate between RMS being run from `rms_run` without  `PYTHONPATH` and just being run not using `rms_run`. That makes a lot of sense, it can potentially be hard to debug, especially since there are already at least two standard ways of running RMS as of today.

However, this information could be backed up by introducing a second environment variable; a `_PRE_RMS_BACKUP` variable that is only set if we backup. In this manner, we can separate the two 🤷

hnformentin (dismissed):
LGTM and the changes requested where answered/implemented (from my understanding only missing to address @dotfloat comment about BACKUP_ENVIRONMENT_VARIABLE).

python/res/fm/rms/rms_run.py
sondreso: I'm not very opinionated in this situation, so I will change it. But in general I think it's nice to collect "magic"/constant strings in one place!

sondreso:
> However, this information could be backed up by introducing a second environment variable; a _PRE_RMS_BACKUP variable that is only set if we backup. In this manner, we can separate the two :shrug: 

Implemented this, and extended the test cases. Also implemented the same changes here: https://github.com/equinor/equilibrium/pull/25.

markusdregi (approved):
I think this looks good 👍

PR #944, opened 2020-05-15
Part 2: Slurm driver get status of jobs

joakim-hove:
**Issue**
Resolves #941 

This is on top of #949 which should be merged first.

**Approach**
This is the last of the initial slurm driver PR's, with this PR the driver is in Works for me&trade; condition. The main new code in this PR is functionality to query the status of jobs in the queue system, this is moderately complex because:

1. Two different commands - `sqeueue` and `scontrol` must be used depending on whether the job is currently in the queue or not.
2. The code will be used in multithreaded context.
3. There is some caching involved to avoid hammering the slurm system with status queries.

Issue #938, opened 2020-05-13
Support for SLURM queue system


joakim-hove:
In cooperation with TNO I am about to start implementing support for the slurm queue system in libres. My initial plan is to implement the slurm support along the lines of current lsf support - in particular the interaction with the slurm system will be by issuing shell commands and capturing and parsing stdout.

I create this issue here to give a heads up to the will be reviewers, and also invite invite early opinions on the implementation strategy. One question right away: Do you prefer one large PR or multiple small PR's? Personally I would prefer to submit multiple small PR's, but in that case the slurm implementation will not be usable before the last PR is merged.

markusdregi:
Good evening Dr. Hove and welcome back to the ERT show 🎉 

As for the question I too prefer smaller PRs. The fact that the slurm implementation is not functional until the end is fine by me 🤷 Another option would be to create a feature branch for Slurm support, review smaller PR's into that one and in the end merge it into master. Do you have any preferences on this @sondreso?

As for keeping track of progress and getting an overview while reviewing; it would be nice if you could create a milestone or epic for slurm support and then you could try to write out a set of issues to highlight the initial plan for an approach and resolving them as you go along. Then we can add more issues as they come up.

Besides that, I can't think of anything else. Oh! Since Slurm is open source in contrast to for instance LSF; do you think it would be possible to write some integration tests that could run on Travis? Would at least be nice to investigate the feasibility of this...

joakim-hove:
> Good evening Dr. Hove and welcome back to the ERT show 

Thank you!

> [...]  I too prefer smaller PRs.The fact that the slurm implementation is not functional until the end is fine by me

That is good - then I will head in that direction.

> it would be nice if you could create a milestone or epic for slurm support and then you could try to write out a set of issues to highlight the initial plan for an approach and resolving them as you go along. 

That is a good idea - I will do that.

> Since Slurm is open source in contrast to for instance LSF; do you think it would be possible to write some integration tests that could run on Travis? Would at least be nice to investigate the feasibility of this...

Whoha; might be possible - I will look into it. The thought of how to test has been goind as a background process for some time - maybe actually installing slurm and firing off proper tests is actually feasible - good idea!

joakim-hove:
>  it would be nice if you could create a milestone or epic for slurm support

Unfortunately it seems I do not have the permission to create milestones :disappointed:

markusdregi:
Created one ;)

sondreso:
[...]

PR #949, opened 2020-05-21
Part 1: Slurm driver submit and cancel

joakim-hove:
**Issue**
Part of #938

This is the next PR in the series of slurm drivers and should be merged before #944 

**Approach**
In this PR slurm_driver function to submit a job and to subsequently cancel it again are added. The functions are based on calling the external programs `sbatch` and `scancel`.

As part of the PR a new block in the `CMakeLists.txt` is created which is condition on finding the `sbatch` executable - i.e. whether the machine we are running on is indeed part of a slurm cluster. If the machine is not part of a slurm cluster the tests are built anyway - but not run.

### DIFF
--- a/lib/job_queue/slurm_driver.cpp
+++ b/lib/job_queue/slurm_driver.cpp
@@ -52,6 +52,7 @@ struct slurm_driver_struct {
   std::string scancel_cmd;
   std::string squeue_cmd;
   std::string scontrol_cmd;
+  std::string partition;
 };
 
 
@@ -92,6 +93,9 @@ const void * slurm_driver_get_option( const void * __driver, const char * option
   if (strcmp(option_key, SLURM_SQUEUE_OPTION) == 0)
     return driver->squeue_cmd.c_str();
 
+  if (strcmp(option_key, SLURM_PARTITION_OPTION) == 0)
+    return driver->partition.c_str();
+
   return nullptr;
 }
 
@@ -118,11 +122,17 @@ bool slurm_driver_set_option( void * __driver, const char * option_key, const vo
     return true;
   }
 
+  if (strcmp(option_key, SLURM_PARTITION_OPTION) == 0) {
+    driver->partition = static_cast<const char*>(value);
+    return true;
+  }
+
   return false;
 }
 
 
 void slurm_driver_init_option_list(stringlist_type * option_list) {
+  stringlist_append_copy(option_list, SLURM_PARTITION_OPTION);
   stringlist_append_copy(option_list, SLURM_SBATCH_OPTION);
   stringlist_append_copy(option_list, SLURM_SCONTROL_OPTION);
   stringlist_append_copy(option_list, SLURM_SQUEUE_OPTION);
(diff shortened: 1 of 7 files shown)


joakim-hove:
I consider the slurm work consisting of this PR and the two subsequent PR's #944 and #951 to be "complete" - a review would be cool.

lib/job_queue/slurm_driver.cpp
```diff
@@ -118,22 +166,69 @@ bool slurm_driver_set_option( void * __driver, const char * option_key, const vo
...
+  std::vector<std::string> sbatch_argv = {"--workdir=" + std::string(run_path), "--job-name=" + std::string(job_name), "--parsable"};
+  if (!driver->partition.empty())
+    sbatch_argv.push_back( "--partition=" + driver->partition );
+  sbatch_argv.push_back( submit_script );
+
+
+  auto file_content = load_stdout(driver->sbatch_cmd.c_str(), sbatch_argv);
+  util_unlink_existing( submit_script.c_str() );
+
+  int job_id = 0;
```
markusdregi: Is this safe? Is it guaranteed that the job id will never be 0?

joakim-hove: It is at least guaranteed it will not be negative; I'll update

markusdregi (commented):
Sorry for the delayed review. We will try to follow them up a bit more rapidly now - hoping to get them in :)

Besides the question regarding the `job_id`, the implementation looks good. However, for the test suite, could you implement a version of the tests that mock the slurm commands that is always executed? In that manner, one can always test the libres relevant slurm code, but the integration will only be tested if one is indeed on a slurm cluster. Furthermore, it would be nice if the tests could be expanded to verify that the submitted job finished successfully (perhaps via output to a predefined disk area that is assumed to be shared between the machine(s)). And that the kill command actually kills the job (will most likely need some sleeping to get this to work).

Besides extending the tests, this look good 👍

joakim-hove:
>[...] could you implement a version of the tests that mock the slurm commands that is always executed? 

I will make some slurm mock commands which can test the basic libres <-> slurm interaction; I will avoid creating stateful slurm mocking, as I think that will be a quite large and error prone task; and will be more about recreating slurm behaviour than the libres <-> slurm interaction.

> Furthermore, it would be nice if the tests could be expanded to verify that the submitted job finished successfully

Here I disagree - that jobs have succeeded (when the queue layer says so) is already handled at the `job_queue` layer which is above the slurm/lsf/local/... drivers.

> And that the kill command actually kills the job (will most likely need some sleeping to get this to work).

The integration tests in part 2 verifies this. I can not see how this can be mocked without creating a stateful slurm mock server?

joakim-hove:
OK - I have fixed the `job_id` detail, then I have added a quite large mock based test in PR2. To make a meaningful test functionality which is only present there was really needed. I hope this can be merged in current form - or alternatively review focus could (should) shift to PR2 - which is superset of this PR.

markusdregi (approved):
Thanks for the tests added in #944

PR #950, opened 2020-05-22
Bump libecl version t 2.9.0

ManInFez:


sondreso (approved):
Thanks for the patch!

PR #951, opened 2020-05-25
Part3: Slurm driver mpi

joakim-hove:
**Issue**
This is the final(?) part of #938. This is on top of #944 

**Approach**
In this PR basic MPI awareness is added to the slurm driver. Slurm allows very detailed control over how the tasks should be distributed over available nodes and cpus, this PR just does the absolute simplest - which is to state how many cpus are needed in total.

The main part of the PR is actually parsing the strings set in the environment variables `SLURM_JOB_NODELIST` and `SLURM_TASKS_PER_NODE` - this is an incredibly compact notation to get the names of the all the nodes in the allocation, and how many cpus are allocated on each of them, this is needed for mpirun.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -465,7 +465,8 @@ foreach(name job_status_test
 endforeach()
 
 find_program(SBATCH "sbatch")
-foreach(name job_slurm_submit)
+foreach(name job_slurm_submit
+             job_slurm_runtest)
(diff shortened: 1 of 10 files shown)


markusdregi (commented):
Looks good 👍 Rebase and we'll have a final look before merging it...

joakim-hove:
Rebased

markusdregi (approved):
Looks good 👍

PR #953, opened 2020-05-29
Don't check for Python packages in CMake

pinkwah:
During configuration, CMake checks for Python package versions using our own hand-rolled CMake code. This is currently causing problems on Komodo, where CMake rejects the build due to a package that verifiably exists (pandas). The code also bypasses pip entirely and shouldn't really be used.

This change deletes the "convenience" of having CMake tell you if the correct Python packages are installed, and can be safely removed.


oysteoh (approved):

pinkwah:
Jenkins test this please

PR #955, opened 2020-06-02
Pin configsuite to 0.5.3 or lower

ManInFez:


markusdregi (approved):
I would have preferred `< 0.6` in case a bug fix for `0.5` was released. But I'll leave it to you ;)

ManInFez:
> I would have preferred `< 0.6` in case a bug fix for `0.5` was released. But I'll leave it to you ;)

I agree 👍

PR #956, opened 2020-06-04
Slurm driver part 4: extra options

joakim-hove:
**Issue**
This is part of #938

**Approach**
This PR just adds support for multiple options which can be set with the
```
QUEUE_OPTION SLURM  <option> <value>
```
syntax


joakim-hove:
rebased

joakim-hove:
rebased - again!

joakim-hove:
I think this should be ready?!

joakim-hove:
For the interaction with everest: https://github.com/equinor/everest/pull/1328 it would be very convenient if this could be reviewed and merged now.

sondreso (approved):
LGTM!

PR #957, opened 2020-06-04
Always load node to ensure gen_param is loaded correct

### LINKED ISSUE #904: General parameter fails when updating
`GEN_PARAM` was reported by a user to fail when running an update step. I've reproduced the problem locally and committed a rather minimalistic case based on the poly case in #903.

sondreso:
**Issue**
Resolves https://github.com/equinor/ert/issues/904

**Approach**
Metadata about report step is not propagated for GEN_PARAM when the entire vector is active (which it almost always is). This will result in a small performance hit, but it ensures that GEN_PARAM is loaded correct when deserializing after an update.  (See also: https://github.com/equinor/libres/pull/310)


sondreso:
Jenkins, test this please

markusdregi (approved):
Well investigated 👏

PR #963, opened 2020-06-24
Identify the slurm status CONFIGURING as running

joakim-hove:
**Issue**
For jobs which are "just about to start" the slurm system can return status `CONFIGURING` - this pr will translate that to running.


sondreso (approved):
LGTM!

Issue #964, opened 2020-06-25
Queue status study


markusdregi:
It seems like there are some queue statuses that is not being handled in the new runner code in Python. This should be investigated.

PR #966, opened 2020-06-26
Changes to job_node code marking node as failed

joakim-hove:
**Issue**
This is tricky. The job_queue_node code has functionality to confirm that a simulation is *actually running*. The purpose of this code is to cater for the following situation which has happened in the past:

1. The job is submitted to the queue system and the queue system reports the job with status `JOB_QUEUE_RUNNING`.
2. File system mounts are not 100% sound - and the output files from the simulation never become visible from the main node running ert. If the status file is still not visible 120 seconds after the queue system reported the job as running the status is set to `JOB_QUEUE_DO_KILL_NODE_FAILURE`.

Now the status `JOB_QUEUE_DO_KILL_NODE_FAILURE` has appeared in the slurm based simulations at TNO and this is problematic for several reasons:

1. When the status monitor code was refactored at some point the `JOB_QUEUE_DO_KILL_NODE_FAILURE` status was forgotten - and the status is no longer handled. This omission is common to all queue drivers and not related to slurm - see #964 

2. The entire `JOB_QUEUE_DO_KILL_NODE_FAILURE` functionality was a *desperate* attempt to address subpar LSF infrastructure, and it is not really general enough to configure it individually for different queue systems.

**Approach**
The "solution" chosen here is quite simple, and not really a complete solution. The code is changed so that if `job_node->max_wait_time == 0` the status of the running jobs is never set to `JOB_QUEUE_DO_KILL_NODE_FAILURE` - a warning is issued and job is retained in the status `JOB_QUEUE_RUNNING`- with the hope that nfs mounts will eventually behave. 

A proper solution to this will at least require:

1. The management layer must adhere to the `JOB_QUEUE_DO_KILL_NODE_FAILURE` - i.e. #964.
2. The `job_node->max_wait_time` must be refactored to be configurable on a per queue system basis. This is probably not very difficult, but there will probably be quite many changes in many places - in quite complex code. I do not really feel comfortable embarking on that in the current situation.

This is a quite significant endeavour.

**Immediate way forward**
For now I suggest that TNO tries this PR - preferably on a ERT run with *many* realizations. If we are lucky the offending nodes are not really broken, and the file system mounts come back online before the simulation has completed. If you can identify problematic nodes you can disable them from the submission with:
```
QUEUE_OPTION SLURM EXCLUDE_HOST host1, host2
QUEUE_OPTION SLURM EXCLUDE_HOST host3, host4
```
@markusdregi

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -479,7 +479,7 @@ endforeach()
 foreach(name job_loadOK
              ext_joblist_test
              job_torque_submit_test
-             job_queue_timeout_test
+             #job_queue_timeout_test
              job_queue_stress_test
              job_queue_stress_task
              job_job_queue_test
@@ -568,8 +568,8 @@ if (SBATCH)
     True)
 endif()
 
-add_test(NAME job_queue_timeout_test
-         COMMAND job_queue_timeout_test $<TARGET_FILE:job_queue_stress_task>)
+#add_test(NAME job_queue_timeout_test
+#         COMMAND job_queue_timeout_test $<TARGET_FILE:job_queue_stress_task>)
 
 file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/job_queue/tests/data/qsub_emulators/
      DESTINATION ${EXECUTABLE_OUTPUT_PATH})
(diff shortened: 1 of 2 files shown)


markusdregi:
Thanks for the PR @joakim-hove. As you say, this is more of a patch then a proper fix of the problem and hence I'm unsure whether this should at all be merged. Have you received any notice from TNO regarding whether this improved their issues?

ertomatic:

ertomatic:

sondreso:
This PR is closed due to inactivity. Feel free to rebase and adapt to the changes that have been introduced (mainly pip installable libres and black formatting), and reopen the PR :smile:

PR #967, opened 2020-06-29
Make COPY_DIRECTORY always copy to target folder

### LINKED ISSUE #608: Forward model COPY_DIRECTORY - confusing beahviour
Strange behavior for forward model COPY_DIRECTORY: When destination folder exist it creates a new directory within the existing. Eg COPY_DIRECTORY (<FROM>=foo, <TO>=bar) crates bar/bar/foo when bar already exists. If bar does not exist it becomes a copy of foo, as expected.

sondreso:
**Issue**
Resolves #608 

With this change, COPY_DIRECTORY will not create a folder with the
same name as the source directory inside the target folder if the target
folder already exists. This deviates from the standard shell commands.

### DIFF
--- a/python/res/fm/shell/shell.py
+++ b/python/res/fm/shell/shell.py
@@ -158,8 +158,6 @@ class Shell(object):
                     Shell.mkdir( target_root )
 
             print("Copying directory structure %s -> %s" % (src_path , target_path))
-            if os.path.isdir( target_path ):
-                target_path = os.path.join( target_path, src_basename )
             distutils.dir_util.copy_tree( src_path , target_path , preserve_times = 0)
         else:
             raise IOError("Input argument:'%s' does not correspond to an existing directory" % src_path)
--- a/python/tests/res/fm/test_shell.py
+++ b/python/tests/res/fm/test_shell.py
@@ -239,11 +239,36 @@ class ShellTest(ResTest):
             self.assertFalse( os.path.exists( "path" ))
             self.assertTrue( os.path.exists("link_target/link_file"))
 
+    def create_source_directory(self):
+        os.mkdir("source")
+        with open("source/file.txt" , "w") as f:
+            f.write("dummy content\n")
+
+
+    def test_copy_directory(self):
+        with TestAreaContext("copy/directory"):
+            self.create_source_directory()
+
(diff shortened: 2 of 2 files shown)


markusdregi:
There is a new scenario coming up with this approach that needs to be considered, documented and tested. What about files residing in `target` that does not exist in `source`? If we remove target before we copy, that should be documented. If we don't remove, the reproducibility of the forward model in the case of a resubmit is lost in cases where `CAREFUL_COPY` or similar logic is used. I think this needs input from @eivindsm

eivindsm:
I lean towards leaving target intact before copying, and possibly adding an option to delete before copying (--del-existing-target or something) I asked @berland to comment on behalf of FMU.

berland:
I support leaving the target intact. Resubmits over existing runs on disk are already very slippery, but will improve with such a changed COPY_DIRECTORY. Such resubmits combined with CAREFUL_COPY already means you are asking for trouble.

markusdregi:
Notice that those resubmits might happen without the users asking for it as the node for instance might go down and then the forward model will by default be resubmitted once. I know the probability is somewhat low, but apply that on a large ensemble, over multiple assets over time and it will occur. Will most likely lead to undetected, yet broken realisations that are taken into account 🤷

berland:
I have had an aimed peek into some ERT model files today, and I found this:
```FORWARD_MODEL COPY_DIRECTORY(<FROM>=<CONFIG_PATH>/../../sim2seis/model,                 <TO>=<RUNPATH>/sim2seis/)```

This is an ert config that is tailored to run only as a (manual) "resubmit" on the same realization directories on scratch. It assumes that the previous ert config did a `MAKE_DIRECTORY(..sim2seis)`, and the intention is that directory `model` should end up as `<RUNPATH>/sim2seis/model`

Will the current PR change this behaviour?

markusdregi:
@berland I believe it will...

berland:
Metoo. Then I vote for **not** solving the original issue, as we are already "working around" it in a lot of ERT configs.

markusdregi:
I just had a discussion with @eivindsm and we agree with @berland. Suggested approach is to close the PR and instead document the behaviour in the forward model documentation...

PR #971, opened 2020-07-13
Use eclrun to start eclipse

### LINKED ISSUE #952: Use ecl_run to run eclipse
The deploy of Eclipse has changed and to be compliant with the deploy we should start using the `ecl_run` binary to execute Eclipse in the forward model job. It will setup the environment correctly for us, but we will have to figure out how the various arguments that we accept today can be passed to this binary instead.

It is the inclusion of version `2019.1` that is driving this change. We can either run the old versions using the old logic, or run all versions using the `ecl_run` script. My preference would be the later to reduce complexity if possible.

sondreso:
**Issue**
Resolves #952

### DIFF
--- a/python/res/fm/ecl/ecl_config.py
+++ b/python/res/fm/ecl/ecl_config.py
@@ -1,5 +1,6 @@
 import os
 import yaml
+import subprocess
 import sys
 import re
 
@@ -71,7 +72,7 @@ class EclConfig(object):
 
     """
 
-    def __init__(self, config_file):
+    def __init__(self, config_file, simulator_name="not_set"):
         with open(config_file) as f:
             try:
                 config = yaml.safe_load(f)
@@ -80,13 +81,19 @@ class EclConfig(object):
 
         self._config = config
         self._config_file = os.path.abspath(config_file)
-
+        self.simulator_name = simulator_name
 
     def __contains__(self, version):
         if version in self._config[Keys.versions]:
             return True
 
         return self.default_version is not None and version in [None, Keys.default]
+
+
+    def get_eclrun_env(self):
+        if "eclrun_env" in self._config:
+            return self._config["eclrun_env"].copy()
+        return None
 
 
     @property
@@ -167,7 +174,7 @@ class Ecl100Config(EclConfig):
 
     def __init__(self):
         config_file = os.getenv("ECL100_SITE_CONFIG", default = self.DEFAULT_CONFIG_FILE)
-        super(Ecl100Config, self).__init__(config_file)
+        super(Ecl100Config, self).__init__(config_file, simulator_name="eclipse")
 
 
 class Ecl300Config(EclConfig):
@@ -176,7 +183,7 @@ class Ecl300Config(EclConfig):
 
     def __init__(self):
         config_file = os.getenv("ECL300_SITE_CONFIG", default = self.DEFAULT_CONFIG_FILE)
-        super(Ecl300Config, self).__init__(config_file)
+        super(Ecl300Config, self).__init__(config_file, simulator_name="e300")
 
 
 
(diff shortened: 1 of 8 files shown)


python/tests/res/fm/test_ecl_run_new.py
```diff
@@ -0,0 +1,197 @@
...
+    def test_failed_run(self):
+        self.init_eclrun_config()
+        shutil.copy(
+            os.path.join(self.SOURCE_ROOT, "test-data/local/eclipse/SPE1_ERROR.DATA"),
+            "SPE1_ERROR.DATA",
+        )
+        ecl_config = Ecl100Config()
+        eclrun_config = EclrunConfig(ecl_config, "2019.3")
+        ecl_run = EclRun("SPE1_ERROR", None)
+        with self.assertRaises(Exception):
```
lars-petter-hauge: No need to do a double run, keep the context instead:

```
with self.assertRaises(Exception) as error_context:
    ecl_run.runEclipse(eclrun_config=eclrun_config)
self.assertIn("ERROR", str(error_context.exception))
```

There are multiple checks further down in the tests here that might benefit from checking the error string, as they do assert on fairly generic errors. But I haven't looked at them in detail, so I am not sure if they are our own or not.

python/tests/res/fm/test_ecl_run_new.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @pytest.mark.equinor_test
+    @tmpdir()
+    def test_run_api(self):
+        self.init_eclrun_config()
+        shutil.copy(
+            os.path.join(self.SOURCE_ROOT, "test-data/local/eclipse/SPE1.DATA"),
+            "SPE1.DATA",
+        )
+        ecl_config = Ecl100Config()
```
lars-petter-hauge, line 97: ~Should this not test `EclrunConfig`? How does this differ from the tests in `test_ecl_run.py`? This goes for multiple of these tests.~
Or is this set through `eclrun_env` set in `init_eclrun_config`

python/tests/res/fm/test_ecl_run_new.py
```diff
@@ -0,0 +1,197 @@
...
+import inspect
+import yaml
+import pytest
+import re
+
+from ecl.summary import EclSum
+from ecl.util.test import TestAreaContext
+from tests import ResTest
+from tests.utils import tmpdir
+from res.fm.ecl import *
```
lars-petter-hauge: It would be nice to avoid these wildcard imports. The sudden `run` method further down comes from `res.fm.ecl` believe, but let's make that explicit.

python/res/fm/ecl/ecl_config.py
```diff
@@ -186,4 +193,49 @@ class FlowConfig(EclConfig):
 
     def __init__(self):
         config_file = os.getenv("FLOW_SITE_CONFIG", default = self.DEFAULT_CONFIG_FILE)
-        super(FlowConfig, self).__init__(config_file)
+        super(FlowConfig, self).__init__(config_file, simulator_name="flow")
+
+
+
+class EclrunConfig:
```
lars-petter-hauge, line 200: The naming here is a bit awkward, as we have the classes `EclConfig` with a subclass (`EclXXXConfig`) for each of the simulators, while `EclrunConfig` is by all means not a subclass of `EclConfig`, and instead is ment to be used in addition to `EclConfig`. I do understand the issue of having a configclass for a binary named `eclrun`, and then what do you name it.. But at least this was a bit confusing

python/res/fm/ecl/ecl_run.py
```diff
@@ -255,7 +252,34 @@ def initMPI(self):
...
+            "eclrun",
+            "-v",
+            eclrun_config.version,
+            "--np",
+            str(self.num_cpu),
+            eclrun_config.simulator_name,
+            "{}.DATA".format(self.base_name)
+        ]
+
+    def _get_legacy_run_command(self):
```
lars-petter-hauge, line 266: The challenge here (correct me if I'm wrong) is that all new versions have gotten a changed entry point with different way of running the executable. However this only applies for one of the simulators, and e.g. flow will be started through the "old" method still. So it isn't really a `legacy` run command.

lars-petter-hauge:
I had a couple comments, but other than that I believe I agree in the method. It feels like it brings unnecessary complexity, and at the same time is done as straight forward as possible, given the issue at hand. I don't have a better suggestion either.

The `eclrun` binary takes as argument the name of the simulator (`e100`, `e300` etc). Would it then not be possible to deprecate more of the configurations here? e.g. it might not be necessary with the `Ecl100Config` aynmore? (`FlowConfig` must be kept though). That would reduce the complexity a bit.

Edit: not enough supported versions in `eclrun`

python/tests/res/fm/test_ecl_run_new.py
sondreso: Yes, `init_eclrun_config` will create a `Ecl100Config` that has the necessary configs for the `eclrun` command and `EclrunConfig` to work (And no config for the old style way of starting eclipse, which will make the tests fail if `eclrun` is not working since there is no fallback).

python/res/fm/ecl/ecl_run.py
sondreso: You are correct, both flow and eclipse versions older than `2018.2` will be run using the old method. So yes for flow it's not a "legacy" run command, but for eclipse it is :man_shrugging: Overall I don't like the fact that we are combining everything in how we run flow and eclipse, when they now differ in the way they are executed.

python/res/fm/ecl/ecl_config.py
sondreso: Completely agree, but I was not able to come up with any better naming! Very open to suggestions

lars-petter-hauge: I hate how annoying naming can be, and in this case specifically, where we end up in a slight pinch due to naming of the binary which is completely out of our hands.

Please provide some documentation to the `EclrunConfig`, as to what it is (and highlight the difference in relation to the other `Ecl*Config` classes, or not necesserily refer to a different class in the doc, but you get it..), and then I think we are good to go.

python/res/fm/ecl/ecl_run.py
lars-petter-hauge: Agree, flow should probably have gotten it's own way of execution.

lars-petter-hauge (approved):
Lgtm

sondreso:
Jenkins test this please

PR #974, opened 2020-08-05
Deprecate CUSTOM_KW

### LINKED ISSUE #909: Deprecate CUSTOM_KW
`CUSTOM_KW` is not, and have never been, functioning as intended. We want a key-value data type in ERT, but it is less effort to design one from scratch based on `JSON` and the new storage implementation. Having `CUSTOM_KW` in your configuration file should yield a deprecation warning, all code related to `CUSTOM_KW` can be removed.

ManInFez:
**Issue**
Resolves equinor/ert#909


markusdregi (approved):
:+1:

PR #976, opened 2020-08-05
Remove LSF, RSH and LOCAL queue KW

### LINKED ISSUE #933: Old style queue options are ignored
Old style queue options like `LFS_QUEUE mr` has been replaced by newer syntax `QUEUE_OPTION LSF LSF_QUEUE mr`. The former is now ignored in the code, but no warning nor error is given. A decision has been made to completely remove the support for old style queue options so that users will be prompted with the warning:

```
** Warning keyword: QUEUE_OPTION not recognized when parsing: super_config.ert --- 
```

instead of nothing but silence. The keyword and it's siblings are added to the schema here:
https://github.com/equinor/libres/blob/e34e4f767fb475a711bdbec4339c1878379da5ff/lib/enkf/site_config.cpp#L384

But we should in addition strive to remove the code parsing the old style syntax if there are still traces of it.

ManInFez:
**Issue**
Resolves #933

### DIFF
--- a/lib/enkf/config_keys.cpp
+++ b/lib/enkf/config_keys.cpp
@@ -46,18 +46,6 @@ const char * config_keys_get_queue_option_key() {
   return QUEUE_OPTION_KEY;
 }
 
-const char * config_keys_get_lsf_resources_key() {
-  return LSF_RESOURCES_KEY;
-}
-
-const char * config_keys_get_lsf_server_key() {
-  return LSF_SERVER_KEY;
-}
-
-const char * config_keys_get_lsf_queue_key() {
-  return LSF_QUEUE_KEY;
-}
-
 const char * config_keys_get_install_job_key() {
   return INSTALL_JOB_KEY;
 }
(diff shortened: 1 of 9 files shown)


lib/enkf/queue_config.cpp
```diff
@@ -367,3 +367,20 @@ void queue_config_add_config_items(config_parser_type * parser, bool site_mode)
...
+  return LSF_SERVER;
+}
+
+char * queue_config_lsf_resource() {
+  return LSF_RESOURCE;
+}
+
+char * queue_config_lsf_driver_name() {
+  return LSF_DRIVER_NAME;
+}
```
oysteoh: would have been nice with a newline

ManInFez: DOH!

ManInFez: fixed :)

lib/enkf/config_keys.cpp
```diff
@@ -46,18 +46,6 @@ const char * config_keys_get_queue_option_key() {
   return QUEUE_OPTION_KEY;
 }
 
-const char * config_keys_get_lsf_resources_key() {
```
oysteoh, line 49: should these names also be removed from `config_keys.hpp`?

and from `config_keys.py`?

ManInFez:
Test this please!

lib/enkf/config_keys.cpp
ManInFez: Yes! good catch :+1:

oysteoh (approved):
LGTM!

PR #977, opened 2020-08-05
Round MIN_REALIZATIONS up if proportion kw is used (min 1 simulation)

### LINKED ISSUE #609: MIN_REALIZATIONS does not work as expected for small ensemble sizes
Kw MIN_REALIZATIONS does not work as expected if a percentage is set and the ensemble size is small. Eg MIN_REALIZATIONS 2%, will halt if 4 out of 5 realizations are successful. Clearly not as advocated. It does however work as expected when ensemble size is big.

Running small ensemble sizes makes sense during testing.

hnformentin:
**Issue**
Resolves #609

### DIFF
--- a/lib/enkf/tests/enkf_analysis_config.cpp
+++ b/lib/enkf/tests/enkf_analysis_config.cpp
@@ -132,6 +132,12 @@ void test_min_realizations_percent() {
     test_min_realizations(num_realizations_str, min_realizations_str, min_realizations_expected_needed);
   }
   {
+    const char * num_realizations_str = "NUM_REALIZATIONS 5\n";
+    const char * min_realizations_str = "MIN_REALIZATIONS 2%\n";
+    int min_realizations_expected_needed              = 1; 
+    test_min_realizations(num_realizations_str, min_realizations_str, min_realizations_expected_needed);
+  }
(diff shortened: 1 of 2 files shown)


hnformentin:
If approved, it will follow with a modification in the documentation:
"Note that the integer value of eg 2% of 20 realizations is 0, and that it is easy to unintentionally instruct ERT to require all simulations to be finished when the intention was the opposite."
by 
"Note that the integer value of eg 2% of 20 realizations is 0 but it will be rounded up to 1."

lib/enkf/tests/enkf_analysis_config.cpp
```diff
@@ -131,6 +131,12 @@ void test_min_realizations_percent() {
     int min_realizations_expected_needed              = 8;
     test_min_realizations(num_realizations_str, min_realizations_str, min_realizations_expected_needed);
   }
+  {
```
ManInFez, line 134: This seems like unnecessary scoping, but it is follows the style of the file at least ;)

ManInFez (approved):
Looks good. It now always rounds up, which may be unexpected. A different solution would be to do proper rounding but set minimum value of 1.

PR #982, opened 2020-08-07
Use ENKF_ALPHA key instead of OVERLAP_LIMIT

oysteoh:


lib/enkf/analysis_config.cpp
```diff
@@ -44,7 +44,7 @@
 #include <ert/enkf/model_config.hpp>
 
 
-#define UPDATE_OVERLAP_KEY      "OVERLAP_LIMIT"
+#define UPDATE_ENKF_ALPHA_KEY      "ENKF_ALPAHA"
```
markusdregi: May I suggest that we change it to simply `ALPHA`, instead of `ENKF_ALPAHA`

oysteoh: hmmm... i'm not sure... ? if people have used enkf_alpha in their configs up until now - i guess we will brake it by changing name? 

feel free to hijack and fix if possible - always nice to get rid of `enkf` if it makes sense !

oysteoh:
By manipulating ENKF_ALPHA either through only ENKF_ALPHA or UPDATE_SETTING ENKF_ALPHA I'm now able to change the threshold where a observation gets deactivated or not. 

The naming - do you want me to change to only ALPHA like  you commented - or is it better to preserve behavior in peoples configurations?  @markusdregi

lib/enkf/analysis_config.cpp
markusdregi: I agree. Let us keep it as is 👍

lib/enkf/analysis_config.cpp
```diff
@@ -430,7 +430,7 @@ void analysis_config_init( analysis_config_type * analysis , const config_conten
     analysis_config_set_std_cutoff( analysis , config_content_get_value_as_double( config , STD_CUTOFF_KEY ));
 
   if (config_content_has_item( config , ENKF_ALPHA_KEY ))
-    analysis_config_set_alpha( analysis , config_content_get_value_as_double( config , ENKF_ALPHA_KEY ));
+    fprintf(stdout, "Setting ENKF_ALPHA directly is deprecated and should be changed to UPDATE_SETTING ENKF_ALPHA to be taken into consideration!");
```
sondreso: Is it possible to use `config_parser_deprecate` as it is done [here](https://github.com/equinor/libres/blob/83f59a78471591ff105118ddd7823e53554f0c68/lib/enkf/analysis_config.cpp#L692)?

oysteoh: will have a look! thx 👍

sondreso (approved):

sondreso:
Jenkins test this please

PR #989, opened 2020-08-19
Print update_log even if points are missing or locally inactive

### LINKED ISSUE #988: Observation name not printed in update_log
The observation name is not printed in the update_log file if the first entry is `MISSING` or `LOCAL_INACTIVE`

Possible solutions:
 1) Print all observations regardless of status (this was the previous behavior, but was changed 4 years ago)
 2) Print all observation names, and never print `...` even if the name is repeated,
 3) Print nan for simulated data

Relevant code: https://github.com/equinor/libres/blob/59bc1a14b9a18f841395672d7a3e5d5455a5243d/lib/enkf/enkf_analysis.cpp#L44

oyvindeide:
Closes #988 

My main concern with this particular fix is if that `if ((active_mode == MISSING) || (active_mode == LOCAL_INACTIVE))` was present to stop some bug occuring, but can find no discussion to suggest that. There observation vector should also be fully initialized so should not be a problem to read from that.

### DIFF
--- a/lib/enkf/enkf_analysis.cpp
+++ b/lib/enkf/enkf_analysis.cpp
@@ -43,6 +43,9 @@
 
 void enkf_analysis_fprintf_obs_summary(const obs_data_type * obs_data , const meas_data_type * meas_data , const int_vector_type * step_list , const char * ministep_name , FILE * stream ) {
   const char * float_fmt = "%15.3f";
+  bool local_inactive_obs;
+  local_inactive_obs = false;
+
   fprintf(stream , "===============================================================================================================================\n");
   fprintf(stream , "Report step...: %04d",int_vector_iget( step_list , 0));
   if (int_vector_size( step_list ) == 1)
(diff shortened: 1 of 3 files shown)


markusdregi:
From the changes in the code, but no changes in the tests I'm assuming this is not very well tested. Could you add one? :) That would be quite valuable when reviewing the changes as well!

oyvindeide:
> From the changes in the code, but no changes in the tests I'm assuming this is not very well tested. Could you add one? :) That would be quite valuable when reviewing the changes as well!

I was afraid you would say that 🤣 No, it is not tested what so ever, should be possible to add an integration test. Unit test might be a bit difficult as it more or less requires a complete setup to run.

lib/enkf/enkf_analysis.cpp
```diff
@@ -70,37 +70,37 @@ void enkf_analysis_fprintf_obs_summary(const obs_data_type * obs_data , const me
...
+            print_key = "  ...";
+
+          fprintf(stream , obs_fmt , obs_count , print_key , obs_block_iget_value( obs_block , iobs ) , obs_block_iget_std( obs_block , iobs ));
+
+          if (active_mode == ACTIVE)
+            fprintf(stream , "  Active   |");
+          else if (active_mode == DEACTIVATED)
+            fprintf(stream , "  Inactive |");
+          else if (active_mode == LOCAL_INACTIVE)
+            fprintf(stream , "  Inactive |");
```
sondreso: Why are we printing the same status for both `DEACTIVATED` and `LOCAL_INACTIVE`? Think we could avoid some confusion on the future by differentiating in the log statement

sondreso (approved):
LGTM!

oyvindeide:
test this please

PR #990, opened 2020-08-25
subprocess tee: check process end *after* reading

pinkwah:
Jenkins has a problem with the unit test when running RHEL7&Py3. The output files are empty.

The problem might be due to a race condition where the process ends before getting to `await_process_tee`. Change that function to read first and ask questions later.


ManInFez (approved):
Shipit!

ManInFez:
Test this please!

PR #991, opened 2020-08-28
Disable black in komodo tests

mortalisk:
We don't need black in komodo tests


ci/jenkins/testkomodo.sh
```diff
@@ -72,3 +72,4 @@ python -m pytest \
  --ignore="tests/res/enkf/test_workflow_list.py"\
  --ignore="tests/res/enkf/test_hook_manager.py"\
  --ignore="tests/legacy"
```
pinkwah: need a `\` at the end of this line

oysteoh (approved):

Issue #993, opened 2020-09-08
Expose file operation jobs as workflows


markusdregi:
The job located in https://github.com/equinor/libres/tree/master/share/ert/forward-models/shell should be exposed as workflow jobs as well to facilitate creation of folders, etc on the "client" side.

jepebe:
It seems that adding job descriptions for workflows is simple but sharing the underlying scripts might not be. Alternatives:
1. Duplicate the scripts
2. Use the same scripts but with `../../`directory paths.
3. Add an "environment" variable in the _site-config_ that the job descriptions can use to look up the shell scripts. (I do not think this already exists.)

Alternatives ordered by effort.

PR #994, opened 2020-09-09
libres-993 Expose file system Forward Model jobs as Workflow jobs

jepebe:
There are several Forward Model jobs that give access to file system operations. Having access to these as Workflow jobs is useful. This PR solves this by moving the common scripts to a higher level and creating Workflow job definitions. 

This solves #993


ertomatic:

pinkwah:
Jenkins test this please

python/tests/share/test_shell_jobs.py
```diff
@@ -0,0 +1,30 @@
...
+                fm_shell_jobs[job.name().lower()] = exe
+
+        list_from_content = res_config.ert_workflow_list
+        wf_shell_jobs = {}
+        for wf_name in list_from_content.getJobNames():
+            exe = list_from_content.getJob(wf_name).executable()
+            if exe and exe.startswith(share_path):
+                wf_shell_jobs[wf_name] = exe
+
+        assert fm_shell_jobs == wf_shell_jobs
```
mortalisk, line 27: On my machine, both these dicts end up as empty when I run this test. ( I assume they are supposed to contain something )

python/tests/share/test_shell_jobs.py
```diff
@@ -0,0 +1,30 @@
+from res.job_queue import ExtJob
```
mortalisk: Some unused imports

python/tests/share/test_shell_jobs.py
jepebe: You need the `site-config` to include the workflows in the *share* directory. But I will add a test with names as well.

mortalisk: Can we make the site-config explicit in the test, so that it runs the same everywhere?

jepebe: Well it works on my machine...

python/tests/share/test_shell_jobs.py
```diff
@@ -0,0 +1,55 @@
...
+        list_from_content = res_config.ert_workflow_list
+        wf_shell_jobs = {}
+        for wf_name in list_from_content.getJobNames():
+            exe = list_from_content.getJob(wf_name).executable()
+            if exe and exe.startswith(share_path):
+                wf_shell_jobs[wf_name] = exe
+
+        assert fm_shell_jobs == wf_shell_jobs
+
+    def test_shell_script_jobs_names(self):
```
mortalisk, line 29: This test fails on my machine. How do I use the correct site-config?

jepebe: Working on it. :) "It works on my machine" as in failing on Jenkins and Travis

python/tests/share/test_shell_jobs.py
jepebe: Maybe a bug but `createSharePath` and the actual executable location differ when running on Travis/Jenkins. The share path is located outside of the **venv** and the executable is inside it. I have worked around that issue in the updated test.

share/ert/site-config
```diff
@@ -1,5 +1,6 @@
 -- This is a very basic site config file, mainly used for testing.
 
+WORKFLOW_JOB_DIRECTORY workflows/jobs/shell
```
sondreso, line 3: The equivalent of this line also needs to be added here ([link](https://github.com/equinor/ert/blob/8f2e7181123c53555520b1d3359b15c893b3b5e3/ert_shared/hook_implementations/jobs.py#L103)) for the plugin system to install them 🙂

mortalisk: Will you make an issue for doing that as well? After doing that this looks good to me.

mortalisk (approved):

share/ert/site-config
jepebe: Already made a PR Equinor/ert#1029.

PR #998, opened 2020-09-11
Remove unused JobQueueManager code in C

jepebe:
This PR removes the C code for JobQueueManager. At the same time several Workflows that required this class to be available has been removed.


ertomatic:

Issue #975, opened 2020-08-05
Remove deprecated keyword CUSTOM_KW


ManInFez:
CUSTOM_KW was tagged deprecated 05.08.2020 and should be removed after an appropriate amount of time has passed.

PR #1001, opened 2020-09-14
libres-975: Remove deprecated keyword CUSTOM_KW

jepebe:
This PR removes all references to the CUSTOM_KW feature in *libres*. In addition, most of the touched Python files have been reformatted with Black.

This solves #975


ertomatic:

markusdregi:
This is a nice effort to get rid of some unfortunately not used code 👏 There seems to be some odd indents in the CMAKE files where custom_kw has been removed.

Also, could you separate the black part into a second PR as this probably needs some discussion? So far we have avoided blacking ERT and libres due to the above massive changes it induces...

markusdregi (approved):
This is really valuable 👏

PR #1004, opened 2020-09-18
Remove Python 3.8 implicit conversion to int warnings

jepebe:
In Python 3.8 this warning started to appear in `libres`:

`DeprecationWarning: an integer is required (got type ...).  Implicit conversion to integers using _int_ is deprecated, and may be removed in a future version of Python.`

This PR fix these warnings by leveraging the already available enums so that the implicit conversion goes away.

### DIFF
--- a/lib/enkf/queue_config.cpp
+++ b/lib/enkf/queue_config.cpp
@@ -112,7 +112,7 @@ queue_config_type * queue_config_alloc_full(char * job_script,
                                             bool user_mode,
                                             int max_submit,
                                             int num_cpu,
-                                            int driver_type) {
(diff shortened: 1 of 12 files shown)


ertomatic:

pinkwah:
~I don't think it's a good idea to remove type information from the function prototypes. Perhaps it's a better idea to patch cwrap to be Python 3.8 compatible?~

Edit: stupid

jepebe:
I didn't *remove* type information I have *added* type information. It is now consistent with the type in C as well.

pinkwah (approved):
I should learn to read diffs. :rofl: 

LGTM

PR #1005, opened 2020-09-23
Cleanup guided by code inspection

jepebe:
Several code inspections tools has revealed that parts of the code is unused or has other shortcomings.


lib/analysis/modules/ies_enkf_data.c
```diff
@@ -32,14 +29,7 @@
 struct ies_enkf_data_struct {
    UTIL_TYPE_ID_DECLARATION;
    int       iteration_nr;            // Keep track of the outer iteration loop
-   double    ies_max_steplength;      // Maximum step length in Gauss Newton iteration
```
pinkwah, line 35: Are we sure deleting these fields won't give us an ABI breakage wrt. stuff that exists in people's `storage/`?

lib/analysis/stepwise.cpp
```diff
@@ -21,12 +17,10 @@ struct stepwise_struct {
   matrix_type      * Y0;
 
   matrix_type      * beta;           // Quantities estimated by the stepwise algorithm
-  double             Y_mean;
```
pinkwah, line 24: Same question about ABI

lib/enkf/block_obs.cpp
```diff
@@ -358,11 +348,11 @@ double block_obs_chi2(const block_obs_type * block_obs, const void * state, node
 /**
    The index is into the the number of active cells which are observed by this observation.
 */
-void block_obs_iget(const block_obs_type * block_obs, int index , double *value , double * std) {
-  const point_obs_type * point_obs = block_obs_iget_point_const( block_obs , index );
-  *value = point_obs->value;
-  *std   = point_obs->std;
-}
+//void block_obs_iget(const block_obs_type * block_obs, int index , double *value , double * std) {
```
pinkwah: I realise this is a draft, but just pointing out the commented-out code here just in case.

lib/enkf/ecl_config.cpp
```diff
@@ -133,13 +124,17 @@ ui_return_type * ecl_config_validate_data_file(const ecl_config_type * ecl_confi
...
+}
+
 
 void ecl_config_set_data_file(ecl_config_type * ecl_config, const char * data_file) {
   ecl_config->data_file = util_realloc_string_copy(ecl_config->data_file, data_file);
   {
     FILE * stream = util_fopen(ecl_config->data_file, "r");
     basic_parser_type * parser = basic_parser_alloc(NULL, NULL, NULL, NULL, "--", "\n");
-    char * init_tag = enkf_util_alloc_tagged_string("INIT");
+    char * init_tag = alloc_tagged_string("INIT");
```
pinkwah: Consider inlining as `alloc_tagged_string` occurs only once.

```c++
const char * init_tag = DEFAULT_START_TAG "INIT" DEFAULT_END_TAG;
/* Or better yet: */
auto init_tag = "<INIT>";
```
and remove the `free(init_tag);`

lib/enkf/enkf_state.cpp
```diff
@@ -779,7 +693,7 @@ bool enkf_state_complete_forward_modelOK(const res_config_type * res_config,
 
   }
 
-  return (result == 0) ? true : false;
+  return (result == 0);
```
pinkwah: `return result == 0;`

pinkwah (changes requested):
Partial review. Github was quite unhelpful in opening the files and my laptop's very unhappy with me right now.

lib/analysis/modules/ies_enkf_data.c
jepebe: Are the structures dumped directly to disk?

jepebe: I may be misunderstanding you but I do not think any of the read/write mechanisms use `structs` directly to/from disk.

jepebe:
This PR was divided into 3 smaller PRs: #1017, #1018 and #1019 to simplify reviewing :)

PR #1006, opened 2020-09-24
Change order of std-cutoff and alpha

jepebe:
The order of _std-cutoff_ and _alpha_ are wrong and this PR fixes that.

For refeference this is the declaration of the function call.
```
double enkf_obs_scale_correlated_std(const enkf_obs_type * enkf_obs,
                                     enkf_fs_type * fs,
                                     const int_vector_type * ens_active_list,
                                     const local_obsdata_type * local_obsdata,
                                     double alpha,
                                     double std_cutoff,
                                     bool verbose) {
```


markusdregi (approved):
Good catch!

I'll follow up to see if we can not deprecate the code being used 👍

PR #1007, opened 2020-09-24
Bump libecl to 2.9.2rc1

sondreso:
Bump libecl to 2.9.2rc1

### DIFF
--- a/.libecl_version
+++ b/.libecl_version
@@ -1,1 +1,1 @@
-export LIBECL_VERSION=2.9.1
+export LIBECL_VERSION=2.9.2rc1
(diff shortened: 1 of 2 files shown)


sondreso:
Jenkins test this please

oyvindeide (dismissed):

oyvindeide:
I am guessing the failed job is not related to this PR?

sondreso:
Thanks! 
Nope, same error as @jepebe and the new release is experiencing, trying to debug...

sondreso:
Jenkins test this please

hnformentin (approved):

jepebe:
Not necessarily part of this PR but would it make sense to have the `wait_until()` print an error message or, alternatively, wrap the inner assertion of the test in a `TimeoutAssertionError` instead? Would have made the bug easier to diagnose since it would actually pinpoint the reason for failing.

sondreso:
> Not necessarily part of this PR but would it make sense to have the `wait_until()` print an error message or, alternatively, wrap the inner assertion of the test in a `TimeoutAssertionError` instead? Would have made the bug easier to diagnose since it would actually pinpoint the reason for failing.

Yes, I think that is a very good idea! Will merge this so it's not blocking everything, but also created this issue: https://github.com/equinor/libres/issues/1009

PR #1008, opened 2020-09-24
Deprecate STD_SCALE_CORRELATED_OBS keyword

jepebe:
A better method for performing the same operation is available so a deprecation warning has been introduced.


sondreso (approved):
:rocket:

PR #1010, opened 2020-09-25
Remove unused PCA plot data classes

jepebe:
@oyvindeide pointed out that the PCA plot data classes in C and Python were unused due to other means of achieving the same thing. This PR removes all of the unused code.


python/res/enkf/plot/__init__.py
```diff
@@ -6,5 +6,4 @@
 from .block_observation_data_fetcher import BlockObservationDataFetcher
 from .ensemble_gen_kw_fetcher import EnsembleGenKWFetcher
 from .ensemble_gen_data_fetcher import EnsembleGenDataFetcher
-from .observation_gen_data_fetcher import ObservationGenDataFetcher
-from .pca_fetcher import PcaDataFetcher
```
oyvindeide, line 10: This is a missing newline here

oyvindeide (approved):
Cant find any references in any other projects either. Thanks for cleaning up!

PR #1011, opened 2020-09-25
Remove unused sched module

jepebe:
The sched module in libres had multiple classes that were not used by the rest of library. This PR removes those.


python/res/sched/CMakeLists.txt
```diff
@@ -2,8 +2,7 @@ set(PYTHON_SOURCES
     __init__.py
     history.py
     history_source_enum.py
-    sched_file.py
-)
+        )
```
markusdregi: Some odd indentation 🤷

markusdregi (commented):
This looks good 👍 Found an odd indentation; when that is fixed this one is good to go!

python/res/sched/CMakeLists.txt
jepebe: Yes, that was odd.

markusdregi (approved):
🚢

PR #1015, opened 2020-09-28
Reduce processing time for Simulation Context test

jepebe:
Simulation Context uses a long time to complete, mostly due to a synchronization issue in `file.py`. However, changing that will change untested behavior of the queue system. This PR changes the processing time of the jobs in the test to minimize the total running time of the test. This will hopefully reduce any flakiness due to long running time.


jepebe:
I believe that the old job used somewhere around 5-10 seconds per run (maybe more on a busy machine). In total 4 jobs were set up to run in two different `SimulationContext` instances. The worst case scenario is when each job took, for example, 10 seconds and ended up running sequentially. There are several delays built in to the queue system that are in addition to the running time (for example 10 seconds in `file.py`). So worst case sequential running time would be at least 80 seconds. 

This PR removes any work in the actual job - replaced with a 1 second sleep so that any waiting time is due to the queue system.

sondreso (approved):
LGTM :running_man: :1st_place_medal:

PR #1018, opened 2020-09-30
Remove unused declarations and implementations

jepebe:
This PR is step 2 of PR #1005. In this PR any global declarations and unused implementations are removed without ceremony.


pinkwah (approved):
:1st_place_medal:

PR #1019, opened 2020-09-30
Remove unused includes

jepebe:
This is the final part of the PR #1005. This removes all unused includes from the C code without ceremony and regard.


lib/config/config_parser.cpp
```diff
@@ -137,20 +126,9 @@ struct config_parser_struct {
...
-
-
-
-
-
-
-
-
-
+int config_get_schema_size( const config_parser_type * config ) {
```
pinkwah, line 129: Where does this come from?

jepebe: The deleted file `config_get.cpp`

pinkwah (approved):
:parrot:

PR #1021, opened 2020-10-01
Remove etc-example directory

sondreso:
The contents of this folder is not up to date with the development of the project, and in it's current state it's more a source of error than guidance.


markusdregi (approved):
Good initiative 👏

PR #1023, opened 2020-10-01
Add PRE_FIRST_UPDATE hook

oyvindeide:
There is sometimes a need for a hook that will only run before the first update and not subsequent times for an iterative algorithm. This is for example relevant if scaling observations, as doing that for each iteration could lead to numeric  instabilities.


lars-petter-hauge:
Alright, looks good. Also did a manual test together with https://github.com/equinor/ert/pull/1044

Ran different modes, repeated and verified that the behaviour is as intended.

lars-petter-hauge (approved):

PR #1024, opened 2020-10-01
Black all python files and add Github Actions check

sondreso:
**Approach**
Run black on all files inside `python`, remove black test in python tests and add Github Action check for black formatting.


oyvindeide:
I agree with doing this, it is annoying to just have a subset being `blacked`. There is a way to avoid it cluttering blame: https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame

jepebe:
In principal I agree, however, I know that some of these files would probably look much better with some human love before reformatting with black. Since it is 306 files I think maybe leave it as an exercise for the next person that enters the file?

markusdregi:
> I agree with doing this, it is annoying to just have a subset being `blacked`. There is a way to avoid it cluttering blame: https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame

I agree, besides that that does not really help you when navigating the code on GitHub :)

markusdregi:
I used to think that this was completely unheard of, as it will take a toll on the git history. However, I think it will never be completed unless we just do it 🤷 

I do agree with Jean-Paul on some manual aiding of the formatting though. Would it be worth it to do it module by module and add some manual labor to verify that the formatting is not completely crazy?

sondreso:
> I think it will never be completed unless we just do it

This was my realization as well! 

> Would it be worth it to do it module by module and add some manual labor to verify that the formatting is not completely crazy?

I don't really see how we are going to do this. Because unless I'm missing something, then `better formatting -> refactoring`. And since not all of these files are in parts of the code that is currently prioritized in the ERT project, and think the best way is just ripping of the band-aid fast.

markusdregi:
Rebase, split it into two commits and I'm inclined to approve this ;)

sondreso:
Done! If this is merged I will add a second PR with the suggestions from here: https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame. But this will need to wait until we have the correct hash on master!

test_requirements.txt
```diff
@@ -1,5 +1,5 @@
 pytest
 decorator
 pylint
-black==19.10b0
+black
```
markusdregi, line 4: I'm being a bit pedantic now, but that is the best I can do when we are to apply a uncompromising code formatter 🤷 I don't think this, nor the removal of the formatting test should be touched in this commit. I would like it to purely be the result of applying black to the source code.

Maybe we can do something like this with the commits(?):
- Bump black to 19.10b0 (and update currently blacked lines)
- Apply black on all files (<- point is I want this to be a PR with no human edits)
- Move black test out of test_formatting and into CI

sondreso: Done :D

markusdregi (approved):
Just what I had in mind 👏 Much easier to review!

Great initiative!

PR #1030, opened 2020-10-12
Use subprocess for rms

ManInFez:
Closes https://github.com/equinor/libres/pull/1030

use wrapper script for rms if available

Should fix the RMS issues by executing rms through a komodo-disable-script defined in the rms_config.yml file

### DIFF
--- a/python/res/fm/rms/rms_run.py
+++ b/python/res/fm/rms/rms_run.py
@@ -175,10 +175,7 @@ class RMSRun(object):
             args += ["-threads", str(self.config.threads)]
 
         if not exec_env:
-            exec_env = os.environ.copy()
-        exec_env["_PRE_RMS_BACKUP"] = "1"
-        if "PYTHONPATH" in os.environ:
-            exec_env["_PRE_RMS_PYTHONPATH"] = os.environ["PYTHONPATH"]
+            exec_env = os.environ
 
         comp_process = subprocess.run(args=args, env=exec_env)
         return comp_process.returncode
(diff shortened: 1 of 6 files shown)


lars-petter-hauge:
Throwing os.fork out and use subprocess instead, sounds good to me! Unsure how much tested the RMS stuff in libres is though

python/res/fm/rms/rms_run.py
```diff
@@ -155,7 +145,10 @@ def wait(self, child_pid):
             )
 
     def exec_rms(self, exec_env):
-        args = [
+        args = []
+        if self.config.wrapper is not None:
+            args = [self.config.wrapper]
```
sondreso: Maybe like this? 
```suggestion
        args = [] if self.config.wrapper is None else [self.config.wrapper]
```

python/res/fm/rms/rms_run.py
```diff
@@ -155,7 +145,10 @@ def wait(self, child_pid):
...
+        args += [
```
sondreso: or like this:
```suggestion
        args = []
        if self.config.wrapper is not None:
            args.append(self.config.wrapper)
        args.extend([
```

NB: Missing ) at the end

sondreso:
I think the commit message is a little misleading since the wrapper functionality is introduced in addition to the fork change.  Would it be possible to split them into two commits and/or explain why the wrapper is introduced in commit body?

sondreso (approved):
LGTM!

PR #1031, opened 2020-10-13
Add error message in wait_until function

### LINKED ISSUE #1009: Better error message for timeouts in test_simulation_context
If the following timeout is hit, it is difficult to pinpoint the reason for failing: 
https://github.com/equinor/libres/blob/f17ca13f1fa0c8d7abdf5cf6bfcb753d763a9e05/python/tests/res/simulator/test_simulation_context.py#L49-L52 
We should have the `wait_until()` print an error message or, alternatively, wrap the inner assertion of the test in a `TimeoutAssertionError` instead.

hnformentin:
**Issue**
Resolves #1009 

**Approach**
Added a message when raising AssertionError in /libres/python/tests/utils/__init__.py.

### DIFF
--- a/python/tests/utils/__init__.py
+++ b/python/tests/utils/__init__.py
@@ -81,4 +81,8 @@ def wait_until(func, interval=0.5, timeout=30):
             return
         except AssertionError:
             if t >= timeout:
-                raise
+                raise AssertionError(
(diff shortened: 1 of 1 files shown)


sondreso (approved):
LGTM!

sondreso:
Remember to squash before merging :slightly_smiling_face:

PR #1040, opened 2020-10-15
Remove unused quantile functions

jepebe:
Two quantile functions pointing to the same C endpoint was only used in a test. This PR remove these from Python.


markusdregi (commented):
I vote that `res.util.stat` is not part of the public API and that we hence can remove it without it being a breaking change. What is your take on this @sondreso ?

markusdregi (approved):

sondreso (approved):
I agree with @markusdregi!

PR #1042, opened 2020-10-15
Facilitate fetching argument values for jobs

### LINKED ISSUE #1039: Expose argument values for jobs in python
The `ext_job_get_arglist` provides the content of `ARGLIST` from the job description files (i.e. `FROM` and `TO` for the `COPY_FILE` job). Extend the interface to also allow the user to retrieve the values of those arguments.

lars-petter-hauge:
**Issue**
Resolves #1039

**Approach**
The values for the arguments for a configured job was not exposed, added another function to facilitate retrieving the values from python.

Probably not handling the allocation correctly - must be checked.

Tests not added!


lars-petter-hauge:
Yeah.. wanted to use regular `std::vector` instead of `stringlist`.. 

```
raise PrototypeError("Illegal prototype definition: %s\n" % self._prototype)
cwrap.prototype.PrototypeError: Illegal prototype definition: std::vector<std::string> ext_job_get_argvalues(ext_job)
```

Having some `Prototype` issues, I'll see if I can't get rid of those

hnformentin:
Why would you prefer `std::vector<std::string> `instead of `stringlist_type`?

lars-petter-hauge:
`StringList` is a type we have made our selves, I would very much prefer to use a type that actually comes from the standard library. It should be mentioned that when this library was written, it was all in C. We are now converting it (slowly..) to C++, and then we definitively should use what exists in the standard library instead.

lars-petter-hauge:
Yeh, I so it seems I didn't think this one through.. cwrap uses c-types, which doesn't support c++ containers. I'm removing my type change commit

lars-petter-hauge:
So, what we are a bit concerned about is that the `get_arglist()` seems to provide different results depending of whether it has been initiated as part of the `ResConfig` initialisation or if it is initialised in the basis of a `jobs.json` file.

Given that we have an ert configuration file with the `FORWARD_MODEL COPY_DIRECTORY(<FROM>="some_dir", <TO>="some_other_dir")` job, the `jobs.json` file will contain only the `"argList" : ["some_dir","some_other_dir"]`.

If we retrieve the `ExtJob` from a `ResConfig` object initialised from the ert configuration file, we will get:
`job.get_arglist() == ["<FROM>", "<TO>"]` - like we see in `EnKFTestSimModel(ResTest)`, where it asserts on a keyword definition from the job description file: https://github.com/equinor/libres/blob/58e69d959f12ead77f68baa358eec7a0cff27a97/python/tests/res/enkf/test_enkf_sim_model.py#L95

If we on the other hand retrieve it after the instance has been created from a `jobs.json` file (e.g. as in `ForwardModelFormattedPrintTest(ResTest)`), the `get_arglist()` will provide `["some_dir", "some_other_dir"]`.

For any ert configurations using `SIMULATION_JOB`, these two will be identical. However, for the `FORWARD_MODEL` jobs that include a keyword argument (later modified with the `subst_list`?), they will be different.

The last issue that arise is that the `get_argvalues` will also change, as it will in the first case (from ert config file) provide `["some_dir", "some_other_dir"]`, while in the second case it will be an empty list.

Edit: updated output of `get_arglist()`

lars-petter-hauge:
Ref jenkins:

`E   AttributeError: /prog/res/komodo/bleeding-py36-rhel7/root/lib64/libres.so.7.0: undefined symbol: ext_job_get_argvalues`

`E   cwrap.prototype.PrototypeError: Can not find function: ext_job_get_argvalues in library: <CDLL 'libres.so.7.0', handle 3c92410 at 0x7f15638f5c88>`

`libres` is built in the pipeline, but it's using the old `so` file? which obviously doesn't have the latest addition. 
Perhaps the `LD_LIBRARY_PATH` isn't updated?

sondreso:
`/prog/res/komodo/bleeding-py36-rhel7/root/lib64/libres.so.7.0` so yes, it's picking up the wrong so file :slightly_smiling_face:

lars-petter-hauge:
From the build script: `export LD_LIBRARY_PATH=$CI_SOURCE_ROOT/cmake-build/lib:$CI_SOURCE_ROOT/cmake-build/lib64:${LD_LIBRARY_PATH:-}`

sondreso:
> If we retrieve the ExtJob from a ResConfig object initialised from the ert configuration file, we will get:
`job.get_arglist() == ["FROM", "TO"]`

Should this be `job.get_arglist() == ["<FROM>", "<TO>"]`?

lars-petter-hauge:
Yepp, updated my original comment

lars-petter-hauge:
Looks like I was slightly off here.

> If we on the other hand retrieve it after the instance has been created from a jobs.json file (e.g. as in `ForwardModelFormattedPrintTest(ResTest))`

This is not a thing, I just misunderstood the intent of the `ForwardModelForm..´ test.

This implementation should really do what the `ext_job_json_fprintf` does when writing the argument list in `__fprintf_python_argList`. Instead of iterating over the `private_args`, we iterate the `args` and substitute any necessary args in `private_args`. This will handle the cases where arguments are made as defaults in the job configuration file and never accessed by the user.

The only issue really is that the `global_vars` are not a part of the `ext_job`, but rather passed down through `forward_model_formatted_fprintf` by using the `run_arg` (`run_arg_get_subst_list(run_arg)`). We don't have that option within the structure itself.

lars-petter-hauge:
Last commit will fail, but I just threw out what's there while we discuss

lars-petter-hauge:
Will still fail on anything using key/value pairs from somewhere in the ert config outside of the private arguments (e.g. `ECL_BASE`/`NUM_CPU`).

Must also test that `<ITER>`, `<IENS>` (~is `<REAL>` another on?~) is passed correctly.

lars-petter-hauge:
After some testing we believe that getting argument values from `runargs` isn't really achievable. So the `ITER` and `IENS`, which are state dependent, might not be possible to include. It is unclear about config keywords, such as `ECLBASE` or `NUM_CPU`.

We decided to revert the modifications to the original implementation, and thus we will have duplicate code for this part. We will rename to `define_args` or something instead of `global_args` though, just to not make it more confusing.

lars-petter-hauge:
We encountered some slight unexpected behaviour when making the test - or perhaps expected from the code, but does the user expect it? We've separated the changes necessary for half the test to pass in a single commit: https://github.com/equinor/libres/pull/1042/commits/21e9a1e7e4d39af8df8453fd359c8c44564708f9

lib/enkf/model_config.cpp
```diff
@@ -506,6 +508,7 @@ void model_config_init(model_config_type * model_config ,
 
     model_config->obs_config_file = util_alloc_string_copy(obs_config_file);
   }
+  // Only subst_list_free for non const. Free const? (define_list);
```
ManInFez: config should be responsible for cleaning up define_list

python/tests/res/enkf/test_enkf_sim_model.py
```diff
@@ -102,6 +103,28 @@ def test_simulation_model(self):
                     forward_model.iget_job(2).get_arglist(), ["word", "<ECLBASE>"]
                 )
 
+                self.assertEqual(forward_model.iget_job(0).get_argvalues(), ["yy"])
+                self.assertEqual(
```
ManInFez, line 107: is this not a copy of line 98?

python/res/job_queue/ext_job.py
```diff
@@ -202,6 +203,9 @@ def get_license_path(self):
     def get_arglist(self):
         return self._get_arglist()
 
+    def get_argvalues(self):
+        return self._get_argvalues()
```
ManInFez: should we make this a list and not have a stringList moving around in the python layer?

lib/job_queue/ext_job.cpp
```diff
@@ -1107,6 +1113,21 @@ const stringlist_type * ext_job_get_arglist( const ext_job_type * ext_job ) {
     return ext_job->argv;
 }
 
+const stringlist_type * ext_job_get_argvalues( const ext_job_type * ext_job ) {
```
ManInFez, line 1116: would it not make sense to create this when setting the arguments and not every time when extracting?

ManInFez (commented):
Looks good! just a few things. Let me know if you want to discuss :)

python/res/job_queue/ext_job.py
lars-petter-hauge: I would second that, I did it to be consistent with `get_arglist`, but perhaps I can change both instead?

lib/job_queue/ext_job.cpp
lars-petter-hauge: It would require the `set_private_args` to be called before the `set_defined_args`, so I'm not sure.. We could do some checking if the `arg_values` have been resolved, it would add more state to the structure though.

lib/enkf/model_config.cpp
lars-petter-hauge: As this is the initiation of the `ModelConfig` it _should_ free it? We are doing a deep copy of the list within `ext_job_set_define_args`, and as such the `ext_job` will not need it.

python/res/job_queue/ext_job.py
ManInFez: I would say change both :+1:

lib/job_queue/ext_job.cpp
ManInFez: Is there some configuration where set_private_args and set_defined_args are called on their own?

lib/enkf/model_config.cpp
ManInFez: The define_list is owned by the config_content object. I might be passed on to some other config object at a later stage.

lib/job_queue/ext_job.cpp
lars-petter-hauge: We do explicitly in the `forward_model_parse_job_deprecated_args` and `forward_model_parse_job_args`

python/res/job_queue/ext_job.py
lars-petter-hauge: I just wrapped it in `list` (and added a small comment as to why..)

ManInFez (approved):
Approved after black!

PR #1045, opened 2020-10-16
Update deprecated commands in GitHub actions

### LINKED ISSUE #1043: Change commands "add-path" and "set-env" in Github Actions to using Environment Files
The add-path and set-env commands are deprecated and will be disabled soon. We need to upgrade to using Environment Files. More info can be found here: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

kvashchuka:
**Issue**
Resolves #1043 

**Approach**
Use GitHiub environment variables instead of soon-to-be-deprecated `set-env` and system path instead of soon-to-be-deprecated `add-path`

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -34,10 +34,10 @@ jobs:
         sudo apt-get install libopm-simulators liblapack-dev valgrind
     - name: Setup environment
       run: |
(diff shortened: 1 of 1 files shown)


sondreso (approved):
LGTM! Nice initiative!

PR #1049, opened 2020-10-22
Rms env

### LINKED ISSUE #1047: RMS job exposing configurable default extensions to PATH and PYTHONPATH
**Introduction**
RMS is running in a separate environment with it's own Python interpreter. However, as RMS also serves the role of a workflow manager within our modeling pipeline, the users needs to be able to extend the RMS environment. In particular, we need to extend both `PATH` and `PYTHONPATH`. The need for passing a different `PATH` occurred quite recently due to RMS picking up an external Python interpreter if launched from within a virtual environment.

**Related work**
The recent work to facilitate `PATH` extensions can be found here. These are left here more as a reference, then necessarily recommended reading before the task is initiated.

- https://github.com/equinor/libres/pull/1046
- https://github.com/equinor/libres/pull/1041
- https://github.com/equinor/equilibrium/pull/42
- https://github.com/equinor/equilibrium/pull/39
- https://github.com/equinor/subscript/pull/217

**Task**
As of now the user will have to manually extend the RMS environment in their ERT configuration to be able to run RMS with what we in Equinor considers a default RMS environment. In particular, they will have to configure `RMS_PYTHONPATH` and `RMS_PATH_PREFIX` in their setup. The suggestion is that if a user only needs the "vanilla" environment, they will not have to do any configuration.

**Suggested approach**
I suggest that we extend the RMS job configuration file such that one for versions of RMS can specify a `RMS_PYTHONPATH` and `RMS_PATH_PREFIX` that is to be set as default. Something along the lines of:

```yaml
executable: /path/to/rms
threads: 1
env:
  10.0.0:
      RMS_PYTHONPATH: some/path
      RMS_PATH_PREFIX: some/other/path
   11.0.0:
      RMS_PYTHONPATH: some/path-11
      RMS_PATH_PREFIX: some/other/path-11
```

Then the RMS job bundles the potentially configuration from the user with the configuration and passes this on to the wrapper script. In particular, I would like to see for both `RMS_PYTHONPATH` and `RMS_PATH_PREFIX` the following priority: `paths/from/ERT-config:paths/from/configuration`.

When doing this it might be more sane to defer the transfer from `RMS_PYTHONPATH` to `PYTHONPATH` to the wrapper script and let it take responsibility of dealing with `RMS_PYTHONPATH` and `PREFIX_PATH`. Notice that this will also induce changes in the script in subscript, but I don't think it is very sane to let `RMS_PYTHONPATH` become `PYTHONPATH` before we actually are about to enter RMS.

**Notes**
- This also induces that the configuration of RMS in ert-configurations must be updated.
- The environment is deduced in the rms job in subscript, so we can at least use that as a reference for which paths needs to be included in equilibrium.

ManInFez:
**Issue**
Resolves #1047 

Using the rms_config.yml  to specify default environment for various rms versions


ManInFez:
test this please!

python/res/fm/rms/rms_run.py
```diff
@@ -99,10 +101,23 @@ def run(self):
...
+        config_env = self.config.env(self.version)
+        if not config_env and not self.allow_no_env:
+            raise RMSRunException(
+                f"RMS environment not specified for version: {self.version}"
+            )
+        exec_env_file = "%s_exec_env.json" % self_exe
+        user_env = {}
         if os.path.isfile(exec_env_file):
-            RMSRun.set_up_env(exec_env, json.load(open(exec_env_file)))
+            user_env = json.load(open(exec_env_file))
```
markusdregi, line 114: May I have a context manager?

markusdregi:
```suggestion
            with open(exec_env_file) as f:
                user_env = json.load(f)
```

python/res/fm/rms/rms_run.py
```diff
@@ -99,10 +101,23 @@ def run(self):
...
+        for var in set(config_env.keys()) | set(user_env.keys()):
+            exec_env[var] = ":".join(
+                filter(None, [user_env.get(var), config_env.get(var)])
+            )
+            if not exec_env[var].strip():
```
markusdregi, line 119:
```suggestion
            # Drop if no content
            if exec_env[var].strip() in ("", ":"):
```

share/ert/forward-models/res/script/rms
```diff
@@ -52,6 +52,11 @@ def _build_argument_parser():
         default=None,
         help='the prefix of all relative paths when rms is exporting',
     )
+    parser.add_argument(
```
markusdregi, line 55: Looking good :+1: Will you add the option to the RMS job description file as well?

markusdregi: Something along the lines of:

```
ARGLIST <IENS>  .... <RMS_OPTS>
```

for which you can then pass `<RMS_OPTS>=-a` to have it skip env validation.

markusdregi (commented):
Besides the two small comments this looks good :clap:

share/ert/forward-models/res/script/rms
```diff
@@ -52,6 +52,11 @@ def _build_argument_parser():
...
+        '-a', '--allow-no-env',
+        action="store_true",
+        help='Allow rms to run with out environment specified in rms configuration',
```
markusdregi: Allow RMS to run without a site configured environment
```suggestion
        help='Allow RMS to run without a site configured environment',
```

markusdregi (approved):
Well done :sunglasses:

PR #1051, opened 2020-10-29
Fix old style rms job

ManInFez:
Set default value for RMS_OPTS to empty


markusdregi:
May I have a test 😇

markusdregi (approved):
Thanks for the test 👍

ManInFez:
Thanks for the review :)

PR #1063, opened 2020-11-10
Catch version import error

oyvindeide:


python/res/__init__.py
```diff
@@ -27,7 +27,10 @@
 
 from cwrap import Prototype
 
-from ._version import version as __version__
+try:
+    from ._version import version as __version__
+except ImportError:
+    pass
```
sondreso, line 33: Is there a reason for not setting it to a dummy value? :slightly_smiling_face: Example from ERT:  https://github.com/equinor/ert/blob/35395372e6cbde1df59020d51bab4752d1952810/ert_shared/__init__.py#L6

oyvindeide: Because this should only be for cases where the package is not installed, and I think it is more correct with not set than a dummy value, also what `setuptools_scm` suggests: https://pypi.org/project/setuptools-scm/

sondreso (approved):
LGTM! Thanks!

PR #1069, opened 2020-11-17
Row scale

joakim-hove:
**Issue**
Part of: #1070

**Approach**
This PR is an initial refactoring to start support the localization described in #1070. The two first commits are 100% pure shuffling of code.

### DIFF
--- a/lib/enkf/local_dataset.cpp
+++ b/lib/enkf/local_dataset.cpp
@@ -19,19 +19,24 @@
 
 #include <stdlib.h>
 
+#include <unordered_map>
+
 #include <ert/util/util.h>
 #include <ert/util/hash.h>
 
 #include <ert/enkf/enkf_macros.hpp>
 #include <ert/enkf/local_ministep.hpp>
-
+#include <ert/enkf/row_scaling.hpp>
 
 #define LOCAL_DATASET_TYPE_ID  6615409
+
+
 
 struct local_dataset_struct {
   UTIL_TYPE_ID_DECLARATION;
   char      * name;
-  hash_type * nodes;            /* A hash table indexed by node keys - each element is an active_list instance. */
+  hash_type * active_size;      /* A hash table indexed by node keys - each element is an active_list instance. */
+  std::unordered_map<std::string, row_scaling> scaling;
 };
 
 
@@ -41,35 +46,39 @@ UTIL_IS_INSTANCE_FUNCTION(local_dataset , LOCAL_DATASET_TYPE_ID)
 
 
 local_dataset_type * local_dataset_alloc( const char * name ) {
-  local_dataset_type * dataset = (local_dataset_type *)util_malloc( sizeof * dataset);
+  local_dataset_type * dataset = new local_dataset_type();
 
   UTIL_TYPE_ID_INIT( dataset , LOCAL_DATASET_TYPE_ID );
-  dataset->nodes = hash_alloc();
+  dataset->active_size = hash_alloc();
   dataset->name  = util_alloc_string_copy( name );
 
   return dataset;
 }
 
+void local_dataset_free( local_dataset_type * dataset ) {
+  free(dataset->name);
+  hash_free( dataset->active_size );
+  delete dataset;
+}
+
 local_dataset_type * local_dataset_alloc_copy( local_dataset_type * src_dataset , const char * copy_name ) {
   local_dataset_type * copy_dataset = local_dataset_alloc( copy_name );
-  hash_iter_type * node_iter = hash_iter_alloc( src_dataset->nodes );
 
-  while (!hash_iter_is_complete( node_iter )) {
-    const char * key = hash_iter_get_next_key( node_iter );
-    active_list_type * active_list = active_list_alloc_copy( (const active_list_type *) hash_get( src_dataset->nodes , key ) );
-    hash_insert_hash_owned_ref( copy_dataset->nodes , key , active_list , active_list_free__);
+  hash_iter_type * size_iter = hash_iter_alloc( src_dataset->active_size );
+  while (!hash_iter_is_complete( size_iter )) {
+    const char * key = hash_iter_get_next_key( size_iter );
+    active_list_type * active_list = active_list_alloc_copy( (const active_list_type *) hash_get( src_dataset->active_size , key ) );
+    hash_insert_hash_owned_ref( copy_dataset->active_size , key , active_list , active_list_free__);
   }
+  hash_iter_free( size_iter );
 
-  hash_iter_free( node_iter );
+  for (const auto& scaling_pair : src_dataset->scaling)
+    copy_dataset->scaling.insert( scaling_pair );
+
   return copy_dataset;
 }
 
 
-void local_dataset_free( local_dataset_type * dataset ) {
-  free(dataset->name);
-  hash_free( dataset->nodes );
-  free( dataset );
-}
 
 void local_dataset_free__( void * arg ) {
   local_dataset_type * local_dataset = local_dataset_safe_cast( arg );
@@ -83,43 +92,68 @@ const char * local_dataset_get_name( const local_dataset_type * dataset) {
 
 
 void local_dataset_add_node(local_dataset_type * dataset, const char *node_key) {
-  if (hash_has_key( dataset->nodes , node_key ))
+  if (hash_has_key( dataset->active_size , node_key ))
     util_abort("%s: tried to add existing node key:%s \n",__func__ , node_key);
 
-  hash_insert_hash_owned_ref( dataset->nodes , node_key , active_list_alloc( ) , active_list_free__);
+  hash_insert_hash_owned_ref( dataset->active_size , node_key , active_list_alloc( ) , active_list_free__);
 }
 
 bool local_dataset_has_key(const local_dataset_type * dataset, const char * key) {
-  return hash_has_key( dataset->nodes , key );
+  return hash_has_key( dataset->active_size , key );
 }
 
 
 void local_dataset_del_node( local_dataset_type * dataset , const char * node_key) {
-  hash_del( dataset->nodes , node_key );
+  hash_del( dataset->active_size , node_key );
 }
 
 
 void local_dataset_clear( local_dataset_type * dataset) {
-  hash_clear( dataset->nodes );
+  hash_clear( dataset->active_size );
+}
+
+row_scaling * local_dataset_get_row_scaling(local_dataset_type * dataset, const char * key) {
+  auto scaling_iter = dataset->scaling.find( key );
+  if (scaling_iter != dataset->scaling.end())
+    return &scaling_iter->second;
+
+  return nullptr;
+}
+
+
+bool local_dataset_has_row_scaling(const local_dataset_type * dataset, const char * key) {
+  return (dataset->scaling.count(key) != 0);
+}
+
+
+row_scaling_type * local_dataset_get_or_create_row_scaling(local_dataset_type * dataset, const char * key) {
+  auto scaling_iter = dataset->scaling.find( key );
+  if (scaling_iter == dataset->scaling.end()) {
+    if (!hash_has_key(dataset->active_size, key))
+      throw std::invalid_argument("Tried to create row_scaling object for unknown key");
+
+    dataset->scaling.emplace( key, row_scaling{});
+  }
+  return local_dataset_get_row_scaling( dataset, key );
 }
 
 
 active_list_type * local_dataset_get_node_active_list(const local_dataset_type * dataset , const char * node_key ) {
-  return (active_list_type *) hash_get( dataset->nodes , node_key );  /* Fails hard if you do not have the key ... */
+  return (active_list_type *) hash_get( dataset->active_size , node_key );  /* Fails hard if you do not have the key ... */
 }
 
 stringlist_type * local_dataset_alloc_keys( const local_dataset_type * dataset ) {
-  return hash_alloc_stringlist( dataset->nodes );
+  return hash_alloc_stringlist( dataset->active_size );
 }
 
 void local_dataset_summary_fprintf( const local_dataset_type * dataset , FILE * stream) {
 {
-  hash_iter_type * data_iter = hash_iter_alloc( dataset->nodes );
+  hash_iter_type * data_iter = hash_iter_alloc( dataset->active_size );
   while (!hash_iter_is_complete( data_iter )) {
     const char * data_key          = hash_iter_get_next_key( data_iter );
     fprintf(stream , "NAME OF DATA:%s,", data_key );
 
-    active_list_type * active_list = (active_list_type *)hash_get( dataset->nodes , data_key );
+    active_list_type * active_list = (active_list_type *)hash_get( dataset->active_size , data_key );
     active_list_summary_fprintf( active_list , local_dataset_get_name(dataset) , data_key , stream);
   }
   hash_iter_free( data_iter );
@@ -128,36 +162,18 @@ void local_dataset_summary_fprintf( const local_dataset_type * dataset , FILE * 
 
 
 int local_dataset_get_size( const local_dataset_type * dataset ) {
-  return hash_get_size( dataset->nodes );
+  return hash_get_size( dataset->active_size );
 }
 
 
 hash_iter_type * local_dataset_alloc_iter(const local_dataset_type * dataset) {
-  return hash_iter_alloc(dataset->nodes);
-}
-
-bool local_dataset_has_row_scaling(const local_dataset_type * dataset, const char * key) {
-  return false;
-}
-
-std::vector<std::string> local_dataset_scaled_keys(const local_dataset_type * dataset) {
-  std::vector<std::string> keys;
-  hash_iter_type *node_iter = hash_iter_alloc( dataset->nodes );
-
-  while (!hash_iter_is_complete( node_iter )) {
-    const char * key = hash_iter_get_next_key( node_iter );
-    if (local_dataset_has_row_scaling(dataset, key))
-      keys.emplace_back( key );
-  }
-
-  hash_iter_free( node_iter );
-  return keys;
+  return hash_iter_alloc(dataset->active_size);
 }
 
 
 std::vector<std::string> local_dataset_unscaled_keys(const local_dataset_type * dataset) {
   std::vector<std::string> keys;
-  hash_iter_type *node_iter = hash_iter_alloc( dataset->nodes );
+  hash_iter_type *node_iter = hash_iter_alloc( dataset->active_size );
 
   while (!hash_iter_is_complete( node_iter )) {
     const char * key = hash_iter_get_next_key( node_iter );
@@ -166,5 +182,15 @@ std::vector<std::string> local_dataset_unscaled_keys(const local_dataset_type * 
   }
 
   hash_iter_free( node_iter );
+  return keys;
+}
+
+
+std::vector<std::string> local_dataset_scaled_keys(const local_dataset_type * dataset) {
+  std::vector<std::string> keys;
+  for (const auto& [key, _] : dataset->scaling) {
+    (void)_;
+    keys.push_back(key);
+  }
   return keys;
 }
--- a/lib/enkf/row_scaling.cpp
+++ b/lib/enkf/row_scaling.cpp
@@ -21,12 +21,195 @@
 #include <stdexcept>
 #include <vector>
 
-
 #include <ert/enkf/row_scaling.hpp>
+#include <ert/res_util/matrix.hpp>
+#include <ert/util/util.hpp>
+
+#define ROW_SCALING_ID 6123199
+#define ROW_SCALING_RESOLUTION 1000
 
 
+/*
+  The values in the row_scaling container are distributed among
+  ROW_SCALING_RESOLUTION discreete values. The reason for this lumping is to be
+  able to reuse the scaled update matrix:
 
-struct row_scaling_struct {
-};
+      X(a) = a*X0 + (1 - a)*I
+
+  for several rows. This grouping is utilized in the row_scaling_multiply()
+  function.
+*/
 
 
+namespace {
+
+  void scaleX(matrix_type* X, const matrix_type * X0, double alpha) {
+    matrix_assign(X, X0);
+    matrix_scale(X, alpha);
+    for (int i=0; i < matrix_get_rows(X); i++)
+      matrix_iset(X, i, i, (1 - alpha) + matrix_iget(X, i, i));
+  }
+
+}
+
+int row_scaling::size() const {
+  return this->data.size();
+}
+
+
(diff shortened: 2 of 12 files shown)


ertomatic:

pinkwah:
Is it defensible to add new C code? We're trying to move away from the legacy code and use Python as much as possible.

In new code, we should:
1. Avoid `util_abort`. Use `std::optional` or error codes. Errors must be recoverable.
2. Avoid C getters. They account for a lot of headache when reading code. Prefer public structs. (This is a recent change of direction and can be discussed how to implement in practice)
3. Have consistent code style. The spacing between functions is inconsistent as well as the space between parens in function calls. You have both `foo( 1, 2 )` and `foo(1, 2)`. There's no standard at this time but at least choose one or the other for this PR.
4. Don't put double-underscore functions in public headers. These functions aren't public and shouldn't be available publicly.
5. Use `black` for new Python code
6. Use `pytest` for new or refactored Python unit tests

tests/res/enkf/test_row_scaling.py
```diff
@@ -0,0 +1,33 @@
...
+#
+#  ERT is distributed in the hope that it will be useful, but WITHOUT ANY
+#  WARRANTY; without even the implied warranty of MERCHANTABILITY or
+#  FITNESS FOR A PARTICULAR PURPOSE.
+#
+#  See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
+#  for more details.
+
+from res.enkf import RowScaling
+from tests import ResTest
```
pinkwah: Rewrite as

```python
import pytest
from res.enkf import RowScaling


def test_access():
    row_scaling = RowScaling()
    assert len(row_scaling) == 0

    row_scaling[9] = 0.25
    assert len(row_scaling) == 10
    assert row_scaling[0] == 1
    assert row_scaling[9] == 0.25

    with pytest.raises(IndexError):
        var = row_scaling[10]
```

python/res/enkf/local_dataset.py
```diff
@@ -39,6 +42,12 @@ def __delitem__(self, key):
         else:
             raise KeyError('Unknown key "%s"' % key)
 
+    def rowScaling(self, key):
```
pinkwah: PEP8: `row_scaling`

python/res/enkf/local_dataset.py
```diff
@@ -39,6 +42,12 @@ def __delitem__(self, key):
...
+        if not key in self:
```
pinkwah: `key not in self` rather than `not key in self`

python/res/enkf/local_dataset.py
```diff
@@ -39,6 +42,12 @@ def __delitem__(self, key):
...
+            raise KeyError("Unknown key: {}".format(key))
```
pinkwah: Python 3 formatting: `f"Unknown key: {key}"`

lib/include/ert/enkf/row_scaling.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <ert/util/type_macros.h>
+#include <ert/res_util/matrix.hpp>
+
+typedef struct row_scaling_struct row_scaling_type;
+
+row_scaling_type * row_scaling_alloc();
+row_scaling_type * row_scaling_alloc_copy(const row_scaling_type * row_scaling);
+void               row_scaling_free(row_scaling_type * row_scaling);
+void               row_scaling_free__(void * arg);
```
pinkwah: Private function in public interface.

Also is it used anywhere?

lib/enkf/local_dataset.cpp
```diff
@@ -24,14 +24,15 @@
...
-
+#include <ert/enkf/row_scaling.hpp>
 
 #define LOCAL_DATASET_TYPE_ID  6615409
 
 struct local_dataset_struct {
   UTIL_TYPE_ID_DECLARATION;
   char      * name;
-  hash_type * nodes;            /* A hash table indexed by node keys - each element is an active_list instance. */
+  hash_type * active_size;      /* A hash table indexed by node keys - each element is an active_list instance. */
```
pinkwah, line 38: Is this better than `std::unordered_map`?

lib/enkf/row_scaling.cpp
```diff
@@ -0,0 +1,208 @@
...
+row_scaling_type * row_scaling_alloc_copy(const row_scaling_type * row_scaling) {
+  row_scaling_type * copy = row_scaling_alloc();
+  copy->data = row_scaling->data;
+  return copy;
+}
+
+
+
+double row_scaling_iget(const row_scaling_type * row_scaling, int index) {
+  if (index < 0)
```
pinkwah: Simplify function: `return row_scaling->data.at(index);` (it throws `std::out_of_range` if out-of-range)

lib/enkf/row_scaling.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+      X(a) = a*X0 + (1 - a)*I
+
+  for several rows. This grouping is utilized in the row_scaling_multiply()
+  function.
+*/
+
+
+struct row_scaling_struct {
+  UTIL_TYPE_ID_DECLARATION;
```
pinkwah: Is it necessary for this object to be safe-casted from `void*`? It seems like unnecessary complexity which will make it difficult to understand how it's used in the future.

lib/include/ert/enkf/row_scaling.hpp
```diff
@@ -0,0 +1,49 @@
...
+void               row_scaling_multiply(const row_scaling_type * row_scaling, matrix_type * A, const matrix_type * X0);
+void               row_scaling_multiply2(const row_scaling_type * row_scaling, matrix_type * A, const matrix_type * X0);
```
pinkwah: How does `multiply2` different from `multiply`? Add a comment.

lib/enkf/row_scaling.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+
+void row_scaling_iset(row_scaling_type * row_scaling, int index, double value) {
+  if (index < 0)
+    throw std::invalid_argument("Invalid index");
+
+  if (index >= row_scaling->data.size())
+    row_scaling->data.resize(index + 1, 1);
+
+  if (value < 0 || value > 1)
```
pinkwah, line 65: Do this before you resize so that catching this exception doesn't invalidate the `row_scaling` object.

lib/enkf/row_scaling.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+void row_scaling_free__(void * arg) {
+  row_scaling_type * row_scaling = row_scaling_safe_cast( arg );
+  row_scaling_free( row_scaling );
+}
+
+
+void row_scaling_multiply2(const row_scaling_type * row_scaling, matrix_type * A, const matrix_type * X0) {
+  if (row_scaling->data.size() != matrix_get_rows(A))
+    util_abort("%s: size mismatch between row_scaling and A matrix", __func__);
```
pinkwah: `std::argument_error`

lib/enkf/row_scaling.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+  if (matrix_get_columns(A) != matrix_get_rows(X0))
+    util_abort("%s: size mismatch between X0 and A matrix", __func__);
+
+  if (matrix_get_columns(X0) != matrix_get_rows(X0))
+    util_abort("%s: X0 matrix not rectangular", __func__);
```
pinkwah: Did you mean square or did you mean to have a `==`?

Also, `throw std::argument_error`

joakim-hove:
> Is it defensible to add new C code? We're trying to move away from the legacy code and use Python as much as possible.

I see that point. I will not argue the case either way - but for the task I have been assigned I really only see as a viable short term solution to proceed with C; if you prefer to let refactoring to Python proceed further before this problem is addressed I can back off for now.

> Avoid util_abort. Use std::optional or error codes. Errors must be recoverable.

I have used `std::exception` instead of `util_abort()`. I can look into using `std::optional<>` instead - but that certainly shifts the responsability for error checking. I don't think there is error checking code which can handle error returns from the very inner loops I am targeting here - but I will certainly keep it in mind.

> Avoid C getters. They account for a lot of headache when reading code. Prefer public structs.

I will do as told - but this is certainly a big change; for good or for bad most of libres is built with private access the way I have started on also with the new `row_scaling_type`.

> Have consistent code style. 

Sure - I will try harder!

> Don't put double-underscore functions in public headers. These functions aren't public and shouldn't be available publicly.

Well - there are patterns and there are patterns; one of the patterns in `libecl` and `libres` is that functions which take a `(void * arg)` argument and subsequently cast and call the real function have a trailing  `__` - i.e. like this:

```C++

void xxx_type_someting( xxx_type * arg ) {
    ....
}

void xxx_type_something__(void * arg) {
       xxx_type * xxx = xxx_type_safe_cast( arg );
       xxx_type_something( xxx );
}
```
The `xxx_type_something__()` function is to typically used to register a suitable `free( )` function in the `vector` and `hash` containers. I am not arguing the virtues of this particular pattern, but right now the quite some `__` functions need to be public to stick with that pattern.

> Use black [and Pytest] for new Python code 

Sure!


I will probably not make any code updates before January - but in general I will eventually try my best to address your comments.

lib/include/ert/enkf/row_scaling.hpp
joakim-hove: Yes it is used - when a `row_scaling` instance is registered in the `local_dataset::` - might still be on my machine though; this is certainly WIP.

python/res/enkf/local_dataset.py
joakim-hove: OK

python/res/enkf/local_dataset.py
joakim-hove: OK

python/res/enkf/local_dataset.py
joakim-hove: Class has `camelCase` right now - but sure I'll do snake.

tests/res/enkf/test_row_scaling.py
joakim-hove: Thanks for the update - I'll copy & paste it in :grin:

lib/include/ert/enkf/row_scaling.hpp
joakim-hove: WIP - one should be removed; `multiply` is more complex and faster code - debating whether to go for simpler code or faster code.

lib/enkf/row_scaling.cpp
joakim-hove: Sure

lib/enkf/row_scaling.cpp
joakim-hove: I want a square matrix.

lib/enkf/row_scaling.cpp
joakim-hove: ok

lib/enkf/local_dataset.cpp
joakim-hove: `std::unordered_map` is probably better; a it more hassle to wrap the pointer in something which will destruct it.

pinkwah:
> I see that point. I will not argue the case either way - but for the task I have been assigned I really only see as a viable short term solution to proceed with C; if you prefer to let refactoring to Python proceed further before this problem is addressed I can back off for now.

Yep, I agree. :+1: 

> I have used `std::exception` instead of `util_abort()`. I can look into using `std::optional<>` instead - but that certainly shifts the responsability for error checking. I don't think there is error checking code which can handle error returns from the very inner loops I am targeting here - but I will certainly keep it in mind.

Imo, even just returning a `bool` is fine. The `util_abort`s are caused programmer errors so it's debatable whether it makes sense to return anything. I think a `std::logic_error` or `std::argument_error` as you've used other places is fine.

> 
> > Avoid C getters. They account for a lot of headache when reading code. Prefer public structs.
> 
> I will do as told - but this is certainly a big change; for good or for bad most of libres is built with private access the way I have started on also with the new `row_scaling_type`.

This one's a bit tricky and I'll need to investigate how to properly interface the C code with the Python code without getters. I'll come up with an alternative.

> > Don't put double-underscore functions in public headers. These functions aren't public and shouldn't be available publicly.
> 
> Well - there are patterns and there are patterns; one of the patterns in `libecl` and `libres` is that functions which take a `(void * arg)` argument and subsequently cast and call the real function have a trailing `__` - i.e. like this:

Oh okay. I read `__` as private, but I guess it's fine as long as there's a use for this function.

> I will probably not make any code updates before January - but in general I will eventually try my best to address your comments.

Thanks so much for you effort!

lib/enkf/row_scaling.cpp
joakim-hove: Sure

lib/enkf/row_scaling.cpp
joakim-hove: Well - the `local_dataset_type` has `hash_type` container of these objects, they need to be castable from `void *` to register a destructor in the `hash` container, and then this "type system" ensures some safety in the casting. Again - I will not argue the virtues of this system, but it has been quite a lot used in `libecl / libres`. The `xxx_safe_cast()` functions seldom trigger - but when they do they give the developer a quite good direction of where the bug is (of course - while casting from `void *` the alternative is silent "success").

lib/include/ert/enkf/row_scaling.hpp
pinkwah: I'd vote for faster. I think comments to explain what's going if it's too complex. But up to you really.

lib/enkf/local_dataset.cpp
pinkwah: It's kind of ugly, but:
```c++
struct active_list_deleter {
    void operator()(active_list *p) const {
        active_list_free(p);
    }
};

/* Not sure about the naming convention */
using active_list_uq = std::unique_ptr<active_list, active_list_deleter>;

struct local_dataset_struct {
    /* [..] */
    std::unordered_map<std::string, active_list_uq> active_size;
}
```

It's easier if the objects are C++ classes that are wrapped by some C classes, however.

lib/enkf/row_scaling.cpp
pinkwah: I will not push hard against this, but I think it's nice to consider alternatives so that we don't add more technical debt. I can try to solve it if you don't have time.

lib/enkf/local_dataset.cpp
joakim-hove: Yes - I know; that is the alternative - in my opinion this is a bit halfways; in my opinion the alternatives are:

1. Full C style approach - for all it is worth that pattern is used extensively in the code.
2. Full C++ style with `std::unordered_map<ActiveList>` and a full blown `class ActiveList` with value semantics.

I'll do as told though :-)

joakim-hove:
Thank you for the review comments - I will address them although it might take some time. I would like to "make a statement" to the general question of C vs C++, homebrew containers vs stl and so on:

1. I have aimed to write code in the existing C style with homemade containers and so on[1]  - not in any way because I consider it superior, but because that is the style of the existing code.
2. If you have wishes in this regard - e.g. use stl for christ sake or something similar I am all ears, but I am a bit hesitant to devise my own solutions in this regard as I consider it highly improbable that I will hit your sweet spot.

[1]: At least in the API.

joakim-hove:
OK - I I now consider this ready for review. Qualitative remarks:

1. I have implemented the new class `row_scaling` as a C++ class and added a C layer for Python interop. For this particular case this was quite simple.
2. I have tried to address all of @dotfloat's comments - with the exception of public struct - which is maybe not particularlt relevant any longer.

When this is merged the next step is to actually start using this code, that will probably induce additional changes.

pinkwah:
Looks good from my side in terms of the C++ stuff. :+1:

ManInFez:
Test this please!

lib/enkf/row_scaling.cpp
```diff
@@ -21,12 +21,152 @@
...
+}
+
+void row_scaling::assign(int index, double value) {
+  if (value < 0 || value > 1)
+    throw std::invalid_argument("Invalid value ");
+
+  if (this->data.size() <= index)
+    this->data.resize(index + 1, 1);
+
+  this->data.at(index) = floor(value * this->resolution) / this->resolution;
```
ManInFez, line 71: Why is this rounded down to four decimal precision?

joakim-hove: This rounding is done to ensure that only a finite number of distinct values are present in the row_scaling vector; then when the multiplication takes the vector is grouped in chunks with equal value. All of these hoops are to speed up the calculation. If you look at the test there is naive multiply function which does not do these hoops - that is slower - and simpler to read.

lib/enkf/row_scaling.cpp
```diff
@@ -21,12 +21,152 @@
...
+        for (int i=0; i < matrix_get_columns(A); i++)
+          sum += src_row[i] * X_data[GET_INDEX(X, i, j)];
+
+        target_row[j] = sum;
+      }
+      matrix_set_row(A, target_row.data(), row);
+    }
+
+    index_offset = end_index;
+    if (index_offset == this->data.size())
```
ManInFez: this could go in the while statement at the beginning in line 90

lib/enkf/row_scaling.cpp
```diff
@@ -21,12 +21,152 @@
...
+      std::vector<double> src_row(matrix_get_columns(A));
+      std::vector<double> target_row(matrix_get_columns(A));
+
+      for (int j=0; j < matrix_get_columns(A); j++)
+        src_row[j] = matrix_iget(A, row, j);
+
+      for (int j=0; j < matrix_get_columns(A); j++) {
+        double sum = 0;
+        for (int i=0; i < matrix_get_columns(A); i++)
+          sum += src_row[i] * X_data[GET_INDEX(X, i, j)];
```
ManInFez: Why are we using the one-dimensional representation of X

lib/enkf/tests/row_scaling.cpp
```diff
@@ -15,10 +15,152 @@
...
+  }
+  matrix_free(X);
+}
+
+void test_multiply(const row_scaling_type * row_scaling, const matrix_type * A0, const matrix_type * X0) {
+  matrix_type * A1 = matrix_alloc_copy(A0);
+  matrix_type * A2 = matrix_alloc_copy(A0);
+
+  row_scaling_multiply(row_scaling, A1, X0);
+  row_scaling_multiply(row_scaling, A2, X0);
```
ManInFez: I assume one of these should be row_scaling_multiply2?

lib/enkf/row_scaling.cpp
ManInFez: 👍

lib/enkf/row_scaling.cpp
```diff
@@ -21,12 +21,152 @@
...
+  if (matrix_get_columns(X0) != matrix_get_rows(X0))
+    throw std::invalid_argument("X0 matrix is not quadratic");
+
+  matrix_type * X = matrix_alloc(matrix_get_rows(X0), matrix_get_columns(X0));
+  std::vector<int> sort_index(this->size());
+  std::iota(sort_index.begin(), sort_index.end(), 0);
+  std::sort(sort_index.begin(), sort_index.end(), [this] (int index1, int index2) { return this->operator[](index1) > this->operator[](index2); });
+
+  std::size_t index_offset = 0;
+  while (true) {
```
autt, line 130: Can this loop and the grouping of alphas be commented so it is easier to follow?

lib/enkf/row_scaling.cpp
```diff
@@ -21,12 +21,152 @@
...
+double row_scaling::operator[](int index) const {
+  return this->data.at(index);
+}
+
+void row_scaling::assign(int index, double value) {
+  if (value < 0 || value > 1)
+    throw std::invalid_argument("Invalid value ");
+
+  if (this->data.size() <= index)
+    this->data.resize(index + 1, 1);
```
ManInFez, line 69: This automatic resizing is a bit questionable.  While it is convenient, a more explicit usage could prevent hard to discover side effects

lib/enkf/row_scaling.cpp
```diff
@@ -21,12 +21,152 @@
...
+      X(a) = a*X0 + (1 - a)*I
+
+  for several rows. This grouping is utilized in the row_scaling_multiply()
+  function.
+*/
+
+
+namespace {
+
+void scaleX(matrix_type* X, const matrix_type * X0, double alpha) {
```
ManInFez: indentation missing for scaleX

lib/enkf/local_dataset.cpp
```diff
@@ -41,35 +46,40 @@ UTIL_IS_INSTANCE_FUNCTION(local_dataset , LOCAL_DATASET_TYPE_ID)
...
+
 local_dataset_type * local_dataset_alloc_copy( local_dataset_type * src_dataset , const char * copy_name ) {
   local_dataset_type * copy_dataset = local_dataset_alloc( copy_name );
-  hash_iter_type * node_iter = hash_iter_alloc( src_dataset->nodes );
-
-  while (!hash_iter_is_complete( node_iter )) {
-    const char * key = hash_iter_get_next_key( node_iter );
-    active_list_type * active_list = active_list_alloc_copy( (const active_list_type *) hash_get( src_dataset->nodes , key ) );
-    hash_insert_hash_owned_ref( copy_dataset->nodes , key , active_list , active_list_free__);
+  {
```
ManInFez: unnecessary block

ManInFez:
Overall looks good. Some minor comments. Please tag me and @autt here if there is any questions/comments :)

lib/enkf/row_scaling.cpp
joakim-hove: For efficiency; I can change that.

lib/enkf/tests/row_scaling.cpp
joakim-hove: Ahhhhh  ;-)

lib/enkf/row_scaling.cpp
joakim-hove: Sure

lib/enkf/row_scaling.cpp
joakim-hove: OK

lib/enkf/local_dataset.cpp
joakim-hove: OK

lib/enkf/row_scaling.cpp
ManInFez: no need... was just curious

joakim-hove:
@ManInFez & @autt : I have addressed all the comments except the one about automatic resizing; changes are force pushed.

ManInFez:
Test this please!

ManInFez (approved):

ManInFez:
@joakim-hove Very nice!

PR #1071, opened 2020-11-17
Remove check which dates back to EnKF updates - i.e. many years

joakim-hove:
**Issue**
Small contribution to #1070 

**Approach**
The test which is removed in this PR comes from the time when ERT could do EnKF with updates of dynamic properties like pressure and saturations. That functionality has not been active for many years and the if test can safely be removed to reduce complexity one tiny bit.


ertomatic:

markusdregi:
Jenkins: Test this please

markusdregi (approved):
Looks good 👍 Thanks for the clean up!

PR #1086, opened 2020-12-22
Update main update code with mock row scaling api.

joakim-hove:
**Issue**
Part of #1070

**Approach**
This PR creates an empty `row_scaling` datastructure in order to create a non-functional api and update the inner loop updating. This should be merged before #1069 - which has an actual implementation of the row_scaling class.

The distance based localization looks like it will be based loosely on three levels of abstraction/code:

1. At the very innermost level are changes to the core updating routine `enkf_main_update()`.
2. At the the intermediate level is a datastructure which represents the resulting row scaling for one parameter; the row scaling will enter into the inner loop code in point 1. A sketch of a possible `row_scaling`implementation can be found here: #1069.
3. Expert users can use the `row_scaling`functionality from Python - but probably it will make sense to create an additional utility layer which creates/initializes `row_scaling` instances based on relevant domain models like grids and 3D fields.

This PR addresses the changes needed in step 1 above; the current PR can certainly be improved - but it should contain the required changes in functionality to address point 1 above. The current PR can be merged without causing any harm to existing functionality, no new functionality will be enabled with this PR.

### DIFF
--- a/lib/enkf/enkf_main.cpp
+++ b/lib/enkf/enkf_main.cpp
@@ -599,15 +599,15 @@ static int enkf_main_serialize_dataset( const ensemble_config_type * ens_config 
   matrix_type * A   = serialize_info->A;
   int ens_size      = matrix_get_columns( A );
   int current_row   = 0;
-  int ikw = 0;
-  hash_iter_type* data_iter = local_dataset_alloc_iter(dataset);
-  serialize_info->active_size.resize( local_dataset_get_size( dataset ));
-  serialize_info->row_offset.resize( local_dataset_get_size( dataset ));
-  while (!hash_iter_is_complete(data_iter)) {
-    const char             * key = hash_iter_get_next_key(data_iter);
-    const active_list_type * active_list = local_dataset_get_node_active_list( dataset , key );
+
+  const auto& unscaled_keys = local_dataset_unscaled_keys(dataset);
+  serialize_info->active_size.resize( unscaled_keys.size() );
+  serialize_info->row_offset.resize( unscaled_keys.size() );
+  for (int ikw = 0; ikw < unscaled_keys.size(); ikw++) {
+    const auto& key = unscaled_keys[ikw];
+    const active_list_type * active_list = local_dataset_get_node_active_list( dataset , key.c_str() );
     enkf_fs_type * src_fs = serialize_info->src_fs;
-    serialize_info->active_size[ikw] = __get_active_size( ens_config , src_fs , key , report_step , active_list );
+    serialize_info->active_size[ikw] = __get_active_size( ens_config , src_fs , key.c_str() , report_step , active_list );
     serialize_info->row_offset[ikw] = current_row;
 
     {
@@ -617,12 +617,10 @@ static int enkf_main_serialize_dataset( const ensemble_config_type * ens_config 
     }
 
     if (serialize_info->active_size[ikw] > 0) {
-      enkf_main_serialize_node( key , active_list , serialize_info->row_offset[ikw] , work_pool , serialize_info );
+      enkf_main_serialize_node( key.c_str() , active_list , serialize_info->row_offset[ikw] , work_pool , serialize_info );
       current_row += serialize_info->active_size[ikw];
     }
-    ikw++;
   }
-  hash_iter_free( data_iter );
   matrix_shrink_header( A , current_row , ens_size );
   return matrix_get_rows( A );
 }
@@ -673,12 +671,11 @@ static void enkf_main_deserialize_dataset( ensemble_config_type * ensemble_confi
                                            thread_pool_type * work_pool ) {
 
   const int num_cpu_threads = thread_pool_get_max_running( work_pool );
-  int ikw = 0;
-  hash_iter_type* data_iter = local_dataset_alloc_iter(dataset);
-  while (!hash_iter_is_complete(data_iter)) {
-    const char             * key = hash_iter_get_next_key(data_iter);
+  const auto& unscaled_keys = local_dataset_unscaled_keys(dataset);
+  for (int ikw=0; ikw < unscaled_keys.size(); ikw++) {
+    const auto& key = unscaled_keys[ikw];
     if (serialize_info->active_size[ikw] > 0) {
-      const active_list_type * active_list      = local_dataset_get_node_active_list( dataset , key );
+      const active_list_type * active_list      = local_dataset_get_node_active_list( dataset , key.c_str() );
 
       {
         /* Multithreaded */
@@ -686,7 +683,7 @@ static void enkf_main_deserialize_dataset( ensemble_config_type * ensemble_confi
         serialize_node_info_type node_info[num_cpu_threads];
         thread_pool_restart( work_pool );
         for (icpu = 0; icpu < num_cpu_threads; icpu++) {
-          node_info[icpu].key            = key;
+          node_info[icpu].key            = key.c_str();
           node_info[icpu].active_list    = active_list;
           node_info[icpu].row_offset     = serialize_info->row_offset[ikw];
           serialize_info[icpu].node_info = &node_info[icpu];
@@ -697,7 +694,6 @@ static void enkf_main_deserialize_dataset( ensemble_config_type * ensemble_confi
       }
     }
   }
-  hash_iter_free( data_iter );
 }
 
 
@@ -1107,11 +1103,23 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
       if (local_dataset_get_size( dataset )) {
         local_obsdata_type   * local_obsdata = local_ministep_get_obsdata( ministep );
 
-        // The enkf_main_serialize_dataset() function will query the storage
-        // layer and fetch data which is serialized into the A matrix which is
-        // buried deep into the serialize_info structure.
         enkf_main_serialize_dataset(enkf_main_get_ensemble_config(enkf_main), dataset , step2 ,  use_count , tp , serialize_info);
         module_info_type * module_info = enkf_main_module_info_alloc(ministep, obs_data, dataset, local_obsdata, serialize_info->active_size.data() , serialize_info->row_offset.data());
+
+        /*
+          The update for one local_dataset instance consists of two main chunks:
+
+          1. The first chunk updates all the parameters which don't have row
+             scaling attached. These parameters are serialized together to the A
+             matrix and all the parameters are updated in one go.
+
+          2. The second chunk is loop over all the parameters which have row
+             scaling attached. These parameters are updated one at a time.
+         */
+
+
+        // Part 1: Parameters which do not have row scaling attached.
+        enkf_main_serialize_dataset(enkf_main_get_ensemble_config(enkf_main), dataset , step2 ,  use_count , tp , serialize_info);
 
         if (analysis_module_check_option( module , ANALYSIS_UPDATE_A)){
           if (analysis_module_check_option( module , ANALYSIS_ITERABLE)){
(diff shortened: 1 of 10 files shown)


ertomatic:

joakim-hove:
Can this get a review/merge?

markusdregi:
Jenkins: Test this please

lib/enkf/enkf_main.cpp
```diff
@@ -599,12 +599,11 @@ static int enkf_main_serialize_dataset( const ensemble_config_type * ens_config
   matrix_type * A   = serialize_info->A;
   int ens_size      = matrix_get_columns( A );
   int current_row   = 0;
-  int ikw = 0;
-  hash_iter_type* data_iter = local_dataset_alloc_iter(dataset);
-  serialize_info->active_size.resize( local_dataset_get_size( dataset ));
-  serialize_info->row_offset.resize( local_dataset_get_size( dataset ));
-  while (!hash_iter_is_complete(data_iter)) {
-    const char             * key = hash_iter_get_next_key(data_iter);
+  stringlist_type * unscaled_keys = local_dataset_alloc_unscaled_keys(dataset);
```
ManInFez: I think we should stop using the libecl vectors in as much of an extent as possible. 
use `std::vector<std::string> unscaled_keys` instead and propagate it downwords in form of iterators
like:
`for( auto & key : unscaled_keys) { }`

lib/enkf/enkf_main.cpp
```diff
@@ -620,9 +619,8 @@ static int enkf_main_serialize_dataset( const ensemble_config_type * ens_config
       enkf_main_serialize_node( key , active_list , serialize_info->row_offset[ikw] , work_pool , serialize_info );
       current_row += serialize_info->active_size[ikw];
     }
-    ikw++;
   }
-  hash_iter_free( data_iter );
+  stringlist_free(unscaled_keys);
```
ManInFez: then this cleanup becomes unnecessary

lib/enkf/enkf_main.cpp
```diff
@@ -673,10 +671,9 @@ static void enkf_main_deserialize_dataset( ensemble_config_type * ensemble_confi
                                            thread_pool_type * work_pool ) {
 
   const int num_cpu_threads = thread_pool_get_max_running( work_pool );
-  int ikw = 0;
-  hash_iter_type* data_iter = local_dataset_alloc_iter(dataset);
-  while (!hash_iter_is_complete(data_iter)) {
-    const char             * key = hash_iter_get_next_key(data_iter);
+  stringlist_type * unscaled_keys = local_dataset_alloc_unscaled_keys(dataset);
```
ManInFez: same here

lib/enkf/enkf_main.cpp
```diff
@@ -1107,31 +1104,97 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
...
+             scaling attached. These parameters are serialized together to the A
+             matrix and all the parameters are updated in one go.
+
+          2. The second chunk is loop over all the parameters which have row
+             scaling attached. These parameters are updated one at a time.
+         */
+
+
+        // Part 1: Parameters which do not have row scaling attached.
+        {
```
ManInFez: Unless there is a particular need for blocks, i.e. reuse of variable names, I would rather than blocks, just use comments or functions to separate code, as too much indentation becomes hard to read.

lib/enkf/enkf_main.cpp
```diff
@@ -1107,31 +1104,97 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
...
         }
 
-        // The enkf_main_deserialize_dataset() function will dismantle the A
-        // matrix from the serialize_info structure and distribute that content
-        // over to enkf_node instances and eventually the storage layer.
-        enkf_main_deserialize_dataset( enkf_main_get_ensemble_config( enkf_main ) , dataset , serialize_info , tp);
 
+        // Part 2: Parameters with row scaling attached - to support distance based localization.
+        {
+          stringlist_type * scaled_keys = local_dataset_alloc_scaled_keys(dataset);
```
ManInFez: use std::vector

lib/enkf/enkf_main.cpp
```diff
@@ -1107,31 +1104,97 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
...
+          if (stringlist_get_size(scaled_keys)) {
+            throw std::logic_error("Sorry - the use of row scaling for distance based localization is not yet implemented");
```
ManInFez, line 1146: are these `throws` necessary?

lib/enkf/enkf_main.cpp
```diff
@@ -1107,31 +1104,97 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
...
+
+            if (analysis_module_check_option( module , ANALYSIS_UPDATE_A))
+              throw std::logic_error("Sorry - row scaling for distance based localization can not be combined with analysis modules which update the A matrix");
+
+            for (int ikw=0; ikw < stringlist_get_size(scaled_keys); ikw++) {
```
ManInFez: use std::vector

lib/enkf/enkf_main.cpp
```diff
@@ -1107,31 +1104,97 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
...
+                                   serialize_info->report_step,
+                                   0,
+                                   column,
+                                   active_list,
+                                   A);
+                }
+              }
+            }
+          }
+          stringlist_free( scaled_keys );
```
ManInFez: remove when using std::vector

lib/enkf/local_dataset.cpp
```diff
@@ -135,3 +135,36 @@ int local_dataset_get_size( const local_dataset_type * dataset ) {
 hash_iter_type * local_dataset_alloc_iter(const local_dataset_type * dataset) {
   return hash_iter_alloc(dataset->nodes);
 }
+
+bool local_dataset_has_row_scaling(const local_dataset_type * dataset, const char * key) {
+  return false;
+}
+
+stringlist_type * local_dataset_alloc_scaled_keys(const local_dataset_type * dataset) {
```
ManInFez: should return reference of type `std::vector<std::string> &`

lib/enkf/local_dataset.cpp
```diff
@@ -135,3 +135,36 @@ int local_dataset_get_size( const local_dataset_type * dataset ) {
...
+    if (local_dataset_has_row_scaling(dataset, key))
+      stringlist_append_copy( keys, key);
+  }
+
+  hash_iter_free( node_iter );
+  return keys;
+}
+
+
+stringlist_type * local_dataset_alloc_unscaled_keys(const local_dataset_type * dataset) {
```
ManInFez: should return reference of type `std::vector<std::string> &`

lib/enkf/local_dataset.cpp
```diff
@@ -135,3 +135,36 @@ int local_dataset_get_size( const local_dataset_type * dataset ) {
 hash_iter_type * local_dataset_alloc_iter(const local_dataset_type * dataset) {
   return hash_iter_alloc(dataset->nodes);
 }
+
+bool local_dataset_has_row_scaling(const local_dataset_type * dataset, const char * key) {
+  return false;
+}
+
+stringlist_type * local_dataset_alloc_scaled_keys(const local_dataset_type * dataset) {
+  stringlist_type * keys = stringlist_alloc_new();
```
ManInFez: std::vector<std::string> keys = new std::vector<std::string>()

lib/enkf/local_dataset.cpp
```diff
@@ -135,3 +135,36 @@ int local_dataset_get_size( const local_dataset_type * dataset ) {
...
+  hash_iter_type *node_iter = hash_iter_alloc( dataset->nodes );
+
+  while (!hash_iter_is_complete( node_iter )) {
+    const char * key = hash_iter_get_next_key( node_iter );
+    if (local_dataset_has_row_scaling(dataset, key))
+      stringlist_append_copy( keys, key);
```
ManInFez: `keys.push_back(key)`

lib/enkf/tests/enkf_local_dataset.cpp
```diff
@@ -0,0 +1,54 @@
...
+void test_create() {
+  local_dataset_type * ld = local_dataset_alloc("DATA");
+  local_dataset_add_node(ld, "PERMX");
+  local_dataset_add_node(ld, "PERMY");
+  local_dataset_add_node(ld, "PERMZ");
+  test_assert_false( local_dataset_has_row_scaling(ld, "PERMX"));
+  test_assert_false( local_dataset_has_row_scaling(ld, "PERMY"));
+  test_assert_false( local_dataset_has_row_scaling(ld, "PERMZ"));
+
+  {
```
ManInFez: not needed block

lib/enkf/tests/enkf_local_dataset.cpp
```diff
@@ -0,0 +1,54 @@
...
+    stringlist_type * unscaled_keys = local_dataset_alloc_unscaled_keys(ld);
+    test_assert_int_equal( stringlist_get_size(unscaled_keys), 3 );
+    test_assert_true( stringlist_contains(unscaled_keys, "PERMX"));
+    test_assert_true( stringlist_contains(unscaled_keys, "PERMY"));
+    test_assert_true( stringlist_contains(unscaled_keys, "PERMZ"));
+    stringlist_free(unscaled_keys);
+  }
+  {
```
ManInFez: not needed block

lib/enkf/tests/enkf_local_dataset.cpp
```diff
@@ -0,0 +1,54 @@
...
+   WARRANTY; without even the implied warranty of MERCHANTABILITY or
+   FITNESS FOR A PARTICULAR PURPOSE.
+
+   See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
+   for more details.
+*/
+
+#include <stdlib.h>
+#include <stdexcept>
+#include <ert/util/test_util.hpp>
```
ManInFez, line 24: remove whitespace

lib/include/ert/enkf/local_dataset.hpp
```diff
@@ -20,6 +20,10 @@
 #ifndef ERT_LOCAL_DATASET_H
 #define ERT_LOCAL_DATASET_H
 #include <ert/tooling.hpp>
+#include <ert/util/stringlist.hpp>
+#include <ert/util/hash.hpp>
```
ManInFez, line 27: remove whitespace

lib/include/ert/enkf/local_dataset.hpp
```diff
@@ -20,6 +20,10 @@
...
+
```
ManInFez, line 28: remove empty line

lib/include/ert/enkf/row_scaling.hpp
```diff
@@ -0,0 +1,33 @@
...
+  FITNESS FOR A PARTICULAR PURPOSE.
+
+  See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
+  for more details.
+*/
+
+#ifndef ROW_SCALING_H
+#define ROW_SCALING_H
+
+#ifdef __cplusplus
```
ManInFez: Libres does not any longer support to be compiled with C and thus this if becomes unnecessary

ManInFez:
I think the new code which you feel belongs in their own blocks, should most likely be moved to new functions to increase readability. Otherwise this looks fine, except for the old style c code, we have mentioned :)

joakim-hove:
Good - thank you for the review; I will update as suggested. 

The reason i did not use `std::vector<std::string>` from the outset was due to Python bindings - if the `std::vector<std::string>` (with value semantics) is to be exposed to Python the changes might ripple quite far, but I will start with `std::vector<std::string>` - and avoid exposure to Python - and then we'll burn that bridge when Python exposure becomes a necessity.

lib/enkf/enkf_main.cpp
joakim-hove: I have used `std::vector<std::string>`as container - but the iteration is still based on an index; because that index is used for other purposes also in the loop body. If you prefer I can change to a range based for and manually increment the index.

lib/enkf/enkf_main.cpp
joakim-hove: I have removed the blocks; my judgement is that it will be too low level to extract in a separate function - but I can do it.

lib/enkf/local_dataset.cpp
joakim-hove: Not a reference I guess :-)

lib/enkf/local_dataset.cpp
joakim-hove: I have used `emplace_back( )` to construct from `const char* `

lib/enkf/tests/enkf_local_dataset.cpp
joakim-hove: Arrrrgh - "Use blocks freely" will be on my tombstone; but I will remove it for now!

lib/include/ert/enkf/local_dataset.hpp
joakim-hove: There is a conscious grouping; where headers which in some sense belong together are grouped together. In this case the grouping is:
```
#include <standard_headers>

#include <libecl_headers>

#include <libres_headers>
```
but sure - I'll remove all blank lines if that is desired.

lib/include/ert/enkf/row_scaling.hpp
joakim-hove: Good point.

lib/enkf/enkf_main.cpp
joakim-hove: Well - I consider a kind of feature toggling; something has gone way south if this exception is ever thrown. I would prefer to let it remain as a marker? Will of course be removed when the implementation has progressed and the row scaling actually is present.

joakim-hove:
OK - I have updated as requested; with some minor exceptions:

1. Using index instead of range based for.
2. The `throw std::logic_error("You are running unimplemented code ...?")` remains.
3. The blank line separating groups of header includes have been retained at one location.

I will not argue any of these further if you still prefer them changed. I have amended and force pushed and updated branch, if you prefer I can for the future push separate fixups[1].

[1]: That said my muscle memory leans heavily on the rebase way - so I might forget.

joakim-hove:
Regarding C linkage: I have done as requested and use `std::vector<std::string>` instead of `stringlist_type` - I agree that using new C++ structures is a good thing - but this can have quite comprehensive ripple effects because the Python bindings require unmangeld symbol names - i.e. C linkage in the API. So - using C++ in the implementation is trivially a good idea, when it comes exposing C++ structures in the api some more care should probably be excercised; at least for the parts of the API which it is desired to expose to Python.

lib/enkf/enkf_main.cpp
ManInFez: Let us keep it for now, but I would prefer to have testable code coming in to libres. In this PR the tests are a little lacking, so lets aim to fix this in a future ;)

ManInFez:
Test this please!

ManInFez (approved):
🥓

PR #1096, opened 2021-01-15
Remove builtin and __future__ imports

### LINKED ISSUE #1094: Remove Python 2 compatibility code in networking module in JobRunner
We no longer support python 2, so this can be removed: 
https://github.com/equinor/libres/blob/a115ecd2c7fa883904919f9b0fe5e409382e0cc9/python/job_runner/reporting/network.py#L7-L9

pinkwah:
Both builtin and __future__ were used to import Python 3 functionality into Python 2.7. As Python 2.7 is no longer supported, remove these unnecessary imports.

Resolves: #1094

### DIFF
--- a/python/job_runner/reporting/network.py
+++ b/python/job_runner/reporting/network.py
@@ -4,9 +4,6 @@ import pwd
 import socket
 import sys
 
-# shadow map builtin in py2, making our map usage py2 compatible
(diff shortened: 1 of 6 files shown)


sondreso (approved):

sondreso:
Nice!

PR #1100, opened 2021-01-25
Remove LSB_JOBID from env when using eclrun

### LINKED ISSUE #1099: MPI does not work when using eclrun
When running MPI runs with eclrun, simulations will not start

sondreso:
**Issue**
Resolves #1099 

**Approach**
MPIrun from Intel MPI will try to connect to LSF if it detects LSB_JOBID in the environment. Unset the LSB_JOBID when using `eclrun`

### DIFF
--- a/python/res/fm/ecl/ecl_config.py
+++ b/python/res/fm/ecl/ecl_config.py
@@ -222,6 +222,12 @@ class EclrunConfig:
             env["PATH"] = eclrun_env["PATH"] + os.pathsep + env["PATH"]
             eclrun_env.pop("PATH")
 
+        for k, v in eclrun_env.copy().items():
+            if v is None:
+                if k in env:
+                    env.pop(k)
+                eclrun_env.pop(k)
+
         env.update(eclrun_env)
         return env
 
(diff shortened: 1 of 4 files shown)


python/res/fm/ecl/ecl_config.py
```diff
@@ -223,6 +223,12 @@ def _get_run_env(self, eclrun_env):
             eclrun_env.pop("PATH")
 
         env.update(eclrun_env)
+
+        # Having the LSB_JOBID set when executing MPI runs will
```
markusdregi: Another option would be to add `LSB_JOBID: ~` in the configuration and drop environment variables mapping to `None`. Then you get to move it out of the code 🤷 I'm not entirely sure which is better 🤔

sondreso: I thought about it, and now I

tests/res/fm/test_ecl_run_new.py
```diff
@@ -22,7 +22,10 @@
 import yaml
 import pytest
 import re
+import json
+from unittest import mock
 
+import res
```
ManInFez: unused import?

sondreso: good catch

ManInFez (approved):
LGTM!

PR #1105, opened 2021-02-05
Send absolute path to stderr and stdout to the ensemble evaluator

mortalisk:

### DIFF
--- a/python/job_runner/reporting/event.py
+++ b/python/job_runner/reporting/event.py
@@ -6,6 +6,7 @@ from job_runner.reporting.message import (
     Running,
     Start,
 )
+from pathlib import Path
 
 _FM_JOB_START = "com.equinor.ert.forward_model_job.start"
 _FM_JOB_RUNNING = "com.equinor.ert.forward_model_job.running"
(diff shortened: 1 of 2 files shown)


python/job_runner/reporting/event.py
```diff
@@ -87,7 +89,14 @@ def _init_handler(self, msg):
                     "datacontenttype": "application/json",
                 },
                 {
-                    "jobs": [job.job_data for job in msg.jobs],
+                    "jobs": {
+                        job.index: {
+                            **job.job_data,
+                            "stderr": os.path.abspath(job.job_data["stderr"]),
+                            "stdout": os.path.abspath(job.job_data["stdout"]),
```
jondequinor: plz use pathlib (`resolve`)

python/job_runner/reporting/event.py
```diff
@@ -87,7 +89,14 @@ def _init_handler(self, msg):
...
+                        }
+                        for job in msg.jobs
```
jondequinor: I find these multi line inline online dict expressions hard to read. Is it just me?

python/job_runner/reporting/event.py
jondequinor: Are these values always set?

python/job_runner/reporting/event.py
mortalisk: Don't know, but let's try on another size and see how it fits.

python/job_runner/reporting/event.py
mortalisk: Ok

mortalisk: I'll add a guard in case not

jondequinor (approved):
🔪🐷🔥🥓

PR #1109, opened 2021-02-08
Row scaling field

joakim-hove:
**Issue**

Significant part of #1069

**Approach**

With this PR the row scaling code is hooked up with the smoother updates. With this PR the "second level" in the row scaling implementation is complete. With this PR the row_scaling code is ready to be used, at the core of the configuration is a function call:
```python
row_scaling.assign( data_size, callable )
```
where `callable` is a user supplied callable which will be called as:
```
for index in range(data_size):
     row_scaling[index] = callable(index)
```
this is quite flexible - and can in principle achieve *anything* - but a quite significant burden is left on the user to actually implement `callable`. So an envisioned third layer is a collection of callables for typical situations. This layer should probably be designed/implemented in cooperation with Real Users&trade;


ertomatic:

ManInFez:
Test this please!

python/res/enkf/row_scaling.py
```diff
@@ -51,3 +51,51 @@ def __getitem__(self, index):
...
+
+            ens_config = ert.ensembleConfig()
+            config_node = ens_config["PORO"]
+            field_config = config.node.getFieldModelConfig()
+            grid = ert.eclConfig().getGrid()
+            pos = grid.get_xyz(ijk=(10,10,1))
+            r0 = 10
+
+            if grid.get_num_active() != field_config.get_data_size():
+                raise ValuError("Ohh fuck - something is rotten)
```
ManInFez: remove profanity

test-data/local/row_scaling/workflows/ROW_SCALING_WORKFLOW1
```diff
@@ -0,0 +1 @@
+ROW_SCALING_JOB1
```
ManInFez, line 1: missing new line

ManInFez: missing new line

python/res/enkf/row_scaling.py
ManInFez, line 84: remove profanity

python/res/enkf/row_scaling.py
```diff
@@ -51,3 +51,51 @@ def __getitem__(self, index):
...
+        to use a class instance which implements the __call__() method.
+
+        It is an important point that the assign() method does not have any
+        context for error checking, i.e. if you call it with an incorrect value
+        for the size argument things will silently pass initially but might
+        blow up in the subsequent update step.
+
+        """
+
+        for index in range(target_size):
```
ManInFez, line 104: Could it be an idea to input a target as a list instead of a value. I can imagine a situation where target is not everyone up to a specific index, but from x to y...

ManInFez:
Please rebase on master as the rng-commit is in both places.

It would be very helpful if we could set up a meeting and go through the tests and test-setup?

test-data/local/row_scaling/workflows/ROW_SCALING_WORKFLOW1
joakim-hove: OK

python/res/enkf/row_scaling.py
joakim-hove: OK

python/res/enkf/row_scaling.py
joakim-hove: I don't fully understand what you mean here. My general comment is that this is a quite low level API, and I think your suggestion will take it even lower - but I might misunderstand. Let's discuss it!

joakim-hove:
> Please rebase on master as the rng-commit is in both places.

Done

> It would be very helpful if we could set up a meeting and go through the tests and test-setup?

Sure - I am available 1300 - 1430 today; or tomorrow.

ManInFez:
Great! Me and @autt  have a meeting at 13:00 today, so tomorrow might be best. 0900?

ManInFez:
Test this please!

ManInFez (approved):

ManInFez:
Test this please!

PR #1110, opened 2021-02-08
field_config_guess_file_type: Use extension when guessing grdecl

joakim-hove:
**Issue**
This is a drive-by change while working on #1069. The .grdecl file format is by far the most important to support for the field_config class. With these changes we remove the restriction that files must be above 256 bytes to check type. 

**Approach**
For grdecl formatted files we start by checking extension, and then content. The grdecl based check is first in the list of file formats tested.

### DIFF
--- a/lib/enkf/field_config.cpp
+++ b/lib/enkf/field_config.cpp
@@ -217,7 +217,29 @@ field_file_format_type field_config_default_export_format(const char * filename)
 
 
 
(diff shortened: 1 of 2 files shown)


ertomatic:

ManInFez:
Test this please!

ManInFez (approved):
👍

PR #1114, opened 2021-02-11
Remove bool flag force_vector from enkf_node_store()

joakim-hove:
**Issue**
Drive by improvement/fix when working with: #1069

A long time ago - when ert was an implementation of the EnKF algorithm all data objects were stored with two coordinates, timestep and realization number. Since for the summary variables one such data object was just one single double value file system access was very inefficient. To improve on that *vector storage* was introduced to store/load an entire timeseries in one go. That was experimental - and optional ~10 years ago, and there were boolean switches to regulate. That boolean switch was just hardcoded to `False` in the `enkf_node.py` implementation, and that bit me while creating a test. With this PR the vector storage unconditionally on for summary variables (which it has been for ~10 years).


ertomatic:

ManInFez:
Test this please!

ManInFez (approved):

PR #1115, opened 2021-02-12
Store ensemble config nodes in an ordered map

markusdregi:
**Issue**
We observe different samplings for iteration 0 in regressions tests for deploys built internally vs deploys built on GitHub Action and then fetched from PyPI.

**Approach**
Store the parameter nodes in an ordered structure (`map`) instead of `unordered_map`.


jondequinor (approved):
As we say in libres/ert, let's cross our fingers 🤞

markusdregi:
Figured the commit deserves a message body. Will merge when CI passes again...

PR #1119, opened 2021-02-23
remove fs dependency for summary observation collector

ManInFez:
Summary observation collector is using the fs to set the time map for  observation data. Now we use the observation.getObservation time instead, reducing the need for fs when collecting the observation data.

### DIFF
--- a/python/res/enkf/export/summary_observation_collector.py
+++ b/python/res/enkf/export/summary_observation_collector.py
@@ -30,40 +30,32 @@ class SummaryObservationCollector(object):
         @type keys: list of str
         @rtype: DataFrame
         """
-        fs = ert.getEnkfFsManager().getFileSystem(case_name)
-
-        time_map = fs.getTimeMap()
-        dates = [time_map[index].datetime() for index in range(1, len(time_map))]
-
+        observations = ert.getObservations()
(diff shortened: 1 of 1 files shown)


python/res/enkf/export/summary_observation_collector.py
```diff
@@ -30,40 +30,32 @@ def loadObservationData(ert, case_name, keys=None):
...
-        fs = ert.getEnkfFsManager().getFileSystem(case_name)
-
-        time_map = fs.getTimeMap()
-        dates = [time_map[index].datetime() for index in range(1, len(time_map))]
-
+        observations = ert.getObservations()
+        history_length = ert.getHistoryLength()
+        dates = [
+            observations.getObservationTime(index).datetime()
+            for index in range(1, history_length)
```
oyvindeide, line 37: Don\`t think I advocate this, but to replicate the old behavior fully we would need to do:
`for index in range(1, history_length +1)`
what that extra does, I don\`t know, as the observation always (at least the examples I have seen) comes out as `nan` on that extra point.

oyvindeide (approved):
LGTM! Good to get rid of `storage` dependencies where we can

PR #1120, opened 2021-02-26
Revert grdecl

joakim-hove:
**Issue**
Revert offending commit from https://github.com/equinor/ert/issues/1423

**Approach**
"git revert" + rewrite testdata files to ensure they are long enough for the file-type detection algorithm

### DIFF
--- a/tests/res/enkf/test_field_config.py
+++ b/tests/res/enkf/test_field_config.py
@@ -30,9 +30,8 @@ class FieldConfigTest(ResTest):
                 f.write("-- my comment\n")
                 f.write("-- more comments\n")
                 f.write("SOWCR\n")
-                for i in range(128 // 8):
+                for i in range(256 // 8):  # technicalities demand file has >= 256B
                     f.write("0 0 0 0\n")
-                f.write("/\n")
 
             ft = FieldConfig.guessFiletype(fname)
             grdecl_type = EnkfFieldFileFormatEnum(5)
--- a/tests/res/enkf/test_field_config.py
+++ b/tests/res/enkf/test_field_config.py
@@ -30,7 +30,11 @@ class FieldConfigTest(ResTest):
                 f.write("-- my comment\n")
                 f.write("-- more comments\n")
                 f.write("SOWCR\n")
-                for i in range(256 // 8):  # technicalities demand file has >= 256B
+                # The function guessing file types determines whether the file
+                # is binary or 7 bit ASCII based on bit 8 heuristics. For this
+                # to be "reliable" the file is required to be more than 256
+                # bytes.
+                for i in range(256 // 8):
                     f.write("0 0 0 0\n")
 
             ft = FieldConfig.guessFiletype(fname)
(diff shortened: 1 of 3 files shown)


ertomatic:

ManInFez:
Test this please!

tests/res/enkf/test_field_config.py
```diff
@@ -30,9 +30,8 @@ def test_field_guess_filetype(self):
                 f.write("-- my comment\n")
                 f.write("-- more comments\n")
                 f.write("SOWCR\n")
-                for i in range(128 // 8):
+                for i in range(256 // 8):  # technicalities demand file has >= 256B
```
jondequinor: Could you expand on this? What technicalities

joakim-hove: This commit is old - I tried (not hard enough) to eradicate it, bringing it back in now.

The point is that deep down there is some code which need to know whether en eclipse file should be opened as ASCII or binary, that code uses some heuristics to guess whether the file is binary or not. The heuristic is based on checking bit 8 - and if bit 8 in the bytestream is set in sufficiently many bytes the files is assumed to be binary -  now this heuristic needs a filesize before it willing to make a guess. This code is soon old enough to vote and should probably be taken out.

jondequinor: Okay, thank you. Can you add this explanation to the comment? This comment is going to disappear, the comment will add value.

Also, why do we have to guess? Can't the user inform us in some way? The way I see it, guessing on binary data is a complete gamble.

joakim-hove: > Can you add this explanation to the comment?

> Also, why do we have to guess?

Don't have to of course - but that is the way it is implemented right now. Typical situation is the `INIT_FILES:` attribute of a field configuration:
```
FIELD PORO PARAMETER poro.grdecl INIT_FILES:fields/poro%d
```
The files in the `INIT_FILES:` attribute can be of the types: 
1. Eclipse grdecl file
2. ASCII Eclipse restart file
3. BInary Eclipse restart file
4. RMS roff file

And that file type is determined/guessed based on file content - with no attention to naming conventions.

jondequinor: Great, thank you 👌🏻

ManInFez:
Test this please!

ManInFez (approved):

tests/res/enkf/test_field_config.py
lars-petter-hauge: Edit: never mind, I misunderstood where the filename came from, the argument to `INIT_FILES` doesn't include the extension then. Leaving the comment here though just to have possible extensions listed.

Or.. you could of course check the extension of the file given by the argument, don't need the user to tell us more than the basename.

So the possible files can have the following extensions:

1. `GRID`, `FGRID`, `EGRID` and `FEGRID` (Reflects: BINARY / ASCII FORMATTED / BINARY EXTENDED/ ASCII FORMATTED EXTENDED)
2. `UNRST` and `Xnnnn`  (Reflects: BINARY / BINAR (per step))
3.  `FUNRST` and `Fnnnn` (Reflects:  FORMATTED / FORMATTED (per step))
4. `ROFF`

Not sure if I've forgotten any?

Is there a reason why we couldn't check on the specific extensions?

@joakim-hove 

I wouldn't be surprised if there was a reason for this being implemented in a different manner.

PR #1126, opened 2021-03-14
Row scaling assign field

joakim-hove:
**Issue**
Part of: #1069 

**Approach**
In this PR is a new function `row_scaling::assign_vector()` which will assign a row scaling value for all the elements in the vector. This is convenience/efficiency function for the case where the row_scaling has is loaded in it's entirety from file - typically because it has been prepared with an external application.

Observe the last commit which chganges a default.

### DIFF
--- a/lib/enkf/tests/row_scaling.cpp
+++ b/lib/enkf/tests/row_scaling.cpp
@@ -118,8 +118,11 @@ void test_multiply() {
   {
     row_scaling_type * row_scaling = row_scaling_alloc();
     matrix_type * A = matrix_alloc_copy(A0);
+    std::vector<float> row_data(data_size);
     for (int row = 0; row < data_size; row++)
-      row_scaling_iset(row_scaling, row, 0);
+      row_data[row] = 0;
+
+    row_scaling_assign_float(row_scaling, row_data.data(), row_data.size());
     row_scaling_multiply(row_scaling, A, X0);
 
     for (int row = 0; row < data_size; row++)
@@ -136,8 +139,16 @@ void test_multiply() {
   {
     row_scaling_type * row_scaling = row_scaling_alloc();
     matrix_type * A = matrix_alloc_copy(A0);
+    std::vector<double> row_data(data_size);
+
+    row_scaling_iset(row_scaling, 2*data_size, 1.0);
+    test_assert_int_equal( row_scaling_get_size(row_scaling), 2*data_size + 1);
+
     for (int row = 0; row < data_size; row++)
-      row_scaling_iset(row_scaling, row, rng_get_double(rng));
+      row_data[row] = rng_get_double(rng);
+
+    row_scaling_assign_double(row_scaling, row_data.data(), row_data.size());
+    test_assert_int_equal( row_scaling_get_size(row_scaling), data_size);
 
     row_scaling_multiply(row_scaling, A, X0);
     for (int row = 0; row < data_size; row++) {
--- a/python/res/enkf/row_scaling.py
+++ b/python/res/enkf/row_scaling.py
@@ -14,6 +14,8 @@
 #  See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
 #  for more details.
 
+import numpy as np
+import ctypes
 from cwrap import BaseCClass
 from res import ResPrototype
 
@@ -27,6 +29,12 @@ class RowScaling(BaseCClass):
     _iset = ResPrototype("double row_scaling_iset(row_scaling, int, double)")
     _iget = ResPrototype("double row_scaling_iget(row_scaling, int)")
     _clamp = ResPrototype("double row_scaling_clamp(row_scaling, double)")
+    _assign_double_vector = ResPrototype(
+        "void row_scaling_assign_double(row_scaling, double*, int)"
+    )
+    _assign_float_vector = ResPrototype(
+        "void row_scaling_assign_float(row_scaling, float*, int)"
+    )
 
     def __init__(self):
         c_ptr = self._alloc()
@@ -103,3 +111,54 @@ class RowScaling(BaseCClass):
 
         for index in range(target_size):
             self[index] = func(index)
+
+    def assign_vector(self, scaling_vector):
+        """Assign tapering value for all elements via a vector.
+
+        The assign_vector() function will resize the row_scaling vector to the
+        number of elements in the scaling_vector, and assign a value to all
+        elements.
+
+        A typical situation might be to load the scaling data as a EclKW
+        instance from a grdecl formatted file. Before the assign_vector() can
+        be called we must transform to an only active elements representation
+        and use a numpy view:
+
+            # Load scaling vector from grdecl file; typically created with
+            # geomodelling software.
+            with open("scaling.grdecl") as fh:
+                kw_global = EclKW.read_grdecl(fh, "SCALING")
+
+            # Create a ecl_kw copy with only the active elements.
+            kw_active = grid.compressed_kw_copy(kw_global)
+
+            # Create a numpy view and invoke the assign_vector() function
+            row_scaling.assign_vector(kw_active.numpy_view())
+
+        """
+        # In current implementation the scaling vector must be of type
+        # numpy.array(float32/float64); it would possible to also accept ecl_kw
+        # and call the same underlying C functions. Currently the expected
+        # approach is rather that calling scope should convert ecl_kw -> numpy.
+        #
+        # In addition it would be possible to also accept a general Python
+        # iterable and repeatedly call __setitem__() from the assign_vector
+        # implementation. This solution has been avoided has been avoided
+        # because it would not be very performant.
+        if isinstance(scaling_vector, np.ndarray):
+            if scaling_vector.dtype == np.float64:
+                func = self._assign_double_vector
+                ptr = scaling_vector.ctypes.data_as(ctypes.POINTER(ctypes.c_double))
+            elif scaling_vector.dtype == np.float32:
+                func = self._assign_float_vector
+                ptr = scaling_vector.ctypes.data_as(ctypes.POINTER(ctypes.c_float))
+            else:
+                raise TypeError(
+                    f"The scaling_vector must be float32/float64, not {type(scaling_vector.dtype)}"
+                )
+
+            func(ptr, len(scaling_vector))
+        else:
+            raise TypeError(
+                f"The assign_vector method expects a numpy vector as argument, not {type(scaling_vector)}"
+            )
--- a/tests/res/enkf/test_row_scaling_case.py
+++ b/tests/res/enkf/test_row_scaling_case.py
@@ -17,6 +17,7 @@
 import random
 import os
 import math
+import numpy as np
 
 from tests import ResTest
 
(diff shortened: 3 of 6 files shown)


ertomatic:

ManInFez:
Test this please!

tests/res/enkf/test_row_scaling_case.py
```diff
@@ -297,3 +298,75 @@ def test_update_code2(self):
                     update_node1.asField(),
                     update_node2.asField(),
                 )
+
+    # This test is identical to test_update_code2(), but the row scaling is
+    # applied with the function row_scaling.scaling_assign_vector() instead of
```
ManInFez:
```suggestion
    # applied with the function row_scaling.assign_vector() instead of
```
This is the actual function name, right?

ManInFez:
Why are the default changing to from 1 to 0?

tests/res/enkf/test_row_scaling_case.py
joakim-hove: Yes - that is right; I have updated.

joakim-hove:
> Why are the default changing to from 1 to 0?

This was no obvious choice; but I think 0 is a better default than 1. The row scaling approach is more general than the existing region based approach, and in particular everything which can be done with the existing region based approach can be done with row_scaling and scaling factors 0 and 1. If we default the row scaling value to 0 we are closer to a situation where the old approach can be completely removed. That said it will be difficult to envisage setting only parts of property with row_scalng - and then defaulting the rest to zero (i.e. not in use); but it is certainly possible. 

In addition to the current assign_vector() function one could add something like `assign_vector_pairs()` - in pseudo code[*]:
```
def assign_vector_pairs( self, pairs ):
      for index, value in pairs:
            self[index] = value
```
then the  default of 0 would make very much sense. 

[*]: To become efficient in C - which would be part of the motivation here one would probably have to have two arrays of scalars - which is less elegant; but anyway.

lib/enkf/tests/row_scaling.cpp
```diff
@@ -118,9 +118,11 @@ void test_multiply() {
   {
     row_scaling_type * row_scaling = row_scaling_alloc();
     matrix_type * A = matrix_alloc_copy(A0);
+    std::vector<float> row_data(data_size);
     for (int row = 0; row < data_size; row++)
-      row_scaling_iset(row_scaling, row, 0);
-    row_scaling_multiply(row_scaling, A, X0);
```
ManInFez, line 123: I don't understand why the multiply call is removed

ManInFez:
Test this please!

lib/enkf/tests/row_scaling.cpp
joakim-hove: Because the same operation is now performed with the `row_scaling_assign_float()` function call.

python/res/enkf/row_scaling.py
```diff
@@ -103,3 +111,52 @@ def exp_decay(grid, pos, r0, data_index):
...
+        # because it would not be very performant.
+        if isinstance(scaling_vector, np.ndarray):
+            if scaling_vector.dtype == np.float64:
+                func = self._assign_double_vector
+                ptr = scaling_vector.ctypes.data_as(ctypes.POINTER(ctypes.c_double))
+            elif scaling_vector.dtype == np.float32:
+                func = self._assign_float_vector
+                ptr = scaling_vector.ctypes.data_as(ctypes.POINTER(ctypes.c_float))
+            else:
+                raise TypeError("The scaling_vector must be float32/float64")
```
jondequinor:
```suggestion
                raise TypeError(f"The scaling_vector must be float32/float64, not {type(scaling_vector.dtype)}")
```

python/res/enkf/row_scaling.py
```diff
@@ -103,3 +111,52 @@ def exp_decay(grid, pos, r0, data_index):
...
+
+            func(ptr, len(scaling_vector))
+        else:
+            raise TypeError(
+                "The assign_vector method expects a numpy vector as argument"
```
jondequinor:
```suggestion
                f"The assign_vector method expects a numpy vector as argument, not {type(scaling_vector)}"
```

tests/res/enkf/test_row_scaling_case.py
```diff
@@ -297,3 +298,75 @@ def test_update_code2(self):
                     update_node1.asField(),
                     update_node2.asField(),
                 )
+
+    # This test is identical to test_update_code2(), but the row scaling is
+    # applied with the function row_scaling.assign_vector() instead of
+    # using a callable.
+    def test_update_code3(self):
```
jondequinor:
```suggestion
    def test_row_scaling_using_assign_vector(self):
```

python/res/enkf/row_scaling.py
```diff
@@ -103,3 +111,52 @@ def exp_decay(grid, pos, r0, data_index):
 
         for index in range(target_size):
             self[index] = func(index)
+
+    def assign_vector(self, scaling_vector):
```
jondequinor, line 115: Why isn't `assign_vector` implemented here, in the python layer, as a specialized call to `assign`?

python/res/enkf/row_scaling.py
```diff
@@ -103,3 +111,52 @@ def exp_decay(grid, pos, r0, data_index):
...
+            # Load scaling vector from grdecl file; typically created with
+            # geomodelling software.
+            with open("scaling.grdecl") as fh:
+                kw_global = EclKW.read_grdecl(fh, "SCALING")
+
+            # Create a ecl_kw copy with only the active elements.
+            kw_active = grid.compressed_kw_copy(kw_global)
+
+            # Create a numpy view and invoke the assign_vector() function
+            row_scaling.assign_vector(kw_active.numpy_view())
```
jondequinor, line 136: If this is typical usage, then is it API somewhere, or do you have to read this doc in order to use it correctly?

python/res/enkf/row_scaling.py
```diff
@@ -103,3 +111,52 @@ def exp_decay(grid, pos, r0, data_index):
...
+        """
+        # In current implementation the scaling vector must be of type
+        # numpy.array(float32/float64); it would possible to also accept ecl_kw
+        # and call the same underlying C functions. Currently the expected
+        # approach is rather that calling scope should convert ecl_kw -> numpy.
+        #
+        # In addition it would be possible to also accept a general Python
+        # iterable and repeatedly call __setitem__() from the assign_vector
+        # implementation. This solution has been avoided has been avoided
+        # because it would not be very performant.
```
jondequinor, line 147: We should not do early optimization, can we start with the python solution?

tests/res/enkf/test_row_scaling_case.py
joakim-hove: OK

python/res/enkf/row_scaling.py
joakim-hove: I perceive this as typical usage - that is why I have added a quite extensive doc string. But I do not really understand your question: ", then is it API somewhere"?

lib/enkf/tests/row_scaling.cpp
ManInFez: I am not seeing how these are the .  Would love to get an explanation :)

joakim-hove:
OK - I have updated a trivial rename, the other comments are all related, so we must agree on a way forward when it comes to `assign_vector` and the use of an underlying C implementation. As I see it there are three possibilities:

**Alternative 1: Only C based implementation** That is the choice taken in this PR. If we stick with this approach I'll just update the error messages and we should be done.

**Alternative 2: Mixed implementation** In the case the vector argument is not a numpy array we can iterate over whatever iterable comes in and repeatedly call `assign()`:

```Python
def assign_vector(self, scaling_vector):
      if isinstance(scaling_vector, np.ndarray):
           # Call C routines 
      else:
           for index,value in enumerate(scaling_vector):
                self[index] = value
```


**Alternative 3: Pure python implementation** In this case the C functions can be just removed and everything is a simple Python function:
```Python
def assign_vector(self, scaling_vector):
      for index,value in enumerate(scaling_vector):
           self[index] = value
```

How to respond to the comment about error message depends on the alternative chosen above. Personally I would like to stick with alternative 1. As illustrated in the doc string I think loading the row_scaling from a grdecl formatted file will be a very natural way to proceed, and then it makes sense to make that approach performant - in my opinion.

python/res/enkf/row_scaling.py
jondequinor: Am being naive here, so bear with me. But if you need many preconditions for something to be used, it's a poor API. But perhaps most users do the steps you mention in this docstring? If that's the case, then it's fine as it is.

If users would _have_ to do these steps, and the only thing they really want to do is call `assign_vector`given a ecl_kw and grdecl file—then this "should be API somewhere". Meaning the code in your docstring should be implemented in libres (preferrably a combination with ERT).

lib/enkf/tests/row_scaling.cpp
joakim-hove: I am sorry you are right; the multiply operation should be moved after the new `row_scaling_assign_float()`function call but not completely removed.

My mistake. Fixed.

python/res/enkf/row_scaling.py
joakim-hove: OK - the fundamental low level operation which is focused on here is
```Python
def assign_vector(self, numpy_scaling_vector)
     ...
```
you do *not* need a grdecl formatted file or a ecl_kw instance beforehand, you could just as well calculate the numpy_scaling vector by other means - for advanced usage that is probably the most interesting. 

Iff people use alternative software (typically geomodelling) to calculate the entire row_scaling vector in one go I think the approach outlined in the doc string makes sense.

oddvarlia:
I talk from the users point of view, a user that want to apply the python code to build 3rd layer (ordinary user level) functionality where config file(s) is used to specify localization, not python scripting:
1. I have so far used the index operation   row_scaling[index] = scaling  of the row_scaling functionality. I saw the usefulness of creating the scaling factors outside of ERT, e.g. by RMS and deliver the scaling factor to ERT (by e.g grdecl file format or preferably by ROFF format if possible as in the ERT FIELD keyword). It is therefore Joakim implemented the possibility to assign the scaling values by a vector in addition to the other method(s). As long as it is easy for a user like me to find the documentation of usage of the various row_scaling assigment methods, that is most important, not actually what the required steps are to prepare the input vector. It is however only for very large cases with FIELDS that some performance is gained with vector as input instead of looping over indices in python code for the localization setup script.

2. About default:  I can see examples where I would like to loop over a subset of all indices for a scaling vector for a FIELD and assign those to the row_scaling and in this case set as scaling factor to the other grid cells equal to 0. But with a default of 0, I can skip assigning the 0 value for the indices not belonging to the subset of indices that I want to use. I have already done that in my testing to create regions (assigning 0 all other indices than those I want to use, like a region). So I think 0 as default value is a good choice.

**Another question is what flexibility we would see a need for when developing the 3rd layer (ordinary user level):**
1. The current way of specifying localization is working in the way that as soon as one ministep for localization is defined, it means that ALL correlations between any observation and any model parameter is set to 0 EXCEPT for the observations and model parameters attached to the specified ministep.  **We could easily think that the opposite could be relevant as well:** Specify which correlations NOT to include (Set scaling factor to 0 for that specified set of correlations). But I think this is not possible now in the way localization is implemented. Therefore setting up localization requires specification of which correlations to include and that can be very  much in cases the goal is to remove correlations from a a few observations correlated with a few model parameters. **I think this is a much more important issue.**

python/res/enkf/row_scaling.py
jondequinor: Thanks! 👍

jondequinor:
> Alternative 1: Only C based implementation That is the choice taken in this PR. If we stick with this approach I'll just update the error messages and we should be done.

Thanks going through this. I'm fine with alternative 1.

joakim-hove:
> I'm fine with alternative 1.

OK - thanks - I'll update.

joakim-hove:
> Another question is what flexibility we would see a need for when developing the 3rd layer (ordinary user level): [...]

Things along this direction is already possible; might not be fully well exposed in Python api - will look.

ManInFez:
Test this please

ManInFez (approved):

PR #1127, opened 2021-03-16
Test 2ministep

joakim-hove:
**Issue**
Part of #1069 

**Approach**
The purpose of this PR was to create a row_scaling test which demonstrates that two consecutive ministeps can be the same as one larger. However due to the internal random sampling it is difficult to get identical results - the test is therefor not that interesting.

The actual interesting test is in the last commit, in addition there are some small improvements/fixes in the preceeding commits

### DIFF
--- a/lib/enkf/enkf_main.cpp
+++ b/lib/enkf/enkf_main.cpp
@@ -1560,9 +1560,11 @@ void enkf_main_create_all_active_config( const enkf_main_type * enkf_main) {
   local_config_clear( local_config );
   {
     local_updatestep_type * default_step = local_config_get_updatestep(local_config);
-    local_ministep_type * ministep = local_config_alloc_ministep( local_config , "ALL_ACTIVE", NULL);
     local_obsdata_type * obsdata = local_config_alloc_obsdata(local_config, "ALL_OBS");
     local_dataset_type * all_active_dataset = local_config_alloc_dataset(local_config, "ALL_DATA");
+    local_ministep_type * ministep = local_config_alloc_ministep( local_config , "ALL_ACTIVE", NULL);
+    if (!ministep)
+      throw std::logic_error("Failed to create initial ALL_ACTIVE ministep");
 
     local_updatestep_add_ministep( default_step , ministep );
 
--- a/lib/enkf/local_config.cpp
+++ b/lib/enkf/local_config.cpp
@@ -172,6 +172,9 @@ void local_config_free(local_config_type * local_config) {
 }
 
 local_ministep_type * local_config_alloc_ministep( local_config_type * local_config , const char * key, analysis_module_type* analysis_module) {
+  if (hash_has_key(local_config->ministep_storage, key))
+    return nullptr;
+
   local_ministep_type * ministep = local_ministep_alloc( key, analysis_module );
   hash_insert_hash_owned_ref( local_config->ministep_storage , key , ministep , local_ministep_free__);
   return ministep;
--- a/python/res/enkf/data/field.py
+++ b/python/res/enkf/data/field.py
@@ -24,7 +24,9 @@ class Field(BaseCClass):
     TYPE_NAME = "field"
 
     _free = ResPrototype("void field_free( field )")
+    _get_size = ResPrototype("int field_get_size(field)")
     _ijk_get_double = ResPrototype("double field_ijk_get_double(field, int, int, int)")
+    _iget_double = ResPrototype("double field_iget_double(field, int)")
     _export = ResPrototype(
         "void field_export(field, char* , fortio , enkf_field_file_format_enum , bool , char*)"
     )
@@ -32,8 +34,17 @@ class Field(BaseCClass):
     def __init__(self):
         raise NotImplementedError("Class can not be instantiated directly!")
 
+    def __len__(self):
+        return self._get_size()
+
     def ijk_get_double(self, i, j, k):
         return self._ijk_get_double(i, j, k)
+
+    def __getitem__(self, index):
+        if 0 <= index < len(self):
+            return self._iget_double(index)
+        else:
+            raise IndexError(f"Index: {index} out of range: [0,{len(self)})")
 
     def export(self, filename, file_type=None, init_file=None):
         output_transform = False
--- a/tests/res/enkf/test_row_scaling_case.py
+++ b/tests/res/enkf/test_row_scaling_case.py
@@ -114,6 +114,19 @@ class GaussianDecay(object):
         return math.exp(exp_arg)
 
 
+class SelectLayer(object):
+    def __init__(self, layer, grid):
+        self.layer = layer
+        self.grid = grid
+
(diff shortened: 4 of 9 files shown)


ertomatic:

lib/enkf/local_config.cpp
```diff
@@ -172,6 +172,9 @@ void local_config_free(local_config_type * local_config) {
 }
 
 local_ministep_type * local_config_alloc_ministep( local_config_type * local_config , const char * key, analysis_module_type* analysis_module) {
+  if (hash_has_key(local_config->ministep_storage, key))
```
ManInFez, line 175: are we sure that this wont affect other users?

ManInFez:
Test this please!

lib/enkf/local_config.cpp
joakim-hove: Well -  current master leads to a use-after-free situation; so this fix I am confident about.

ManInFez:
#1126 is merged. Rebase this on top of master?

joakim-hove:
Rebased!

ManInFez:
Test this please!

python/res/enkf/data/field.py
```diff
@@ -24,17 +24,28 @@ class Field(BaseCClass):
...
+        return self._get_size()
+
     def ijk_get_double(self, i, j, k):
         return self._ijk_get_double(i, j, k)
 
+    def __getitem__(self, index):
+        if index < len(self):
+            return self._iget_double(index)
+        else:
+            raise IndexError("Index: {} out of range: [0,{})", index, len(self))
```
ManInFez:
```suggestion
            raise IndexError(f"Index: {index} out of range: [0,{len(self)})")
```

lib/enkf/local_config.cpp
ManInFez: Very well, but it should either exit with a meaningful error or it should be dealt with here as well:
https://github.com/joakim-hove/libres/blob/388d0b29eb86a6745d5847f486c324ff0544dc39/lib/enkf/enkf_main.cpp#L1563

python/res/enkf/data/field.py
joakim-hove: OK

lib/enkf/local_config.cpp
joakim-hove: I don't really understand what you request here. The python part - which is the user accessible will raise `KeyError()` - when it comes to the C++ function you link to that is called exactly once at object construction/initialization. I can of course add a 
```C++
if (!ministep) 
   throw std::logic_error("Failed to create ALL_ACTIVE ministep");
```
- or maybe I misunderstood your request?

ManInFez: That is what I meant :) As we are changing the behavior of the function, it should be reflected wherever the function is used.

python/res/enkf/data/field.py
```diff
@@ -24,17 +24,28 @@ class Field(BaseCClass):
...
         raise NotImplementedError("Class can not be instantiated directly!")
 
+    def __len__(self):
+        return self._get_size()
+
     def ijk_get_double(self, i, j, k):
         return self._ijk_get_double(i, j, k)
 
+    def __getitem__(self, index):
+        if index < len(self):
```
jondequinor:
```suggestion
        if index >= 0 and index < len(self):
```

tests/res/enkf/test_row_scaling_case.py
```diff
@@ -370,3 +383,82 @@ def test_row_scaling_using_assign_vector(self):
...
+                field2 = node2.asField()
+                for iv, v in zip(init_field, field2):
+                    assert iv != v
+
+                    # If it were not for the random sampling in the
+                    # obs_data_allocE() function called once per ministep in
+                    # enkf_main_analysis_update() - we should have had this
+                    # assert:
+                    #
+                    # assert v1 == v2
```
jondequinor:
```suggestion
```

Maybe add it to the top comment as a NOTE

python/res/enkf/data/field.py
joakim-hove: ok

lib/enkf/local_config.cpp
joakim-hove: ok

jondequinor:
> The purpose of this PR was to create a row_scaling test which demonstrates that two consecutive ministeps can be the same as one larger.

What does mean? The same as one larger? How is this connected to the diff?

tests/res/enkf/test_row_scaling_case.py
joakim-hove: Essentially the same information is already present at the top - I have just removed it.

lib/enkf/enkf_main.cpp
```diff
@@ -1560,9 +1560,11 @@ void enkf_main_create_all_active_config( const enkf_main_type * enkf_main) {
   local_config_clear( local_config );
   {
     local_updatestep_type * default_step = local_config_get_updatestep(local_config);
-    local_ministep_type * ministep = local_config_alloc_ministep( local_config , "ALL_ACTIVE", NULL);
     local_obsdata_type * obsdata = local_config_alloc_obsdata(local_config, "ALL_OBS");
     local_dataset_type * all_active_dataset = local_config_alloc_dataset(local_config, "ALL_DATA");
+    local_ministep_type * ministep = local_config_alloc_ministep( local_config , "ALL_ACTIVE", NULL);
+    if (!ministep)
+      throw std::logic_error("Failed to create initial ALL_ACTIVE ministep")
```
ManInFez:
```suggestion
      throw std::logic_error("Failed to create initial ALL_ACTIVE ministep");
```

joakim-hove: Yes

ManInFez:
Test this please!

joakim-hove:
>>The purpose of this PR was to create a row_scaling test which demonstrates that two consecutive ministeps can be the same as one larger.

> What does mean? The same as one larger? How is this connected to the diff?

Well; 

I was asked by user whether it was possible/convenient to split an update with row scaling over two ministeps - so in order to demonstrate (and verify) this possibility I created this PR. The main content of this PR is the last commit which adds the test method:
```Python
def test_2ministep(self):
      ....
      ....
```
This test uses the row_scaling functionality, the first ministep scales layer 0 and the second ministep scales layer 1. Since the  updates involved are linear this should give identical results as one ministep update with scaling to both layers[1]. Due to stochastic sampling internally we unfortunately do not get identical results, but they are compared approximately.

[1]: Observe that the situation scaling to both layers is not explicit in the test code. The first update is just the defauilt update everything without any scaling case, and the scaling chosen for update two is 0 / 1 to ensure that the net effect is the same.

ManInFez:
Test this please!

ManInFez (approved):

jondequinor:
Thanks for explaining @joakim-hove!

PR #1129, opened 2021-03-18
Add Python 3.9 to CI

pinkwah:


xjules (approved):
LGTM!

.github/workflows/python-bdist.yml
```diff
@@ -1,135 +0,0 @@
-name: Python Binary Distribution
```
markusdregi, line 1: Why is this workflow going away?

markusdregi (changes requested):
Please explain why the bdist-workflow disappears before merging. Besides that, Python 3.9 support is neat 😎

pinkwah:
It's a workflow that does nearly the same as `testing`. Except it doesn't run on PRs. So we have two workflows that eg. upload to PyPI. You'll notice that `master` fails CI most of the time. Most likely someone forgot to delete it.

markusdregi:
Thanks for the explanation @dotfloat! As the removal of the workflow is independent of introducing Python 3.9, can you please separate it into a separate commit? And in addition, rename the now `testing.yml` so that it becomes clear that is also does `distribution`? I applaud the cleaning of the workflow files though 👏🏻

markusdregi (approved):
📦

PR #1132, opened 2021-03-22
Make TEMPLATE_RENDER robust towards no parameters.json

### LINKED ISSUE #1128: Make template_render run without a parameters.json file present
Both during testing with ert3, potential Everest usage and DESIGN_KW based sensitivity studies the wish for the possibility to run `template_render` without a `parameters.json` file present has appeared. Instead, only files provided via the `--input-files` argument will be used.

I think both should be catered for, so a sensible set of cli flags or similar will have to be added.

markusdregi:
**Issue**
Resolves #1128


sondreso (approved):
LGTM!

PR #1133, opened 2021-03-22
Matrix size

### LINKED ISSUE #1131: New localisation functionality using row_scaling fails if grid size exceeds 62x62x62
Preliminary testing of row_scaling functionality fails if grid size exceeds 62x62x62. The error message (if not segmentation fault appear) is Size mismatch between row_scaling and A matrix.
[Test localisation using new functionality.pptx](https://github.com/equinor/libres/files/6178615/Test.localisation.using.new.functionality.pptx)

joakim-hove:
**Issue**
Part of #1069 Fixes: https://github.com/equinor/libres/issues/1131

**Approach**
Ensure that resizing of the A matrix works together with row scaling.

### DIFF
--- a/lib/enkf/enkf_main.cpp
+++ b/lib/enkf/enkf_main.cpp
@@ -517,8 +517,7 @@ typedef struct {
 
 
 static void serialize_node( enkf_fs_type * fs ,
-                            const ensemble_config_type * ensemble_config,
-                            const char * key ,
+                            const enkf_config_node_type * config_node,
                             int iens ,
                             int report_step ,
                             int row_offset ,
@@ -526,7 +525,6 @@ static void serialize_node( enkf_fs_type * fs ,
                             const active_list_type * active_list,
                             matrix_type * A) {
 
-  const enkf_config_node_type * config_node = ensemble_config_get_node( ensemble_config , key );
   enkf_node_type * node = enkf_node_alloc( config_node );
   node_id_type node_id = {.report_step = report_step, .iens = iens  };
   enkf_node_serialize( node , fs , node_id , active_list , A , row_offset , column);
@@ -536,14 +534,13 @@ static void serialize_node( enkf_fs_type * fs ,
 
 static void * serialize_nodes_mt( void * arg ) {
   serialize_info_type * info = (serialize_info_type *) arg;
-  int iens;
-  for (iens = info->iens1; iens < info->iens2; iens++) {
+  const auto * node_info = info->node_info;
+  const enkf_config_node_type * config_node = ensemble_config_get_node( info->ensemble_config , node_info->key );
+  for (int iens = info->iens1; iens < info->iens2; iens++) {
     int column = int_vector_iget( info->iens_active_index , iens);
     if (column >= 0) {
-      const auto * node_info = info->node_info;
       serialize_node( info->src_fs ,
-                      info->ensemble_config,
-                      node_info->key ,
+                      config_node,
                       iens ,
                       info->report_step ,
                       node_info->row_offset ,
@@ -627,15 +624,13 @@ static int enkf_main_serialize_dataset( const ensemble_config_type * ens_config 
 
 static void deserialize_node( enkf_fs_type * target_fs,
                               enkf_fs_type * src_fs,
-                              const ensemble_config_type * ensemble_config,
-                              const char * key ,
+                              const enkf_config_node_type * config_node,
                               int iens,
                               int target_step ,
                               int row_offset ,
                               int column,
                               const active_list_type * active_list,
                               matrix_type * A) {
-  const enkf_config_node_type * config_node = ensemble_config_get_node( ensemble_config , key );
 
   node_id_type node_id = {.report_step = target_step, .iens = iens  };
   enkf_node_type * node = enkf_node_alloc( config_node );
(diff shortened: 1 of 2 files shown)


ertomatic:

ManInFez:
Test this please!

lib/enkf/enkf_main.cpp
```diff
@@ -1139,11 +1139,17 @@ static void enkf_main_analysis_update( enkf_main_type * enkf_main ,
             for (int ikw=0; ikw < scaled_keys.size(); ikw++) {
               const auto& key = scaled_keys[ikw];
               const active_list_type * active_list = local_dataset_get_node_active_list(dataset, key.c_str());
+              const auto matrix_rows = matrix_get_rows(A);
+              const auto* config_node = ensemble_config_get_node(serialize_info->ensemble_config, key.c_str());
+              const auto node_size = enkf_config_node_get_data_size(config_node, serialize_info->report_step);
+              if (matrix_get_rows(A) < node_size)
+                  matrix_resize( A , node_size , active_ens_size , false);
```
ManInFez, line 1146: Is the A matrix empty at this point?

joakim-hove: No - the matrix is created with a default size - variable `matrix_start_size` further up in the same function.

ManInFez: ok. But why is the A matrix resized and not the row_scaling?

joakim-hove:
Rebased!

ManInFez:
#1127 merged, needs a rebase :)

joakim-hove:
rebased

tests/res/enkf/test_row_scaling_case.py
```diff
@@ -458,3 +463,57 @@ def test_2ministep(self):
 
                     diff = abs(v1 - v2)
                     assert diff < 0.05
```
ManInFez: This fails on python 3.9, why is it set to 0.05?

joakim-hove: The problem here is that there is a stochastic element and the test is just "somewhat close" - 0.05 was OK on my machine. I realize this is an extremely weak test, but to actually get control over the random state to the level needed to get this identical is a quite large job.

I have increased the test ensemble size which should reduce fluctuations but of course also increase runtime; alternatively the actual numerical test can just be removed?

ManInFez:
Test this please!

tests/res/enkf/test_row_scaling_case.py
ManInFez: I am worried that this flakyness will hit us at later point in time. Maybe it is best to remove it?

joakim-hove: I don't agree with the term flakyness - the situation is understood; but I agree that the value is very limited - happy to remove it.

ManInFez: It will not occasionally fail? or is it deterministic and will always fail on python 3.9? What does the it mean then; that the difference between one ministep and two ministeps is less than 5%?

joakim-hove: No it will not occasionally fail; by setting the seed in the test the sampling is deterministic and will be reproducible. The difference comes from the function `obs_data_allocE`. Assume that we have a reproducible stream of random numbers:
```
R = {R_1, R_2, R_3, R_4, R_5, R_6}
``` 

then in the case of one ministep with two observations and three ensemble members we will get:
```
     [ R_1  R_2 ]
E =  [ R_3  R_4 ]
     [ R_5  R_6 ]
``` 
whereas we with two ministeps with one observation in each we get the two E matrices:
```
     [R_1]           [R_4] 
E1 = [R_2],     E2 = [R_5]     
     [R_3]           [R_6] 
```

i.e. because the the ensemble members and observations are interleaved in the E matrix 2 x 1 != 2.

I must admit I was little surprised by the Python 3.9 failure - my guess is/was that there has been some change in the Python random number functionality so we do not get the same sequence on 3.9 as on 3.8 - I  have not investigated that further.

> What does the it mean then; that the difference between one ministep and two ministeps is less than 5%?

That they are *quite close*; which is what we should expect. It should be added that the difference is not 5% - the comparison is absolute.

ManInFez: Would it be possible to attach a mocked random number generator to the enkf_main object before creating E which would then create? 
```
     [R_1]           [R_4] 
E1 = [R_2],     E2 = [R_5]     
     [R_3]           [R_6] 

     [ R_1  R_4 ]
E =  [ R_2  R_5 ]
     [ R_3  R_6 ]

```

joakim-hove: >Would it be possible to attach a mocked random number generator to the enkf_main object before creating E which would then create?

Everything is possible - but I would say: *No it is not worth it*. What you suggest would certainly be a necessary step - I have not verified if it whether it is sufficient, my guess is that additional changes would be necessary.

ManInFez: Then I would say the numeric comparison is not having any value, and it can be removed. I guess the test only shows that it is possible to split up in multiple ministeps, but it is not identical to doing it in one step.

joakim-hove: >  [...] and it can be removed

I agree. I have removed it - ans reset the test ensemble size to 10.

ManInFez:
Test this please!

ManInFez (approved):

PR #1134, opened 2021-03-25
Local del obs

joakim-hove:
**Issue**
Part of #1069

**Approach**
This is not really about row_scaling - but inspired by that work. Functionality to reuse default all active configuration like `ALL_OBS`.

An active local configuration is a tree-like structure consisting of `local_updateset` at the top, the `local_updateset` can have several `local_ministep`, each `local_ministep` can have exactly one `local_obsdata` and multiple `local_dataset`- i.e. a json representation might look something like this:
```json
updatestep = 
{<local_updatestep> 
    "name": "the_only_update_step", 
    "ministep": 
    [
         {<local_ministep> 
              "name" : "ministep1", 
              "obs_data" : 
                     {<local_obsdata> 
                           "name": "local_obs1", 
                           "nodes" : { "..." }
                     }
              "dataset" : { {<local_dataset> "name": "dataset1"}
         },
        {<local_ministep> 
              "name" : "ministep2", 
              "obs_data" : 
                     {<local_obsdata> 
                           "name": "local_obs2", 
                           "nodes" : { "..." }
                     }
              "dataset" : { {<local_dataset> "name": "dataset1"}
         },
    ]
}
```
The types indicated with `<...>` are the same types managed in the `local_config` class. The master object for the configuration of the local configuration is the `local_config` class - the members look like this:
```C++
local_updatestep_type 
  local_updatestep_type * default_updatestep;   
  hash_type                    * updatestep_storage;    
  hash_type                    * ministep_storage; 
  hash_type                    * dataset_storage;
  hash_type                    * obsdata_storage;
};
```
The `default_updatestep` member points to the (only) active updatestep[1], and by tracing hat object as shown in the json represention you can drill down to the `local_ministep`, `local_obsdata` and so on. However the various objects in the local configuration do *not* own their children, rather ownership is managed by the `lcoal_config` object itself with the `xxx_storage` maps. Now the orginasation with separate store is partly due to poor C memory management - *but mostly to be able to reuse named objects* - observe how the `dataset1` is used in both ministeps in the example.
 
When ert loads it generates a default local configuration with everyuthing active. The point of this PR is to be able to reuse theis ALL_ACTIVE configuration in your own setup - by selectively removing some elements in the ALL_ACTIVE.

[1]: Some of the complexity in this code stems back to EnKF where the ensemble update was repeated for every timestep, and one could have a per timestep local updatestep. Currently that does not really make sense and letting the `local_config` class potentially manage more than one updatestep is an unneccesary complexity - it will only ever manage one updatestep.


ertomatic:

joakim-hove:
Python 3.7 / macos failure seems unrelated?

ManInFez:
Might be, restarting tests

ManInFez:
Test this please!

lib/enkf/local_config.cpp
```diff
@@ -138,11 +138,7 @@ static local_updatestep_type * local_config_alloc_updatestep( local_config_type
 
 
 void local_config_clear( local_config_type * local_config ) {
-  local_config->default_updatestep  = NULL;
   hash_clear( local_config->updatestep_storage );
-  hash_clear( local_config->ministep_storage );
```
ManInFez, line 143: I dont understand why `clear` is not clearing anymore

jondequinor:
Could you update the "Approach"? To me it doesn't really correspond with the changeset? Maybe it does, but I still think it needs some expansion.

jondequinor:
I'll let Åsmund give the green light, but I am fine with this. THanks for explaining!

lib/enkf/local_config.cpp
joakim-hove: Well - the `update_step` is the interesting for waht is the currently active config - that is cleared. The others are retained so that they can - optionally - be reused as named objects in a different to-be-defined configuration.

ManInFez (approved):

PR #1137, opened 2021-04-13
Do not produce hdf5 files output when run ecl100

### LINKED ISSUE #1111: ecl100 using eclrun should default to not produce hdf5 files
By default eclrun creates an hdf5 file with data from summary, restart and init files. Currently, nobody in EQNR is using these files and they pile up for no benefit. Pertel is capable of loading the file, but there is limited usage of this. I propose that the ecl100 job defaults to not creating this file by invoking the '--summary-conversion no' option in eclrun.

kvashchuka:
**Issue**
Resolves #1111 

**Approach**
Add flag "--summary-conversion" with default "no" to EclRun

### DIFF
--- a/python/res/fm/ecl/ecl_run.py
+++ b/python/res/fm/ecl/ecl_run.py
@@ -161,10 +161,13 @@ class EclRun(object):
 
     """
 
-    def __init__(self, ecl_case, sim, num_cpu=1, check_status=True):
+    def __init__(
+        self, ecl_case, sim, num_cpu=1, check_status=True, summary_conversion=False
+    ):
         self.sim = sim
         self.check_status = check_status
         self.num_cpu = int(num_cpu)
+        self.summary_conversion = summary_conversion
 
         # Dechipher the ecl_case argument.
         input_arg = ecl_case
@@ -253,12 +256,15 @@ class EclRun(object):
                 fileH.write("%s\n" % host)
 
     def _get_run_command(self, eclrun_config):
+        summary_conversion = "yes" if self.summary_conversion else "no"
         return [
             "eclrun",
             "-v",
             eclrun_config.version,
             eclrun_config.simulator_name,
             "{}.DATA".format(self.base_name),
+            "--summary-conversion",
+            summary_conversion,
         ]
 
     def _get_legacy_run_command(self):
--- a/share/ert/forward-models/docs/examples/ECLIPSE100.rst
+++ b/share/ert/forward-models/docs/examples/ECLIPSE100.rst
@@ -1,9 +1,13 @@
-The version, number of cpu and whether or not to ignore errors can
+The version, number of cpu, and whether or not to ignore errors and whether or not to produce `YOUR_CASE_NAME.h5` output files can
 be configured in the configuration file when adding the job, as such:
 
 .. code-block:: bash
 
-    FORWARD_MODEL ECLIPSE100(<ECLBASE>, <VERSION>=xxxx, <OPTS>="--ignore-errors")
+    FORWARD_MODEL ECLIPSE100(<ECLBASE>, <VERSION>=xxxx, <OPTS>={"--ignore-errors", "--summary-conversion"})
 
-The :code:`OPTS` argument is optional and can be removed, thus running eclipse
-without ignoring errors
+The :code:`OPTS` argument is optional and can be removed, fully or partially.
+In absence of :code:`--ignore-errors` eclipse will fail on errors.
+Adding the flag :code:`--ignore-errors` will result in eclipse ignoring errors.
+
+And in absence of :code:`--summary-conversions` eclipse will run without producing `YOUR_CASE_NAME.h5` output files.
+Add flag :code:`--summary-conversions` to produce `YOUR_CASE_NAME.h5` output files.
--- a/tests/res/fm/test_ecl_run_new.py
+++ b/tests/res/fm/test_ecl_run_new.py
@@ -165,6 +165,32 @@ with open("env.json", "w") as f:
 
     @pytest.mark.equinor_test
     @tmpdir()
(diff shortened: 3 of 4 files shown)


tests/res/fm/test_ecl_run_new.py
```diff
@@ -163,6 +163,19 @@ def test_failed_run_OK(self):
...
+    def test_no_hdf5_output_with_ecl100(self):
+        self.init_eclrun_config()
+        shutil.copy(
+            os.path.join(self.SOURCE_ROOT, "test-data/local/eclipse/SPE1.DATA"),
+            "SPE1.DATA",
+        )
+        ecl_config = Ecl100Config()
+        run(ecl_config, ["SPE1.DATA", "--version=2019.3"])
+
+        self.assertTrue(os.path.isfile("non-existing-file.HDF5"))
```
sondreso: BTW, think HDF5 uses small letters in the extension 🙂 
```python
>>> import os
>>> os.path.isfile("setup.py")
True
>>> os.path.isfile("setup.PY")
False
```

sondreso: And linux uses a case sensitive filesystem by default while mac doesnt

kvashchuka: I assumed capital letter as in the rest of eclipse output but I might be wrong, will double check, thanks! However, the tests pass anyway and as you can see in the name of the file I want the test to fail..

sondreso: That is very interesting 🤔

markusdregi:
If we wish to expose the possibility of exporting as hdf5 to our users we should also add the argument in the job description files...

Also, it might be worth thinking about how close to `eclrun` we want to be. If we want to be almost compatible it makes sense to keep it as is, but if not then the more standard use of cli flags would be to have `--summary-conversion` give hdf5, and no flag give no export instead of the `"yes"/"no"` strings...

python/res/fm/ecl/ecl_run.py
```diff
@@ -161,10 +161,13 @@ class EclRun(object):
 
     """
 
-    def __init__(self, ecl_case, sim, num_cpu=1, check_status=True):
+    def __init__(
+        self, ecl_case, sim, num_cpu=1, check_status=True, summary_conversion="no"
```
markusdregi: Either way I suggest we make `summary_conversion` here into a boolean

kvashchuka:
> If we wish to expose the possibility of exporting as hdf5 to our users we should also add the argument in the job description files...
> 
> Also, it might be worth thinking about how close to `eclrun` we want to be. If we want to be almost compatible it makes sense to keep it as is, but if not then the more standard use of cli flags would be to have `--summary-conversion` give hdf5, and no flag give no export instead of the `"yes"/"no"` strings...

That is a very good comment. To be honest, I do not have any insight or preferences on this matter, I just adopted eclrun way of setting the flag

markusdregi:
That is fair. Since, we are reversing default behaviour I suggest we expose it. I just had a closer look and we already expose `OPTS`, which means that users can enable this by passing the flag as part of that argument:

https://github.com/equinor/libres/blob/81bfcb2493b1937fdaa37400b4315322b3ca4586/share/ert/forward-models/old_style/ECLIPSE300#L5

I suggest we move forward with `--summary-conversion` 👍🏻 Also, the job documentation should be updated...

share/ert/forward-models/docs/examples/ECLIPSE100.rst
```diff
@@ -1,9 +1,13 @@
...
 
 .. code-block:: bash
 
-    FORWARD_MODEL ECLIPSE100(<ECLBASE>, <VERSION>=xxxx, <OPTS>="--ignore-errors")
+    FORWARD_MODEL ECLIPSE100(<ECLBASE>, <VERSION>=xxxx, <OPTS>={"--ignore-errors", "--summary-conversion"})
 
-The :code:`OPTS` argument is optional and can be removed, thus running eclipse
-without ignoring errors
+The :code:`OPTS` argument is optional and can be removed, fully or partially.
+In absence of :code:`--ignore-errors` eclipse will run without ignoring errors.
```
markusdregi: `run without ignoring errors` -> `fail on errors`

python/res/fm/ecl/ecl_run.py
```diff
@@ -292,7 +298,6 @@ def execEclipse(self, eclrun_config=None):
                 else self._get_legacy_run_command()
             )
             env = eclrun_config.run_env if use_eclrun else self._get_legacy_run_env()
-
```
markusdregi: Please revert this line change to keep the footprint minimal.

tests/res/fm/test_ecl_run_new.py
```diff
@@ -163,6 +163,23 @@ def test_failed_run_OK(self):
...
+        shutil.copy(
+            os.path.join(self.SOURCE_ROOT, "test-data/local/eclipse/SPE1.DATA"),
+            "SPE1.DATA",
+        )
+        ecl_config = Ecl100Config()
+        # check that by default .h5 file IS NOT produced
+        run(ecl_config, ["SPE1.DATA", "--version=2019.3"])
+        self.assertFalse(os.path.exists("SPE1.h5"))
+
+        # check that with flag .h5 file IS produced
```
markusdregi, line 188: Should we split this up into two tests so that the the two states are tested completely independently?

markusdregi (changes requested):
Had a few minor comments. Besides those, this is ready to be merged

tests/res/fm/test_ecl_run_new.py
kvashchuka: sure, we can do that

markusdregi (approved):
Looks good 👍🏻 

Remember to squash your commits...

PR #1142, opened 2021-04-15
Fix 1220

jondequinor:
**Issue**
Resolves #1220

**Approach**
job_dispatch only sends job-level events, whereas step-level events are sent by the queue. Stage-level events are retired.


sondreso (approved):
LGTM!

PR #1147, opened 2021-04-27
Revert local_config_clear()

joakim-hove:
This commit reverts the changes in behavior in the local_config_clear() function
from commit: 7d29fdf0d and also introduces a new function
local_config_clear_active() which lets retains the named ministep, local_data
and obsdata objects for later reuse.

### DIFF
--- a/lib/enkf/local_config.cpp
+++ b/lib/enkf/local_config.cpp
@@ -137,11 +137,31 @@ static local_updatestep_type * local_config_alloc_updatestep( local_config_type 
 
 
 
+/*
+  The local_config_clear() function will remove all current local configuration,
+  and then reallocate a new empty updatestep configuration.
+*/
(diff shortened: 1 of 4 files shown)


ertomatic:

lars-petter-hauge:
Test this please!

lars-petter-hauge (approved):
Thanks for the quick PR

The test will implicitly test the functionality of `local_config_clear` versus `local_config_clear_active`. I tried writing a test which explicitly checked the content of `local_config`, and whether the added `Ministep` was retained in one scenario and removed in the other, but encountered some challenges where e.g. `hasMinistep` (currently not included) should be used instead of `getMinistep` whereas the latter would abort python if given ministep does not exist.

This will still be the case though of course, but I believe there are similar scenarios for more than just `ministep` when using the localisation functionality anyway.

PR #1149, opened 2021-04-28
Re-use Client for dispatch lifecycle

jondequinor:
**Issue**
We do a fairly expensive ssl handshake for _every_ event sent by a dispatcher. Even though this performs fine, it is wasteful.

**Approach**
Match the lifecycle of the client with that of the event reporter, using queues and threads and stuff.


sondreso (approved):
🚀

PR #1159, opened 2021-05-04
Allow cert to be None

sondreso:
**Issue**
If cert is None we cannot write it to a file.

**Approach**
Avoid writing cert to file if it's set to None, and write None as cert_path


markusdregi:
Would it be possible to accompany this with a test where `cert` is `None`?

markusdregi (approved):
Looks good 👍🏻 Thanks for the test!

PR #1164, opened 2021-05-07
Add method keys() to LocalDataset

### LINKED ISSUE #1160: Be able to get a list of nodes/nodenames from a modellparametergroup object
A model parameter group   (created by local_config.createDataset(model_group_name)) can contain one or more model parameter nodes (which is specified in ERT config). There does not exist any python  method for this object to get the list of nodes (or nodenames). It will simplify the use of libres in some cases if it is possible to get the nodes within a group without having to store that information in additional data structure (e.g dict) in the application script and send it into all functions that need that information.

One example of a use case is to check which model parameters in nodes belonging to a model parameter group is active or not.
By retrieving the list of nodes from the group and the active list from the nodes this is possible, but  not yet without additional book-keeping in the application script.

joakim-hove:
**Issue**
Resolves #1160

**Approach**
Expose a function in Python which was already available in C

### DIFF
--- a/python/res/enkf/local_dataset.py
+++ b/python/res/enkf/local_dataset.py
@@ -2,6 +2,7 @@ from cwrap import BaseCClass
 from res import ResPrototype
 from ecl.grid import EclRegion
 from ecl.util.geometry import GeoRegion
+from ecl.util.util import StringList
 
 
 class LocalDataset(BaseCClass):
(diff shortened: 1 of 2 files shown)


ertomatic:

markusdregi:
Jenkins: Test this please

markusdregi (approved):
Thanks for the contribution 👍🏻

PR #1168, opened 2021-06-30
Fix bug where last summary obs was not loaded

oyvindeide:
This is a back port of: https://github.com/equinor/ert/commit/edeb1fee2696e398e35f2cc94eda07d35a8e8086


markusdregi (approved):
I think this is worth a back port 👏🏻

